PR #1, opened 2017-08-17
add remimplementation of centrosome.cpmorphology.relabel

MikeDMorgan:
This is an almost verbatim re-implementation of centrosome.cpmorphology.relabel, only very subtle changes (use of a set object instead of np.unique, stylistic alterations). 

centrosome uses a BSD-3 license, which might clash with the MIT license on starfish as I believe BSD-3 propagates.

In the absence of specific unit tests, I've run this without error on the Mignardi notebook.


dganguli:
looks good to me!

PR #3, opened 2017-08-22
Reorganize into a package

mckinsel:
A few changes to make this a pip-installable package.


REQUIREMENTS.txt
```diff
@@ -0,0 +1,7 @@
+click
```
dganguli: should these be explicitly versioned?

dganguli:
lgtm

PR #5, opened 2017-08-23
Input format

dganguli:
Re-factor io.stack to read from a json file instead of two csv files.


mckinsel:
Right now `starfish register --u 3 output/org.json registered/` and `starfish register --u 3 output/org.json registered` behave differently, where the second command doesn't write to a directory but just creates a bunch of files that start with "registered". I think if you use `os.path.join` rather than string concatenation to create paths, it'll solve the problem and you won't have to worry about keeping track of slashes.

PR #8, opened 2017-08-27
renaming data  fields

freeman-lab:
Just some suggested renaming here:

`meta_data` -> `metadata` (this is a more common way to write it)
`aux_data` -> `aux` (possibly we'll have aux metadata)

So the fields are now `data`, `metadata`, and `aux`, which feels cleaner to me.

This change should only affect how we're writing the example data, and our `io` functions.


dganguli:
lgtm

PR #9, opened 2017-08-27
geojson ouput

freeman-lab:
This PR adds support for converting `spot` and `region` information into the `geoJSON` format, and then exports to those formats in the CLI operations for `detect_spots` and `segment`. Munging functions derived from example Jupyter notebook. Confirmed that generated `regions.json` files could be used in web-based data visualization.

### DIFF
--- a/starfish/munge.py
+++ b/starfish/munge.py
@@ -37,6 +37,32 @@ def gather(df, key, value, cols):
     return pd.melt(df, id_vars, id_values, var_name, value_name)
 
 
+def spots_to_geojson(r):
+    '''
+    Convert spot geometrical data to geojson format
+    '''
+    def make_dict(row):
+        row = row[1]
+        d = dict()
+        d['properties'] = {'id':int(row.spot_id), 'radius': int(row.r)}
+        d['geometry'] = {'type':'Point', 'coordinates': [int(row.x), int(row.y)]}
+        return d
+    return [make_dict(row) for row in spots_viz.iterrows()]
+
+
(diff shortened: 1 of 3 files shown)


starfish/munge.py
```diff
@@ -37,6 +37,32 @@ def gather(df, key, value, cols):
     return pd.melt(df, id_vars, id_values, var_name, value_name)
 
 
+def spots_to_geojson(r):
```
dganguli: this function already lives in viz.py::spots_geo_json. we have three options:

1. delete existing function in viz and use the one in munge. merge PR as is.
2. move your function to viz to replace it. fix imports.
3. use function in viz. remove changes to munge. update cli with new paths and functions

i'm in favor of 1 because this really is more of a munge operation than a viz operation and sees simplest. what do you think?

starfish/munge.py
```diff
@@ -37,6 +37,32 @@ def gather(df, key, value, cols):
...
+    def make_dict(row):
+        row = row[1]
+        d = dict()
+        d['properties'] = {'id':int(row.spot_id), 'radius': int(row.r)}
+        d['geometry'] = {'type':'Point', 'coordinates': [int(row.x), int(row.y)]}
+        return d
+    return [make_dict(row) for row in spots_viz.iterrows()]
+
+
+def regions_to_geojson(r):
```
dganguli: same as above, this lives in viz.py::regions_geo_json

dganguli:
Went with option 1: delete existing function in viz and use the one in munge.

PR #12, opened 2017-10-12
travis integration

ttung:
pip install and try to run the top-level binary with multiple python versions.

first run: https://travis-ci.org/chanzuckerberg/starfish/builds/287249490

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+python:
+- 2.7
+- 3.4
+- 3.5
+- 3.6
+dist: trusty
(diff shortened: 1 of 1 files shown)


ttung:
> Is there an advantage to creating a new starfish.py rather than installing with setup or pip and using the entry_point?

I think the primary advantage (when I wrote this) is that it avoids the ambiguity of which starfish you're running.  But now that I think about it, it's not even clear that this script accomplishes that.  I'll revert it back to the way it was.

ttung:
updated.

PR #18, opened 2017-10-27
ISS data integration test.

ttung:
This takes the instructions from `README.md` and turns it into an integration test.

Also this adds:

1. A `REQUIREMENTS-DEV.txt`, which is a superset of `REQUIREMENTS.txt` and includes what is needed for running tests.
2. A `Makefile` with rules for tests and lint.  Lint is currently disabled as there's a ton of stuff that doesn't pass lint right now.

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -1,6 +1,6 @@
 click
 matplotlib
-numpy
+numpy != 1.13.0
 opencv-python
 pandas
 regional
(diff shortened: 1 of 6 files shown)


REQUIREMENTS.txt
```diff
@@ -1,6 +1,6 @@
 click
 matplotlib
-numpy
+numpy != 1.13.0
```
dganguli: the issue is with skimage.peak_local_max (used by starfish.watershedsegmenter) calling the '-' operator on binary np arrays. the issue has been encountered by others (https://github.com/scikit-image/scikit-image/issues/2681)

PR #23, opened 2017-11-16
Share file formats and API for both MERFISH and ISS decoders

dganguli:
Spot detectors [Gaussian/Pixel] --> Input: Stack, Output: 'encoded' dataframe
Decoders [Merfish, ISS/ --> Input: 'encoded' dataframe. Output: 'decoded' dataframe

'encoded' and 'decoded' dataframes share the same schema.

This allows us to test the cross of pixel-based or gaussian-based spot detection with either a MERFISH or ISS data.


starfish/decoders/iss.py
```diff
@@ -2,44 +2,54 @@
 import pandas as pd
 
 
-def decode(spots_df_tidy, letters=['T', 'G', 'C', 'A']):
-    num_ch = spots_df_tidy.ch.max() + 1
-    num_hy = spots_df_tidy.hyb.max() + 1
-    num_spots = spots_df_tidy.spot_id.max() + 1
+class IssDecoder:
+    def __init__(self, encoded, codebook):
```
ttung: API-wise, I think it makes more sense for the decoder to be instantiated with the codebook, and the decode method to take the encoded data as an input.

starfish/spots/gaussian.py
```diff
@@ -6,30 +6,32 @@
 from showit import image
 from skimage.feature import blob_log
 
-from starfish.munge import stack_to_list, gather
+from starfish.munge import gather
 
 
 class GaussianSpotDetector:
-    def __init__(self, stack, blobs, measurement_type='max'):
+    def __init__(self, stack):
```
ttung: ditto here.  the detector should be parameterized in the scope of the constructor.  data is passed in.

imagine that you have a large number of images, and you set up the detector and do any heavy work involved with the setup there.  you can keep feeding in stuff, and extract out what you want.

starfish/starfish.py
```diff
@@ -141,15 +140,21 @@ def detect_spots(in_json, results_dir, aux_img, min_sigma, max_sigma, num_sigma,
     s = Stack()
     s.read(in_json)
 
-    gsp = GaussianSpotDetector(s.squeeze(), s.aux_dict[aux_img])
+    # create 'encoder table' standard (tidy) file format.
+    gsp = GaussianSpotDetector(s)
+    spots_df_tidy = gsp.detect(min_sigma=min_sigma,
```
ttung: not a requirement, but if you do:

```
spots_df_tidy = gsp.detect(
        min_sigma=min_sigma,
        ...,
)
```

you avoid having to reformat all the lines in the future should you choose to rename a function name or variable, such that your open paren moves.

starfish/spots/gaussian.py
ttung: Having said that, I am not opposed to you landing this, and I'll just add this to the list of cleanup tasks I'm going to do as I go through the repo.

starfish/decoders/iss.py
dganguli: 👍

dganguli:
The comments on this were addressed and was merged as PR 26.

PR #27, opened 2017-11-21
remove extra state about the current image in Merfish decoder.

ttung:
The decoder should be stateless if we want to treat it like a pure function.


ttung:
> pure function

It's cleaner, and less free-floating state leads to more maintainable code.

> Is there an alternative way to maintain access to this state that you would prefer?

Check out the new diff and see if it's to your liking.  Basically, `_decode()` is an API that returns all the intermediate state for your debugging pleasure, and ordinary consumers of the API can call `decode()`.

PR #30, opened 2018-01-02
switch to using argparse instead of click.

ttung:
Trying to modularize starfish with click is turning out to be real nightmare.  The reason is that click is very procedurally oriented, whereas applying an OO pattern to the various pipeline stages requires a tremendous amount of hackery.

### DIFF
--- a/tests/test_profiler.py
+++ b/tests/test_profiler.py
@@ -14,12 +14,18 @@ from starfish.starfish import PROFILER_NOOP_ENVVAR
 class TestProfiler(unittest.TestCase):
 
     def test_profiler(self):
-        """Make sure that `starfish --profile noop works."""
+        """Make sure that `starfish --profile --noop works."""
         cmdline = [
             "starfish",
             "--profile",
-            "noop",
+            "--noop",
         ]
+        env = os.environ.copy()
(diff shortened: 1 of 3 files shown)


tests/test_profiler.py
```diff
@@ -14,34 +14,18 @@
...
-        Ensure we do not expose the noop command unless the magic environment flag is present.  This is to avoid
-        confusion from the end users.
-        """
-        cmdline = [
-            "starfish",
-            "noop",
-        ]
+        if sys.version_info < (3, 0):
+            # this is a hideous hack because argparse is not consistent between 2.7 and 3.x about whether the subcommand
+            # is required.  this entire hack can be dropped if we deprecate python2.7.
```
dganguli: I'm OK deprecating python 2.7 -- especially if we keep having to put weird hacks in like this. What do you think?

ttung: I think dropping python2.7 is fine for the CLI.  It's much less clear whether we should drop python2.7 for API because people might have legacy 2.7 codebases that they want to integrate with starfish.  However, you know far more whether this is likely, and if you're okay, then YAY!!!

PR #32, opened 2018-01-03
[easy] add quick sanity test to ensure that the pipeline actually ran.

ttung:
Currently, we just verify that starfish runs and doesn't crash.  This adds a quick sanity test to ensure that _some_ result is generated and that the result seems reasonable.


ttung:
Added test for the top two barcodes detected.

ttung:
Got oral confirmation this is good to go. :)

PR #33, opened 2018-01-03
base class for registration

ttung:
Creates a base class for registration algorithms.  As long as the individual algorithms are imported into registration/__init__.py, they will be auto-discovered and be made available to the CLI / python API.

### DIFF
--- a/starfish/__init__.py
+++ b/starfish/__init__.py
@@ -1,0 +1,2 @@
+from .starfish import starfish
+from . import registration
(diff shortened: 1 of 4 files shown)


starfish/registration/_maxproject.py
```diff
@@ -0,0 +1,37 @@
+from ._base import RegistrationAlgorithmBase
+
+
+class MaxProjectionRegistration(RegistrationAlgorithmBase):
```
dganguli: rename to FourierTranslationRegistration or FourierTranslation

Issue #66, opened 2018-02-12
Example notebook for ISS data


dganguli:
Upload worked example of processed in-situ sequencing data (RCA w/ Padlock probes) in human breast tissue

Issue #54, opened 2018-02-08
Example notebook for MERFISH data


dganguli:
Upload worked example of processed MERFISH data

PR #74, opened 2018-02-14
Upload notebooks + code to localize sample dataset

ambrosejcarr:
**Not for merging**

* Adds code to localize data from from AWS. Right now it's coded against a private pe'er repo, but if we switch to a public repo using the same pattern, I've verified everything works. I will need to send you my zip files, which have been modified to have consistent structure across the datasets:
```
DATASET_NAME/
    /codebook.csv
    /fov_001
        /org.json
        /<images>.tiff
```
* Restricts numpy version. 1.14 deprecated a bunch of things and matplotlib needs to catch up. 
* added worked notebooks with notes explaining how the two datasets should be processed. I would be extremely surprised if I got everything right. Read my notes carefully. :)

Addresses #54, #66


starfish/io.py
```diff
@@ -22,6 +26,48 @@ def __init__(self):
...
+        :param str directory: (optional, default=None) if provided, localize the stack object to
+          this directory
+        :return Stack:
+        """
+        bucket = None  # wherever your data is
+        s3 = boto3.resource('s3')
+
+        # set a place to download the file
+        if directory is None:
+            directory = os.environ['TMPDIR']  # this isn't the most portable solution.
```
ttung: https://docs.python.org/2/library/tempfile.html

starfish/io.py
```diff
@@ -22,6 +26,48 @@ def __init__(self):
         self.read_fn = None  # set by self._read_metadata
         self.write_fn = np.save  # asserted for now
 
+    @classmethod
+    def download(cls, dataset, directory=None):
+        """load an example dataset as a stack
```
ttung: I prefer multiline docblocks like:

```
    """
    gibber jabber
    have a nice day.
    """
```

starfish/io.py
```diff
@@ -22,6 +26,48 @@ def __init__(self):
...
+          this directory
+        :return Stack:
+        """
+        bucket = None  # wherever your data is
+        s3 = boto3.resource('s3')
+
+        # set a place to download the file
+        if directory is None:
+            directory = os.environ['TMPDIR']  # this isn't the most portable solution.
+        if not directory.endswith('/'):  # neither is this, I hate windows.
```
ttung: drop this, and use the following later:

```
import posixpath

s3.Bucket(bucket).download_file(dataset, posixpath.join(directory, dataset))
```

starfish/io.py
```diff
@@ -22,6 +26,48 @@ def __init__(self):
         self.read_fn = None  # set by self._read_metadata
         self.write_fn = np.save  # asserted for now
 
+    @classmethod
+    def download(cls, dataset, directory=None):
```
ttung: Would recommend that this take bucket, path

starfish/io.py
```diff
@@ -22,6 +26,48 @@ def __init__(self):
...
+        archive.close()
+
+        # read in the archive
+        # looks like your zip files have more than one stack. do you want download to give multiple
+        # stacks? or should we pare down the number of fovs in the archive?
+        stack = cls()
+
+        # do we hard code the examples to have the same json name and location within the archive
+        # to support this kind of unpacking?
+        stack.read(directory + "org.json")
```
ttung: also posixpath

ambrosejcarr:
@dganguli did you want this tool to work like:

`s = Stack.download("merfish")`

Or

`s = Stack.download("merfish-bucket", "merfish-key.tar", "auth-stuff")`

dganguli:
@ambrosejcarr re: your comment: 1. @ttung probably has a strong opinion. 2. i think download functionality should not be a part of Stack. i like the way pytorch does it:

```
import torchvision.datasets as dset

train_set = dset.MNIST(root=root, train=True, transform=trans, download=download)
```

analogously you could do something like

```
import starfish.datasets 

stack = datasets.MERFISH(bucket=<some_bucket>, download=True)
type(stack) = Stack

```

You should be able to specify a bucket from which to download, be able to cache the data on disk, and be able to also load data from disk if it's already been cached.

ambrosejcarr:
@dganguli 
Ok, I've split this into 3 branches as requested offline and will deal with just the file upload in this one. I will open the other two branches up for PR once we figure out the download method. 

@ttung I'm looking for some insight on: 
(1) where should the zip files live? Referencing my original comment, I've restructured their organization a bit so they're consistent, and right now I've put them into an s3 bucket for private testing. I can slack the zip files to you if you want to put them somewhere public that I can code against. 
(2) Do you also approve of the pytorch implementation and @dganguli 's suggested adaptation, and use case requirements? If not, do you have any alternative or additional requirements?

ttung:
I think that it makes sense to build a generalized module that fetches a data set from http/S3/GS.  On top of that, we could build fetchers for specific data sets as convenience methods.

ambrosejcarr:
@ttung @dganguli I'm closing this, I think it's dead.

PR #76, opened 2018-02-15
Update the notebook to reflect the API changes.

ttung:


notebooks/Starfish Mignardi Pipeline.py
```diff
@@ -189,7 +189,9 @@
 dec_filt = pd.merge(dec, spots_viz, on='spot_id',how='left')
 dec_filt = dec_filt[dec_filt.qual>.25]
 
-rgb = np.zeros(s.im_shape + (3,))
+assert s.aux_dict['dapi'].shape == s.aux_dict['dots'].shape
```
dganguli: why is this here?

ttung: Just to check the input shapes are the same. I guess it's really not necessary.

Issue #73, opened 2018-02-14
aux lib to build starfish experiment dataset from the cloud


ttung:
This can make the .ipynb portable (or .py if we have a better tool for converting ipynb <-> py).

ambrosejcarr:
@ttung is "aux lib" slicedimage? Is this one?

ttung:
Yes, and yes!

PR #80, opened 2018-02-26
rename notebooks

ttung:
1. It's in the starfish repo.  Starfish prefix is redundant.
2. Make (i.e., GNU Make) chokes on files with spaces.  :(


ttung:
> These notebooks have not been run -- which makes it hard for the community to see what the actual results look like. Is there a way to make sure these notebooks are run (and actually work) either before or after a merge?

Yes, we can do that.  It requires that #73 be done first, however.

PR #95, opened 2018-03-07
Refactor decode stage into a PipelineComponent.

ttung:
Follow-on changes:
1. Get the codebook from a file.  This requires us to add the codebook to the pipeline API. (#96)
2. Merge `starfish.decoders.iss` into `starfish.pipeline.decoder._iss` (this will require refactoring the notebooks, but with @ambrosejcarr having landed the notebooks, it should be a lot easier). (#97)

### DIFF
--- a/starfish/pipeline/decoder/__init__.py
+++ b/starfish/pipeline/decoder/__init__.py
@@ -1,0 +1,48 @@
+from starfish.pipeline.pipelinecomponent import PipelineComponent
+from starfish.util.argparse import FsExistsType
+from . import _base
+from . import _iss
+
+
+class Decoder(PipelineComponent):
+    @classmethod
+    def implementing_algorithms(cls):
+        return _base.DecoderAlgorithmBase.__subclasses__()
+
+    @classmethod
+    def add_to_parser(cls, subparsers):
+        """Adds the decoder component to the CLI argument parser."""
+        decoder_group = subparsers.add_parser("decode")
+        decoder_group.add_argument("-i", "--input", type=FsExistsType(), required=True)
+        decoder_group.add_argument("-o", "--output", required=True)
+        decoder_group.add_argument("-c", "--codebook", type=FsExistsType(), required=True)
+        decoder_group.set_defaults(starfish_command=Decoder._cli)
+        decoder_subparsers = decoder_group.add_subparsers(dest="decoder_algorithm_class")
+
+        for algorithm_cls in cls.algorithm_to_class_map().values():
+            group_parser = decoder_subparsers.add_parser(algorithm_cls.get_algorithm_name())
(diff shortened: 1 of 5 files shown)


ttung:
To be done:
1. Get the codebook from a file.  This requires us to add the codebook to the pipeline API.
2. Merge `starfish.decoders.iss` into `starfish.pipeline.decoder._iss` (this will require refactoring the notebooks, but with @ambrosejcarr having landed the notebooks, it should be a lot easier).

starfish/pipeline/decoder/_iss.py
```diff
@@ -0,0 +1,26 @@
...
+        return "iss"
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        pass
+
+    def decode(self, encoded, codebook):
+        import pandas
+
+        # TODO this should be loaded from disk
```
ambrosejcarr: We should be able to write this toy example to disk, right? I think this probably belongs in a test case, since we expect the codebook to be experiment specific.

starfish/pipeline/decoder/_iss.py
```diff
@@ -0,0 +1,26 @@
...
+        d = {'barcode': ['AAGC', 'AGGC'], 'gene': ['ACTB_human', 'ACTB_mouse']}
+        codebook = pandas.DataFrame(d)
+        decoder = iss.IssDecoder(codebook, letters=['T', 'G', 'C', 'A'])
```
ambrosejcarr: I found this a bit confusing, as I thought this was initializing an instance of itself. I'm wondering if there is a better way to distinguish `iss.IssDecoder` and `_iss.IssDecoder`.

starfish/pipeline/decoder/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+            cls.decoder_group.print_help()
+            cls.decoder_group.exit(status=2)
+
+        instance = args.decoder_algorithm_class.from_cli_args(args)
+
+        encoded = pandas.read_json(args.input, orient="records")
+        codebook = pandas.read_json(args.codebook, orient="records")
+
+        results = instance.decode(encoded, codebook)
+        print("Writing | spot_id | gene_id to: {}".format(args.output))
```
ambrosejcarr: @dganguli I think we should make this message more expressive. From the message I don't fully understand what I should expect to be in the decoded file.

starfish/pipeline/decoder/_iss.py
ambrosejcarr: JK I just read 96. Nice job. 👍

starfish/pipeline/decoder/_iss.py
ttung: The long-term plan is to fold all the stuff in `starfish.decoders` into `starfish.pipeline.decoder`.

PR #96, opened 2018-03-07
Use the codebook loaded from a file.

ttung:
1. Modify `examples/get_iss_data.py` to create the codebook that was previously hardcoded in starfish.py
2. Modify `tests/test_iss_data.py` to extract the path of the codebook and feed it into the decode stage.

Depends on #95

### DIFF
--- a/tests/test_iss_data.py
+++ b/tests/test_iss_data.py
@@ -7,12 +7,32 @@ import sys
 import tempfile
 import unittest
 
+import jsonpath_rw
+
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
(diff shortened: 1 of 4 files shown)


tests/test_iss_data.py
```diff
@@ -7,6 +7,8 @@
 import tempfile
 import unittest
 
+import jsonpath_rw
```
ambrosejcarr: Is this sophisticated enough to warrant inclusion as a dependency? 

Why can't you just: 
```
with open(json_filepath, "r") as fh:
    document = json.load(fh)
return document['codebook'][0]
```

 Perhaps I don't understand what is happening in this function.

ttung:
#140 is the follow-on for this.

PR #102, opened 2018-03-09
Add a geojson-to-region converter.

ttung:
This will be necessary to separate gene assignment from segmentation, as we will need to load up geojson output from segmentation to do gene assignment.


ttung:
Nope, not as far as I know.

Kind of surprised the existing flow doesn't trigger the code in question though.

PR #103, opened 2018-03-09
Separate assignment from segmentation.

ttung:
Create a gene assignment pipeline component, and move the relevant code there.

Note that this changes the output of segmentation to output the full geometry of the regions, rather than the hull.

Depends on #102, #101, #100


ttung:
@dganguli says he can't take credit for deepassigner.

it's called PointInPoly.

ttung:
renamed to PointInPoly

PR #106, opened 2018-03-15
Makefile rules to regenerate .py/.ipynb notebooks.

ttung:
`make regenerate_ipynb` will regenerate all the `.ipynb` notebooks.
`make regenerate_py` will regenerate all the `.py` counterparts.


ttung:
`make validate_notebooks` validates, and is run on every commit. This only makes it easier to sync the .py with the .ipynb.

PR #107, opened 2018-03-16
RFC: New input format for starfish

ttung:
Modified to use https://github.com/ttung/slicedimage

### DIFF
--- a/examples/get_iss_data.py
+++ b/examples/get_iss_data.py
@@ -1,10 +1,14 @@
+import argparse
 import io
+import os
 import json
 import zipfile
 
-import click
 import requests
 from skimage.io import imread, imsave
+from slicedimage import ImageFormat, ImagePartition, Tile, Writer
+
+from starfish.util.argparse import FsExistsType
 
 
 def download(input_dir, url):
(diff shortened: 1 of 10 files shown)


examples/get_iss_data.py
```diff
@@ -16,97 +19,121 @@ def download(input_dir, url):
...
-    stack = imread(path)
-    return stack
+def build_hybridization_stack(input_dir, output_dir):
+    prefix = "fov_0"
 
+    imgs = list()
+    for hyb in range(1, 5):
+        hyb_round = list()
+        for ch in range(2, 6):
+            hyb_round.append(imread(os.path.join(input_dir, str(hyb), "c{}.TIF".format(ch))))
```
ambrosejcarr: I would find a way to do reading and writing inside the for loop that starts below on line 37 so that you only have one image in memory at a time. This could run into problems if we mirror it for larger datasets. 

If you store the input file stem in `imgs` instead of the loaded image, I think that would suffice.

notebooks/Mignardi_Pipeline.py
```diff
@@ -128,7 +128,7 @@
 disk_size_mask = None
 min_dist = 57
 
-seg = WatershedSegmenter(s.aux_dict['dapi'], s.aux_dict['stain'])
+seg = WatershedSegmenter(s.aux_dict['nuclei'], s.aux_dict['stain'])
```
ambrosejcarr: I think this change also needs to happen in the `ISS_Pipeline_-_Breast_-_1_FOV.[py|ipynb]` files. Related: We should write a CI function to run our `.py` encoded notebooks (and ideally the notebooks themselves), as discussed yesterday.

starfish/image/_stack.py
```diff
@@ -22,24 +22,22 @@ def __init__(self, data, tile_format, num_hybs, num_chs, tile_shape):
...
+        num_chs = len(set(tile['coordinates']['ch'] for tile in image_stack['tiles']))
+        tile_shape = tuple(image_stack['legend']['default_tile_shape'])
+        tile_format = ImageFormat[image_stack['legend']['default_tile_format']]
+        data = numpy.zeros((num_hybs, num_chs) + tile_shape)
 
-        for data_dict in data_dicts:
-            h = data_dict['hyb']
-            c = data_dict['ch']
-            fname = data_dict['file']
+        for tile_dict in image_stack['tiles']:
```
ambrosejcarr: Is there a more descriptive name we could use than `tile_dict`? This contains all the information about the provenance of a tile... 

@dganguli What would you call this? I think this is an object we'll reference a lot in vignettes / notebooks. `tile_metadata`?

examples/get_iss_data.py
```diff
@@ -16,97 +19,121 @@ def download(input_dir, url):
...
 
 
-def load_tiff_stack(path):
-    stack = imread(path)
-    return stack
+def build_hybridization_stack(input_dir, output_dir):
+    prefix = "fov_0"
 
+    imgs = list()
+    for hyb in range(1, 5):
```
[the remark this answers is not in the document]
ttung: As I understand it, we don't control the ISS data.  @dganguli would know more.

ttung:
Assuming this passes tests (which I really hope it does, as it does locally!), it's ready for review.

Things to keep in mind when reviewing:

1. `get_iss_data.py` is a mess because we're transforming from a bunch of images into the expected structure.  I have optimized for the main codebase to be neater with the expectation that this script is not likely to stick around forever.
2. Most of `stack/io.py` will likely go away.  
3. `starfish/image/_stack.py` will likely **stick around**.

ttung:
Here's an example of hybridization.json:

https://gist.github.com/ttung/4d5d08b3601904e0bcb9490b55aa244b

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+                    'x': (0.0, 0.0001),
+                    'y': (0.0, 0.0001),
+                },
+                {
+                    'hyb': hyb,
+                    'ch': ch,
+                },
+            )
+            path = os.path.join(input_dir, str(hyb + 1), "c{}.TIF".format(ch + 2))
+            tile.set_source_fh_callable(
```
ambrosejcarr: Is this checking that the file handle can be opened, and then adding it to the imageslice?

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+            )
+            imageslice.add_tile(tile)
+
+    return imageslice
+
+
+def build_fov(input_dir, hybridization_stack_name, output_dir):
+    prefix = "fov_0"
+
+    nuclei = imread(input_dir + "DO/c1.TIF")
```
ambrosejcarr: It seems to me like there is metadata in TIFF files that we could extract to learn about the shape without reading the whole file into memory. This could make these kinds of interfaces much faster for large datasets. 

ref: http://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_TIFF_files

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+    dots = imread(input_dir + "DO/c2.TIF")
+
+    experiment = {
+        'version': "0.0.0",
+        'hybridization_images': hybridization_stack_name,
+        'auxiliary_images': {},
+    }
+
+    nuclei_fname = "{}_{}.tiff".format(prefix, "nuclei")
+    imsave(output_dir + nuclei_fname, nuclei)
```
ambrosejcarr: could we `os.rename` instead of rewriting the files?

starfish/image/_stack.py
```diff
@@ -1,54 +1,47 @@
...
 
-from ._base import ImageBase, ImageFormat
+from ._base import ImageBase
 
 
 class ImageStack(ImageBase):
-    def __init__(self, data, tile_format, num_hybs, num_chs, tile_shape):
-        self._data = data
-        self._tile_format = tile_format
+    def __init__(self, slicedimage):
```
ambrosejcarr: use of `slicedimage` here is a bit confusing since it's also the name of the package. I went and looked, I guess this should be `image_partition`, instead?

starfish/image/_stack.py
```diff
@@ -1,54 +1,47 @@
...
-        return ImageStack(data, tile_format, num_hybs, num_chs, tile_shape)
+        return ImageStack(slicedimage)
 
     @property
     def numpy_array(self):
         return self._data
 
+    @numpy_array.setter
+    def numpy_array(self, data):
+        for tile in self._slicedimage.get_matching_tiles():
```
ambrosejcarr: This name is also a bit confusing for me, since what we're doing is just getting all the tiles. `filter_tiles(filter=None)` is a bit clearer to me.

Related, since the setter doesn't directly set on the `self._data` attribute referenced in the property, I think this deserves a comment; it's almost certainly going to confuse contributors.

starfish/image/_stack.py
```diff
@@ -72,27 +65,46 @@ def tile_shape(self):
...
+        seen_x_coords, seen_y_coords = set(), set()
+        for tile in self._slicedimage.get_matching_tiles():
+            seen_x_coords.add(tile.coordinates['x'])
+            seen_y_coords.add(tile.coordinates['y'])
+
+        sorted_x_coords = sorted(seen_x_coords)
+        sorted_y_coords = sorted(seen_y_coords)
+        x_coords_to_idx = {coords: idx for idx, coords in enumerate(sorted_x_coords)}
+        y_coords_to_idx = {coords: idx for idx, coords in enumerate(sorted_y_coords)}
+        # TODO: should handle Z.
```
ambrosejcarr: I would handle Z in this PR unless it's a huge nightmare. The two data types I'm working on now and next are volumetric. It would be nice to know it works! 

I have some example data that I can upload including org jsons in the _old_ specification if that gives you enough testing data.

examples/get_iss_data.py
ambrosejcarr: I hear this metadata isn't standardized 👎. Retracted.

ambrosejcarr: Posed more generally: can we make the metadata json without reading the whole dataset?

examples/get_iss_data.py
```diff
@@ -1,13 +1,17 @@
...
 
+import argparse
 import io
+import os
 import json
 import zipfile
 
-import click
 import requests
+from slicedimage import ImageFormat, ImagePartition, Tile, Writer
```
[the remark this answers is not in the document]
ttung: What does `lims` stand for?

examples/get_iss_data.py
ttung: No, all tiles either have data set via numpy array, or they're given a future.  Only when the data is requested is the file opened and read.

examples/get_iss_data.py
ttung: Um, possibly, but this script is of rather minor importance and will likely go away in the long-term.  I don't want to invest a whole lot in making it good. :)

examples/get_iss_data.py
ttung: Yes, but same as above.

starfish/image/_stack.py
ttung: Yeah, this part is not great IMO.  It might be that we need to treat `._data` as a cache of sorts, with dirty bits and all.

Let me ponder this.

starfish/image/_stack.py
ttung: I think the tiles will remain 2D unless we see a compelling 3D compression format.

Therefore, making it work for Z will probably be on the rather dull side.  Will fix.

examples/get_iss_data.py
ambrosejcarr: Yep, definitely agree. I was asking more generally for the loaders/converters into slicedimage that will eventually be written. No need to make changes here.

examples/get_iss_data.py
ambrosejcarr: Ah, cool. I definitely need to go read slicedimage. :)

examples/get_iss_data.py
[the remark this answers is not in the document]
ambrosejcarr: It determines what values the minimum and maximum value of the (usually linear) colormap map onto in the data. It's to prevent cases where a very large high value outlier compresses the colorspace by, for example, assigning 99% of the values to something that looks more or less like "black" in a greyscale colormap.

starfish/image/_stack.py
ambrosejcarr: Sounds good. 👍

starfish/image/_stack.py
ambrosejcarr: Yep, I agree re: 2d vs 3d. Thanks, it will make it easier for the data to come. :)

examples/get_iss_data.py
ttung: Theoretically, the shape of the tiles should be known without opening any image files.

ttung: If there is metadata we want to know, we should put it in the json file because the tiffs/npy files might be on the network, in which case, opening any of it is slow.

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+        'tile_shape': dots.shape,
+        'tile_format': "TIFF",
+        'coordinates': {
+            'x': (0.0, 0.0001),
+            'y': (0.0, 0.0001),
+        },
+    }
+
+    return experiment
+
```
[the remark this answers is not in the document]
ttung: #96 adds that.

examples/get_iss_data.py
```diff
@@ -1,13 +1,17 @@
 from __future__ import division
 
+import argparse
```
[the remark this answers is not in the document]
ttung: eh?  which import is out of order?

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+def build_fov(input_dir, hybridization_stack_name, output_dir):
+    prefix = "fov_0"
+
+    nuclei = imread(input_dir + "DO/c1.TIF")
+    dots = imread(input_dir + "DO/c2.TIF")
+
+    experiment = {
+        'version': "0.0.0",
+        'hybridization_images': hybridization_stack_name,
+        'auxiliary_images': {},
```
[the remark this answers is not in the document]
ttung: TOC partition / image partition / tiles do not require `extras`.  it's an optional field and there's code that handles this.

I do not know if `extras` in experiment.json should be optional.  I think so, though.

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
-
-    dapi_fname = '{}_{}.tiff'.format(prefix, 'dapi')
-    imsave(output_dir + dapi_fname, dapi)
-    res['aux'].append({'type': 'dapi', 'file': dapi_fname, 'format': "TIFF"})
-
-    dots_fname = '{}_{}.tiff'.format(prefix, 'dots')
+def build_hybridization_stack(input_dir):
+    default_shape = imread(os.path.join(input_dir, str(1), "c{}.TIF".format(1))).shape
+
+    imageslice = ImagePartition(["x", "y", "hyb", "ch"], {'hyb': 4, 'ch': 4}, default_shape, ImageFormat.TIFF)
```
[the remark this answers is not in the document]
ttung: got it, will do.

examples/get_iss_data.py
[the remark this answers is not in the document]
dganguli: oh i just use pycharm to optimize imports. it spits this order out:

import json
import os

import numpy as np
import pandas as pd
from slicedimage import ImageFormat
from slicedimage.io import resolve_url

from .image import ImageStack
from .munge import list_to_stack

pretty minor comment. sorry for nit picking!

starfish/image/_stack.py
```diff
@@ -72,27 +65,46 @@ def tile_shape(self):
...
+                    "wb")
+
+        Writer.write_to_path(
+            self._slicedimage,
+            filepath,
+            pretty=True,
+            tile_opener=tile_opener)
 
     def max_proj(self, dim):
         valid_dims = ['hyb', 'ch', 'z']
```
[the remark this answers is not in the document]
ttung: #122

starfish/image/_stack.py
```diff
@@ -1,54 +1,47 @@
...
-        self._tile_shape = tile_shape
+        self._data = numpy.zeros((self._num_hybs, self._num_chs) + self._tile_shape)
 
-        if len(tile_shape) == 2:
+        for tile in slicedimage.get_matching_tiles():
+            h = tile.indices['hyb']
+            c = tile.indices['ch']
+            self._data[h, c, :] = tile.numpy_array
+
+        if len(self._tile_shape) == 2:
```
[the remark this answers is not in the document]
ttung: Good catch.

examples/get_iss_data.py
```diff
@@ -16,97 +20,132 @@ def download(input_dir, url):
...
+        json.dump(res, outfile, indent=4)
 
-    res['metadata']['num_hybs'] = 4
-    res['metadata']['num_chs'] = 4
-    res['metadata']['shape'] = img.shape
-    res['metadata']['is_volume'] = is_volume
-    res['metadata']['format'] = "TIFF"
 
-    return res
+def tile_opener(toc_path, tile, ext):
```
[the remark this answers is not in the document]
ttung: will fix.

starfish/image/_stack.py
```diff
@@ -72,27 +65,46 @@ def tile_shape(self):
...
-                tile = {
-                    'hyb': hyb,
-                    'ch': ch,
-                    'file': "{}.{}".format(tile_filename, ImageFormat.NUMPY.file_ext),
-                }
-                numpy.save(os.path.join(basepath, tile_filename), self._data[hyb, ch, :])
-                data.append(tile)
-
-        return data
+    def clone_shape(self):
```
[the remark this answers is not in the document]
ttung: I'm getting rid of this.  It turns out to be unnecessary.

ttung:
I think I've resolved all the issues brought up **except** handling 3D data.  I'd like to handle that in a separate PR because this PR is complex enough.  I suspect it'll just get more complicated if I try to do 3D here.  I won't land this until the 3D PR is out.

PR #114, opened 2018-04-04
unbreak the build

ttung:
Ignore the notebook metadata when validating that the .ipynb and the .py match.  This is because subtle and unimportant changes are written there, such as the version of the python interpreter used to evaluate the notebook.  None of this matters as far as how the notebook code and layout is preserved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL := /bin/bash
+
 MODULES=starfish tests
 
 lint:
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
Can you explain how the makefile process works here? I'm not familiar with it.

ambrosejcarr:
Also, is it possible to have the Codecov hook not run when coverage isn't run?

PR #116, opened 2018-04-09
Use numpy to reshape images

olgabot:
~4 orders of magnitude faster to use `numpy.ndarray.reshape` instead of lists and double for loops.

screen shot 2018-04-09 at 10 21 32 am


ttung:
@olgabot  can you rebase?  i think it should pass now.

PR #126, opened 2018-04-24
Don't call squeeze/unsqueeze for filter.

ttung:
It's superfluous.

### DIFF
--- a/starfish/starfish.py
+++ b/starfish/starfish.py
@@ -111,14 +111,13 @@ def filter(args, print_help=False):
     s = Stack()
     s.read(args.in_json)
 
-    # filter raw images, for all hybs and channels
-    stack_filt = []
-    for im_num, im in enumerate(s.squeeze()):
-        print("Filtering image: {}...".format(im_num))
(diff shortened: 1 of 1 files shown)


starfish/starfish.py
```diff
@@ -116,14 +116,13 @@ def filter(args, print_help=False):
     s = Stack()
     s.read(args.in_json)
 
-    # filter raw images, for all hybs and channels
-    stack_filt = []
-    for im_num, im in enumerate(s.squeeze()):
-        print("Filtering image: {}...".format(im_num))
-        im_filt = white_top_hat(im, args.ds)
-        stack_filt.append(im_filt)
+    res = np.zeros(s.image.shape)
```
ambrosejcarr: General comment: We should make sure the `dtype` for `np.zeros` matches the output here, otherwise pre-allocation is worthless. If `input_dtype==output_dtype`, we can use `np.zeros_like()`. Otherwise we should specify.  I suspect the answer is rarely `float64` (default).

ttung: This goes away in #135

PR #127, opened 2018-04-24
Fix decode documentation and filenames

ttung:
#95 changed how decode works.  This fixes up the documentation.

Removed -c as an option for specifying codebook.

I also snuck in a change renaming decoder_table to decoded_table. :)


ttung:
> I like it. If we're going to use keyword parameters (-i, c), we should be consistent across the CLI.

-i/--input is standardized.  -c/--codebook i should switch to just --codebook, huh...

> I note this because segment (above decode) is now different - I assume that comes later?

yes.

ttung:
Changed it so -c is gone.

PR #128, opened 2018-04-24
Fix some import issues

ttung:
1. starfish/image/_base.py don't need those imports.
2. lazy-import iss because that brings in a ton of dependencies.


ttung:
I folded this into #107 since that introduced the code movement that necessitated the removal of imports.

PR #129, opened 2018-04-24
Support max projection along multiple axes.

ttung:
This is required for 3d support.

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -104,19 +104,17 @@ class ImageStack(ImageBase):
             pretty=True,
             tile_opener=tile_opener)
 
-    def max_proj(self, dim):
-        valid_dims = [Indices.HYB, Indices.CH, Indices.Z]
-        if dim not in valid_dims:
-            msg = "Dimension: {} not supported. Expecting one of: {}".format(dim, valid_dims)
-            raise ValueError(msg)
+    def max_proj(self, *dims):
+        valid_dims = {
+            Indices.HYB: 0,
+            Indices.CH: 1,
+            Indices.Z: 2,
+        }
+        axes = list()
+        for dim in dims:
+            try:
+                axes.append(valid_dims[dim])
+            except KeyError:
+                raise ValueError("Dimension: {} not supported. Expecting one of: {}".format(dim, valid_dims.keys()))
 
-        if dim == Indices.HYB:
-            res = numpy.max(self._data, axis=0)
-        elif dim == Indices.CH:
-            res = numpy.max(self._data, axis=1)
-        elif dim == Indices.Z and len(self._tile_shape) > 2:
-            res = numpy.max(self._data, axis=4)
-        else:
-            res = self._data
-
-        return res
+        return numpy.max(self._data, axis=tuple(axes))
--- a/starfish/io.py
+++ b/starfish/io.py
@@ -86,8 +86,8 @@ class Stack:
             }
         self.aux_dict[key] = img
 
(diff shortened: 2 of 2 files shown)


starfish/image/_stack.py
```diff
@@ -104,19 +104,17 @@ def tile_opener(toc_path, tile, ext):
...
-            res = numpy.max(self._data, axis=4)
-        else:
-            res = self._data
-
-        return res
+    def max_proj(self, *dims):
+        valid_dims = {
+            Keys.HYB: 0,
+            Keys.CH: 1,
+            Keys.ZLAYER: 2,
```
dganguli: rename keys.ZLAYER -> keys.Z to maintain consistency with spec.

ttung: I would change the spec in that case. :)  Z could be misconstrued as a coordinate.

ambrosejcarr: This skimage spec seems to say, for video, that we should be using `(t, pln, row, col, ch)` http://scikit-image.org/docs/dev/user_guide/numpy_images.html

I would interpret our mapping to this as: 
```
t = hyb
pln = z
row = y
col = x
ch = ch
```
This would suggest we move `ch` to the end. I'm not sure I completely understand why they suggest this.

ttung: Actually, let's have the naming discussion on #130. I sent out the PRs in the wrong order. My bad.

ttung:
@dganguli i'm re-requesting review on this one because i think the concerns you have pertain to #130 and not this PR. :)

starfish/image/_stack.py
```diff
@@ -104,19 +104,17 @@ def tile_opener(toc_path, tile, ext):
...
+            Indices.HYB: 0,
+            Indices.CH: 1,
+            Indices.Z: 2,
+        }
+        axes = list()
+        for dim in dims:
+            try:
+                axes.append(valid_dims[dim])
+            except KeyError:
+                raise ValueError("Dimension: {} not supported. Expecting one of: {}".format(dim, valid_dims))
```
ambrosejcarr: should `valid_dims` be the full dictionary or just the keys? Related, do you need to call `repr()` on it or will this be done automatically by the string formatting?

starfish/image/_stack.py
```diff
@@ -104,19 +104,17 @@ def tile_opener(toc_path, tile, ext):
...
-        elif dim == Indices.CH:
-            res = numpy.max(self._data, axis=1)
-        elif dim == Indices.Z and len(self._tile_shape) > 2:
-            res = numpy.max(self._data, axis=4)
-        else:
-            res = self._data
-
-        return res
+    def max_proj(self, *dims):
+        valid_dims = {
```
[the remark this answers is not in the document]
ttung: I agree.  Let's do that in a separate PR.

starfish/image/_stack.py
ttung: good catch.

starfish/io.py
```diff
@@ -86,8 +86,8 @@ def set_aux(self, key, img):
             }
         self.aux_dict[key] = img
 
-    def max_proj(self, dim):
-        return self.image.max_proj(dim)
+    def max_proj(self, *dim):
```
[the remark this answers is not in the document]
ttung: will fix.

starfish/image/_stack.py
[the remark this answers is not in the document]
ttung: #134

PR #130, opened 2018-04-24
Use "constants" instead of magic strings.

ttung:
Things like 'hyb', 'ch', etc. should be referred to by constants rather than magic strings.  That makes 1) usage easier to track, and 2) easier to refactor.

### DIFF
--- a/examples/get_iss_data.py
+++ b/examples/get_iss_data.py
@@ -8,6 +8,7 @@ import requests
 from skimage.io import imread, imsave
 from slicedimage import ImageFormat, ImagePartition, Tile, Writer
 
+from starfish.image import Coordinates, Indices
 from starfish.util.argparse import FsExistsType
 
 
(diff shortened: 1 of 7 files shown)


starfish/image/_stack.py
```diff
@@ -104,16 +105,16 @@ def tile_opener(toc_path, tile, ext):
             tile_opener=tile_opener)
 
     def max_proj(self, dim):
-        valid_dims = ['hyb', 'ch', 'z']
+        valid_dims = [Keys.HYB, Keys.CH, Keys.ZLAYER]
```
ambrosejcarr: Does this assume the user will request the dim by its name? Can we also support the numeric value of the axis?

examples/get_iss_data.py
```diff
@@ -10,6 +10,7 @@
 from skimage.io import imread, imsave
 from slicedimage import ImageFormat, ImagePartition, Tile, Writer
 
+from starfish.image import Keys
```
dganguli: @ambrosejcarr what do you think of the name Keys for this? at the risk of starting this over, i prefer Dims or Axes 😁

starfish/image/_constants.py
```diff
@@ -0,0 +1,8 @@
+class Keys(object):
+    X_COORDS = 'x'
+    Y_COORDS = 'y'
+    Z_COORDS = 'z'
+
+    HYB = 'hyb'
+    CH = 'ch'
+    ZLAYER = 'zlayer'
```
dganguli: rename to Z

starfish/image/_stack.py
ambrosejcarr: @dganguli I'm interested in your thoughts here. I see Tony's point, but I'm wondering what you think users will think about having to use names only.

starfish/image/_constants.py
ttung: Actually, we could do:

```
class CoordinateKey(object):
    X = 'x'
    Y = 'y'
    Z = 'z'

class IndexKey(object):
    HYB = 'hyb'
    CH = 'ch'
    Z = 'z'
```

it would disambiguate between the two.

ttung: Would that be ok with you @dganguli ?

dganguli: love it. i would drop the Key though, e.g, CoordinateKey -> Coordinates and IndexKey -> Indices or Axes. Key is somehow the wrong word for both dimensions/indices/axes and coordinates

ttung: it is primarily used as a key to a dict, but not always, i guess.  ok, i'll make the change.

starfish/image/_stack.py
dganguli: i personally love it. i'm not a fan of keeping track of numeric values of the axes -- it leads to a lot of painful bugs for the user (e.g., me when i first started working on this) and makes it hard to reorder axes internally (which i also struggled with initially). but i can be convinced otherwise 😀 i think people can always get the numpy array if they want to just deal with that too right?

ttung:
Updated as per discussion.

starfish/image/_stack.py
ambrosejcarr: Ah, fair point. Ok, I'm fine with this then.

PR #131, opened 2018-04-24
Support for 3D stacks.

ttung:
I made some guesses for what might make sense for fourier_shift and stain creation.

Depends on #126, #129, #130

### DIFF
--- a/starfish/pipeline/registration/fourier_shift.py
+++ b/starfish/pipeline/registration/fourier_shift.py
@@ -25,7 +25,8 @@ class FourierShiftRegistration(RegistrationAlgorithmBase):
     def register(self, stack):
         import numpy as np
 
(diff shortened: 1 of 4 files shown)


starfish/pipeline/registration/fourier_shift.py
```diff
@@ -25,7 +25,7 @@ def add_arguments(cls, group_parser):
     def register(self, stack):
         import numpy as np
 
-        mp = stack.max_proj(Keys.CH)
+        mp = stack.max_proj(Keys.CH, Keys.ZLAYER)
```
dganguli: hmmm. add a #TODO here "we should think harder how fourier registration should work in 3D"

ttung:
Note that this means everything is treated as 3D.  2D data is auto-converted into 3D.

PR #134, opened 2018-04-25
Add slice operations to ImageStack

ttung:
This allows users to specify the axes and the values by which to slice the data.  This is the preferred way of accessing data, since it allows us to change the order of dimensions, though the `numpy_array()` accessor remains accessible.

Depends on #133

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -1,4 +1,5 @@
 import os
+import typing
 
 import numpy
 from slicedimage import Reader, Writer
@@ -8,6 +9,13 @@ from ._constants import Coordinates, Indices
 
 
 class ImageStack(ImageBase):
+    AXES_MAP = {
+        Indices.HYB: 0,
+        Indices.CH: 1,
+        Indices.Z: 2,
+    }
+    N_AXES = max(AXES_MAP.values()) + 1
+
     def __init__(self, image_partition):
         self._image_partition = image_partition
         self._num_hybs = image_partition.get_dimension_shape(Indices.HYB)
@@ -25,7 +33,7 @@ class ImageStack(ImageBase):
             h = tile.indices[Indices.HYB]
             c = tile.indices[Indices.CH]
             zlayer = tile.indices.get(Indices.Z, 0)
-            self._data[h, c, zlayer, :] = tile.numpy_array
+            self.set_slice(indices={Indices.HYB: h, Indices.CH: c, Indices.Z: zlayer}, data=tile.numpy_array)
 
     @classmethod
     def from_image_stack(cls, image_stack_name_or_url, baseurl):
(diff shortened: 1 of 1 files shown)


starfish/image/_stack.py
```diff
@@ -45,6 +52,23 @@ def numpy_array(self, data):
         self._data_needs_writeback = True
         data.setflags(write=False)
 
+    def slice(self, indices):
```
ambrosejcarr: This definitely needs a docstring. Looks like you expect a dict of 3-tuples with `start, stop, step` as indices? 

This is a bit clunky. I think at the very least we should be able to default to step=None since almost no one will want that (except, i guess, uniform sub-sampling?)

starfish/image/_stack.py
```diff
@@ -45,6 +52,23 @@ def numpy_array(self, data):
...
+    def set_slice(self, indices, data):
+        slice_list = [slice(None, None, None) for _ in range(ImageStack.MAX_INDEX)]
+        for name, val in indices.items():
+            idx = ImageStack.INDEX_MAP[name]
+            slice_list[idx] = val
+        self._data[tuple(slice_list)] = data
+        self._data_needs_writeback = True
+
     @property
     def shape(self):
```
ambrosejcarr: Given our switch over to named dimensions, can we have shape output explicit dimension names instead of just (8, 2, 34, 2048, 2048)

e.g. 
```
> self.shape
HYB: 8
CH: 2
Z: 34
X: 2048
Y: 2048
```

starfish/image/_stack.py
```diff
@@ -8,6 +8,13 @@
 
 
 class ImageStack(ImageBase):
+    INDEX_MAP = {
+        Indices.HYB: 0,
+        Indices.CH: 1,
+        Indices.Z: 2,
+    }
+    MAX_INDEX = max(INDEX_MAP.values()) + 1
```
ambrosejcarr: maybe `N_INDICES` ? I think of max_index as the largest dimension.

starfish/image/_stack.py
ttung: i like that.

starfish/image/_stack.py
ttung: will fix.

starfish/image/_stack.py
dganguli: +1 -- all this stack stuff should be over-documented, as it's the fundamental implementation of the input file format. from reading the code it's not  super obvious to me what's going on here

ttung: Added docs and some simple examples.

starfish/image/_stack.py
ttung: Changed.

starfish/image/_stack.py
ttung: #137

starfish/image/_stack.py
```diff
@@ -45,6 +53,107 @@ def numpy_array(self, data):
         self._data_needs_writeback = True
         data.setflags(write=False)
 
+    def get_slice(
```
dganguli: rename to get_tile? we don't use the word slice in our spec doc, but i think it means tile?

starfish/image/_stack.py
```diff
@@ -45,6 +53,107 @@ def numpy_array(self, data):
...
+            self,
+            indices: typing.Mapping[Indices, typing.Union[int, slice]]
+    ) -> typing.Tuple[numpy.ndarray, typing.Sequence[Indices]]:
+        """
+        Given a dictionary mapping the index name to either a value or a slice range, return a numpy array representing
+        the slice, and a list of the remaining axes beyond the normal x-y tile.
+
+        Example:
+            ImageStack axes: H, C, and Z with shape 3, 4, 5, respectively.
+            ImageStack Implicit axes: X, Y with shape 10, 20, respectively.
```
dganguli: ordering should be H, C, Z, Y, X -- i think you have H, C, Z, X Y

starfish/image/_stack.py
dganguli: oh i see, a slice can return a tensor in addition to a matrix. ignore prevous comment

starfish/image/_stack.py
```diff
@@ -45,6 +53,107 @@ def numpy_array(self, data):
...
+            Called to slice with indices {Z: 5}.
+            Result: a 4-dimensional numpy array with shape (3, 4, 10, 20) and the remaining axes [H, C].
```
ambrosejcarr: This comment is a bit unclear I think -- the remaining axes are already in the `(3, 4, 10, 20)` shape, so the last clause confused me a bit.

starfish/image/_stack.py
```diff
@@ -45,6 +53,107 @@ def numpy_array(self, data):
...
+
+        Example:
+            Original axes: H, C, and Z.
+            Implicit axes: X, Y.
+            Called to slice with indices {Z: 5, C: slice(2, 4)}.
+            Result: a 4-dimensional numpy array with shape (3, 2, 10, 20) and the remaining axes [H, C].
```
ambrosejcarr: Same comment as above, but otherwise I think these two examples make really clear what you can do here. 👍

starfish/image/_stack.py
```diff
@@ -45,6 +53,107 @@ def numpy_array(self, data):
...
+            raise ValueError("source shape {} mismatches destination shape {}".format(
+                data.shape, self._data[slice_list].shape))
+
+        self._data[slice_list] = data
+        self._data_needs_writeback = True
+
+    def _build_slice_list(
+            self,
+            indices: typing.Mapping[Indices, typing.Union[int, slice]]
+    ) -> typing.Tuple[typing.Tuple[typing.Union[int, slice], ...], typing.Sequence[Indices]]:
```
ambrosejcarr: Nice.

starfish/image/_stack.py
```diff
@@ -1,4 +1,5 @@
 import os
+import typing
```
ambrosejcarr: 👍

PR #135, opened 2018-04-25
use the slice operations to get and set data.

ttung:
This removes the need to allocate a new array, and makes the pipeline code unaware of the ordering of the axes.

Depends on #134

### DIFF
--- a/starfish/starfish.py
+++ b/starfish/starfish.py
@@ -116,23 +116,23 @@ def filter(args, print_help=False):
     print("Filtering dots ...")
     dots_filt = white_top_hat(s.aux_dict['dots'], args.ds)
 
-    print("Writing results ...")
     # create a 'stain' for segmentation
     # TODO: (ambrosejcarr) is this the appropriate way of dealing with Z in stain generation?
     stain = np.mean(s.max_proj(Indices.CH, Indices.Z), axis=0)
     stain = stain / stain.max()
-
-    res = np.zeros(s.image.shape)
 
     # filter raw images, for all hybs, channels
     for hyb in range(s.image.num_hybs):
         for ch in range(s.image.num_chs):
             for zlayer in range(s.image.num_zlayers):
                 print("Filtering image: hyb={} ch={} zlayer={}...".format(hyb, ch, zlayer))
-                res[hyb, ch, zlayer, :, :] = white_top_hat(s.image.numpy_array[hyb, ch, zlayer, :, :], args.ds)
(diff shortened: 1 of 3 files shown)


starfish/starfish.py
```diff
@@ -120,23 +120,21 @@ def filter(args, print_help=False):
...
-    res = np.zeros(s.image.shape)
-
     # filter raw images, for all hybs, channels
     for hyb in range(s.image.num_hybs):
         for ch in range(s.image.num_chs):
             for zlayer in range(s.image.num_zlayers):
                 print("Filtering image: hyb={} ch={} zlayer={}...".format(hyb, ch, zlayer))
-                res[hyb, ch, zlayer, :, :] = white_top_hat(s.image.numpy_array[hyb, ch, zlayer, :, :], args.ds)
+                indices = {Indices.HYB: hyb, Indices.CH: ch, Indices.Z: zlayer}
+                result = white_top_hat(s.image.slice(indices=indices), args.ds)
```
ambrosejcarr: args.ds should probably have a long-form name that is more expressive. I have no idea what this is. 🙃

starfish/starfish.py
```diff
@@ -120,23 +120,21 @@ def filter(args, print_help=False):
...
+                s.image.set_slice(indices=indices, data=result)
 
-    # update stack
-    s.set_stack(res)
+    print("Writing results ...")
```
ambrosejcarr: I might include a verbosity trigger here. Users might want to be able to turn off the reporting.

starfish/starfish.py
ttung: I believe it's disk-size.  Time for another "TODO: (@dganguli) docblock this" :)

starfish/starfish.py
ttung: This is a move from L123.  I agree though that we need global verbosity settings.

starfish/starfish.py
dganguli: yeah ds is the disk-size parameter used for the white top-hat filter. i think it makes sense to take a pass at documenting all of this after the filter command is re-expressed like the register command, e.g., when you can do starfish filter <filter_type> <filter_type specific arg 1> ... <filter_type specific arg N> etc. This is because the disk-size is specific to the white-tophat filter. The other filters, e.g., Gaussian low pass at richardson lucy deconv take quite different parameters. Please bug me to doc these once the filters are split out!

starfish/starfish.py
dganguli: 👍

PR #138, opened 2018-05-07
Tool to convert org.json to the new format.

ttung:


dganguli:
does this belong in the starfish codebase? should it be a part of the starfish CLI? i think the answer to both of those questions is no. 

here's my reasoning: 
1) we only have three partial datasets (partial means 1 FOV) in the old format (ISS.zip, MERFISH.zip, Human vs. Mouse ISS). 

2) we have old code that writes the old org.json for these singular FOVs. with this converter, we can finish writing out ALL the FOVs in those datasets in the new format. 

3) once 2 is done we have no need for this converter. we will write all new datasets in the new format.

4) @ambrosejcarr  has two new datasets, Brian Long's and Simone's. He can use this tool to put Brian's data in the new format. He can simply write Simone's converter to adhere to the new format -- although he's blocked on having the spec be up to date.

ambrosejcarr:
Agree on @dganguli comments. 

Regarding blocking, what I need for my work is for _any_ of the datasets to: 

1. have an updated experiment json (v0.0.0)
2. load into starfish and not break. 

Then I can pattern my work on the other datasets off of that and start integrating things. 

Finally, we will be able to make the notebooks into PR tests.

ttung:
Ok I won't land this, but I'll leave the PR around for a while in case anyone needs the code to convert any existing files.

PR #141, opened 2018-05-08
support 3D image stacks

ambrosejcarr:
- Add support for 3D stacks given current org.json structure.


ttung:
Won't this break downstream code because it assumes a 4D tensor?

PR #144, opened 2018-05-09
Update ISS notebook to match current codebase.

ttung:
Depends on #142 and #143


notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
```diff
@@ -31,7 +31,7 @@
 from starfish.io import Stack
 
 s = Stack()
-s.read('ISS/fov_001/org.json')
+s.read('ISS/fov_001/experiment.json')
```
dganguli: this implies each fov has an experiment.json. i thought we landed on one experiment.json linking out to multiple fov.jsons?

ttung: Your reading is correct.  In the long-term, Stack() will go away, or will shift to something that's encompasses multiple FOVs, and then one would select a FOV to operate on.  This is just a stepping stone.

PR #148, opened 2018-05-10
allen_smFISH notebook [Not for merging]

ambrosejcarr:
- Work in progress. 

@dganguli How do these locations sound for integrating the functions into starfish? 

From the notebook: 

```
Image processing function list:

1. clip & floor (at 10th percentile)
  1. `starfish.transform.threshold` or `.clip`
  2. Also called _after_ bandpass. 
2. trackpy bandpass
  1. `filters.bandpass`
4. gaussian filter (over z) 
  1. `starfish.filters.gaussian`. The skimage function supports nd images, so we just need to adjust the object in starfish. 
5. call peak locations (trackpy locate)
  1. `starfish.starfish.spots.crocker-grier` OR `starfish.starfish.spots.gaussian` (second class in this file)
```

### DIFF
--- a/notebooks/brian_long_smFISH.ipynb
+++ b/notebooks/brian_long_smFISH.ipynb
@@ -17,7 +17,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -32,12 +32,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# package this up same as before\n",
     "experiment_json = os.path.expanduser('./brian_long_smFISH/experiment.json')"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


notebooks/brian_long_smFISH.ipynb
```diff
@@ -0,0 +1,131 @@
...
+    "s.read(experiment_json)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# brian's data is uint, but starfish loads it as floats. Cast it to int here, which does not cause any loss of precision. \n",
```
dganguli: we should enforce starfish loading the data as uint16 -- at least that's what we request in our pipeline spec.

notebooks/brian_long_smFISH.ipynb
```diff
@@ -0,0 +1,131 @@
...
+      ]
+     },
+     "execution_count": 7,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# this doesn't work, do we want this abstraction? \n",
+    "# s.image.dtype"
```
dganguli: actually, what is a s.image? i feel as though image is probably not the right word for whatever it is because i think it's a tensor?

notebooks/brian_long_smFISH.ipynb
```diff
@@ -0,0 +1,131 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Image processing function list:\n",
+    "\n",
+    "1. clip & floor (at 10th percentile)\n",
+    "  1. `starfish.transform.threshold` or `.clip`\n",
```
dganguli: i like clip. i like specifying upper and lower percentiles as a tuple

notebooks/brian_long_smFISH.ipynb
```diff
@@ -0,0 +1,131 @@
...
+    "  2. Also called _after_ bandpass. \n",
+    "2. trackpy bandpass\n",
+    "  1. `filters.bandpass`\n",
```
dganguli: 👍

notebooks/brian_long_smFISH.ipynb
```diff
@@ -0,0 +1,131 @@
...
+    "4. gaussian filter (over z) \n",
+    "  1. `starfish.filters.gaussian`. The skimage function supports nd images, so we just need to adjust the object in starfish. \n",
+    "5. call peak locations (trackpy locate)\n",
+    "  1. `starfish.starfish.spots.crocker-grier` OR `starfish.starfish.spots.gaussian` (second class in this file)"
```
dganguli: i like crocker-grier over second class in gaussian. a more general name would be peakfinder. crocker-grier is more general than peak finding, it actually implies particle tracking, of which one component is finding the peaks.

notebooks/brian_long_smFISH.ipynb
ambrosejcarr: Agree. #149

notebooks/brian_long_smFISH.ipynb
ambrosejcarr: `s.image` is the `ImageStack` (which I think of as the top-level object in our package, akin to the `numpy.ndarray` or `pandas.DataFrame`).

notebooks/brian_long_smFISH.ipynb
ambrosejcarr: Ok. is `peakfinder` specific enough?

notebooks/brian_long_smFISH.ipynb
dganguli: in keeping the code parallel with the language of the spec-- should this be s.hybridization_images? I also find it confusing that there's a class called Stack that has an ImageStack in it ... any good ideas for other names?

notebooks/brian_long_smFISH.ipynb
ambrosejcarr: That sounds good to me.

notebooks/brian_long_smFISH.ipynb
ttung: Keep in mind that Stack will go away in the long term, as operations should be happening on images, with the recipe gluing operations together.

ambrosejcarr: Wouldn't that mean that the Stack will change form from an array to an interface, but some kind of abstraction here will stay in the code base?

ttung: Not really.  Check out https://github.com/spacetx/starfish/blob/268d43a22d0041aa2769963d7ba298ac77595e4e/tests/test_iss_data.py#L75.  This is the model for how things will be done in the future, where nothing really references experiment.json (which is what `Stack` represents) itself.  All commands should reference the data files directly.

dganguli: I don't get it. Let's whiteboard next week! That'll be super fun!

PR #151, opened 2018-05-10
Pin slicedimage 0.0.0-dev14

ttung:


ambrosejcarr:
Ok, so you want to push a tag and pin the tag? Fine with me.

ttung:
> Ok, so you want to push a tag and pin the tag? Fine with me.

That's what I've actually been doing.

PR #154, opened 2018-05-10
osmFISH pipeline functions [Not for merging]

ambrosejcarr:
Similar to the allen_smFISH notebook, the markdown file contains: 
1. an analysis of the parts of osmFISH pipeline
2. a suggestion about where they could fit into starfish
3. where relevant, an enumeration of basic parameters. 

I'm wondering if @dganguli has any suggestions about existing implementations of what I'm calling `calculate_peak_threshold_abs` (see `md` file, point 4 in `Overview of methods needed`)

The notebook is safe to ignore.

### DIFF
--- a/notebooks/simone_pipeline.md
+++ b/notebooks/simone_pipeline.md
@@ -1,10 +1,19 @@
-
-
 # Summary of Simone's pipeline
 
-## `process_standalone_experiment.py`
+## Overview of methods needed
 
-This is primarily about munging files, and loads in `counting.filtering_and_counting` to do the work.
+1. 3d gaussian filter (`skimage.filters.gaussian` or `scipy.ndimage.gaussian`) -> `starfish.filters.gaussian` (already present)
+2. `ndimage.gaussian_laplace` -> `starfish.filters.gaussian_laplace`
+	1. Can we assume that any trimming of the image to fit within the dynamic range should be associated with the function that produces the aberration? E.g. `gaussian_laplace` can produce negative values -- should we wrap that function so that it doesn't have this characteristic?
+  2. If no, we will need some clean-up functions
+3. `max_projection` over z -> `starfish.image._stack` (currently present)
+4. Threshold finding (is there an equivalent library function for the gradient calculation + trimming that this func does? -> `starfish.stats.calculate_peak_threshold_abs`. Replace `calculate` with your favorite word.
+5. `peak_local_max` -> `starfish.spots.peakfinder` (location from brian-long)
+
+## The pipeline:
+
+- First, `process_standalone_experiment.py` is called. 
+- This is primarily about munging files, and loads in `counting.filtering_and_counting` to do the work.
 
 ### Arguments:
 
@@ -15,9 +24,11 @@ This is primarily about munging files, and loads in `counting.filtering_and_coun
 5. min-plane (default None) - allows selection of a plane to start from (exclude out-of-focus)
 6. max-plane (default None) - allows selection of a plane to end at (exclude out-of-focus)
 
-## `counting.filtering_and_counting`:
+- Next, `counting.filtering_and_counting` is called
 
 ### Arguments:
+
+List Simone's, they're well documented. 
 
 ```
 fpath_img_to_filter: str
@@ -46,18 +57,17 @@ skip_tags_counting: list
 list of the tags inside the genes/stainings name to avoid to count.
 ```
 
+Procedure: 
 
-Pipeline:
-
-Expects np.float64 images (how to address this? will everything work?)
+Note: Expects np.float64 images (how to address this? will everything work?)
 
 1. `filtering.nuclei_filtering`
 	1. called if this is in part of skip-genes, designed to skip nuclei, I think
-  2. 3d gaussian filter, sigma (2, 100, 100) (z, y, x?)  # todo check, not explicit
+  2. 3d gaussian filter, sigma (2, 100, 100) (z, y, x) (could be z, x, y), but x <-> y for our purposes.
   3. set negative values to zero
   4. max project over z (np.amax(stack, axis=0)
-2. `filtering.smFISH_filtering`
-  1. Gaussian filter, sigma=(1, 8, 8)
+2. otherwise, call `filtering.smFISH_filtering`
+  1. `gaussian_filter`, sigma=(1, 8, 8)
   2. `gaussian_laplace(stack, (0.2, 0.5, 0.5)`  # todo look up
     3. this makes the peaks negative, so invert signal `img_stack = -img_stack`
     4. ... and zero out the negative values again
@@ -67,7 +77,8 @@ Expects np.float64 images (how to address this? will everything work?)
 
 ### Thr Calculator
 
-This method calculates a threshold to determine if dots are peaks or not.
+This method calculates a threshold to determine if dots are peaks or not. Specifically, it estimates
+threshold_abs for `skimage.feature.peak_local_max`
 
 Arguments:
 1. `min_distance`: if two peaks are not > 5 pixels apart they are the same peak
--- a/notebooks/simone_pipeline.md
+++ b/notebooks/simone_pipeline.md
@@ -1,0 +1,123 @@
+
+
+# Summary of Simone's pipeline
+
+## `process_standalone_experiment.py`
+
+This is primarily about munging files, and loads in `counting.filtering_and_counting` to do the work.
+
+### Arguments:
+
+1. path - path to experiment
+2. analysis name - prefix for experiment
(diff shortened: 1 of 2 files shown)


notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
+# Summary of Simone's pipeline
+
+## Overview of methods needed
+
+1. 3d gaussian filter (`skimage.filters.gaussian` or `scipy.ndimage.gaussian`) -> `starfish.filters.gaussian` (already present)
+2. `ndimage.gaussian_laplace` -> `starfish.filters.gaussian_laplace`
```
dganguli: there's also: http://scikit-image.org/docs/dev/api/skimage.filters.html#skimage.filters.laplace. I *think these both implement a laplacian of gaussian (LoG) filter? can you try both and see if they're actually the same?

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+	1. Can we assume that any trimming of the image to fit within the dynamic range should be associated with the function that produces the aberration? E.g. `gaussian_laplace` can produce negative values -- should we wrap that function so that it doesn't have this characteristic?
```
dganguli: I think it's OK for values to go negative -- if that's the intended consequence of the filter. Negative values make it hard to write out pngs and tiffs, but i don't think they seriously break anything further on down the line (e.g., spot detection, decoding, etc.). 

perhaps, as a start, we simply log a warning if we detect negative values in the image after filtering, then think about what to do if that happens. alternatively, we can specify various flags for what to do if we see funky dynamic range issues. e.g., in opencv, there's often a flag called preserve_range=True/False that handles this case. 

i think starting with the warning first, but keeping these other options in mind is a good start.

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+  2. If no, we will need some clean-up functions
+3. `max_projection` over z -> `starfish.image._stack` (currently present)
+4. Threshold finding. Is there an equivalent library function for the gradient calculation + trimming that this func does? -> `starfish.stats.calculate_peak_threshold_abs`. Replace `calculate` with your favorite word.
```
dganguli: sorry -- i don't understand the question. starfish.stats.calculate_peak_threshold_abs is not a real function yet right? what is it supposed to do?

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+## The pipeline:
+
+- First, `process_standalone_experiment.py` is called. 
+- This is primarily about munging files, and loads in `counting.filtering_and_counting` to do the work.
+
+### Arguments:
+
+1. path - path to experiment
+2. analysis name - prefix for experiment
+3. stringency (default 0) - ? TODO
```
dganguli: what's this?

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Arguments:
+
+List Simone's, they're well documented. 
+
+```
+fpath_img_to_filter: str
+path to the file to process
+filtered_png_img_gene_dirs: list
+list of the paths of the directories where the filtered images as are
```
dganguli: interesting -- this reminds me, can you create an issue for the filter CLI and API to write out pngs? also pinging @ttung -- this is something that came up during spec review that users really wanted. we should be able to just learn from simone's code how he handled this (including what he does about negative values)

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+saved as pngs.
+filtered_img_gene_dirs: list
+list of the paths of the directories where the filtered images are saved
+as .npy.
+counting_gene_dirs: list
+list of the paths of the directories where the countings of the filtered
+images are saved.
+illumination_correction: bool
+if True the illumination correction is run on the dataset.
```
dganguli: i'm curious to learn how he does illumination correction. not high priority since we're punting on it for now.

notebooks/simone_pipeline.md
dganguli: nm i figured it out below -- something about a threshold for calling a spot a spot

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+1. `filtering.nuclei_filtering`
+	1. called if this is in part of skip-genes, designed to skip nuclei, I think
+  2. 3d gaussian filter, sigma (2, 100, 100) (z, y, x) (could be z, x, y), but x <-> y for our purposes.
+  3. set negative values to zero
+  4. max project over z (np.amax(stack, axis=0)
+2. otherwise, call `filtering.smFISH_filtering`
+  1. `gaussian_filter`, sigma=(1, 8, 8)
+  2. `gaussian_laplace(stack, (0.2, 0.5, 0.5)`  # todo look up
+    3. this makes the peaks negative, so invert signal `img_stack = -img_stack`
+    4. ... and zero out the negative values again
```
dganguli: seems like a default parameter in the filters module, e.g., zero_negatives = False, could be flipped if the user wants to zero out negativ evalues

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+2. stringency: higher stringency == higher threshold from `thr_array` to get peaks (tuning parameter)
+
+Returns:
+
+Lots of stuff in the form of a dictionary
+1. `selected_thr`: threshold for counting after application of stringency
+2. `calculated_thr`: threshold after all the trimming is calculated but before stringency application
+3. `selected_peaks`: 2d coordinates of the peaks
+4. `thr_array`: array of 100 points between Img.min() and Img.max(); this is a linspace between min and max
+5. `peaks_coords`: coordinates of peaks at each threshold value.
```
dganguli: are these in units of pixels or coordinates (sample space)?

notebooks/simone_pipeline.md
```diff
@@ -0,0 +1,134 @@
...
+6. `total peaks`: number of peaks at each threshold
+7. `thr_idx`: index of the calculated threshold (some value within Thr index)
+
+Procedure:
+- for each of 100 thresholds between min and max, call peaks with `peak_local_max`
```
dganguli: interesting. i wonder how this procedure compares to simple otsu-thresholding (http://scikit-image.org/docs/dev/auto_examples/xx_applications/plot_thresholding.html) could be something worth asking simone about

notebooks/simone_pipeline.md
dganguli: ah now i understand the question. maybe this lives as an optional thresholding method for the peak local max spot detector, instead of in starfish.stats. thoughts?

notebooks/simone_pipeline.md
ambrosejcarr: > there's often a flag called preserve_range=True/False that handles this case

Got it, I wasn't sure what that referred to. That's helpful. If you happen to know this already, do you know if it rescales such that min(image) becomes zero, or if it typically clips such that `x[x < 0] = 0`? I'm trying to get a sense of the norms. If you don't know I'll look it up.

notebooks/simone_pipeline.md
ambrosejcarr: Yea, that might make sense. We could make it a private method in the same module.

notebooks/simone_pipeline.md
ambrosejcarr: Oops, will fix that TODO.

notebooks/simone_pipeline.md
ambrosejcarr: Simone clips negative values. 

#157

notebooks/simone_pipeline.md
ambrosejcarr: pixels.

ambrosejcarr:
Thanks for the comments @dganguli . I think it would be a good idea to take Simone up on his offer to meet and discuss this after I've dug in and responded to comments. Do you think it would make sense to schedule this sometime mid-late next week?

notebooks/simone_pipeline.md
ambrosejcarr: I think the function you linked is more similar to https://github.com/scipy/scipy/blob/v0.16.1/scipy/ndimage/filters.py#L397 than to https://github.com/scipy/scipy/blob/v0.16.1/scipy/ndimage/filters.py#L413 (what the osmFISH pipeline is using). 

The functions are definitely not the same, but it's possible they're similar enough to use, or that a preprocessing step could make them equivalent. the skimage function is _much_ faster. (3s vs 50s).

PR #155, opened 2018-05-11
tests to verify the get_slice API.

ttung:


ambrosejcarr:
These look good to me. Thanks for writing these. Was helpful to debug my work.

PR #159, opened 2018-05-11
add stack function apply method

ambrosejcarr:
- Adds `Stack.apply` which enables the iterative application of any 2d or 3d function to the data in Stack.numpy_array using the `get_slice` and `set_slice` API. 

- I excluded multiprocessing, but left a comment where the map function could be swapped out, when we have more confidence what our numpy back-end will look like.

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -156,27 +156,27 @@ class ImageStack(ImageBase):
 
         return tuple(slice_list), axes
 
-    def _iter_indices(self, is_3d: bool=False) -> Iterator[Mapping[Indices, int]]:
-        """Iterate over indices of image tiles or image volumes if is_3d is True
+    def _iter_indices(self, is_volume: bool=False) -> Iterator[Mapping[Indices, int]]:
+        """Iterate over indices of image tiles or image volumes if is_volume is True
 
         Parameters
         ----------
-        is_3d, bool
-          if True, yield indices necessary to extract volumes from self, else return
-          indices for tiles
+        is_volume, bool
+            If True, yield indices necessary to extract volumes from self, else return
+            indices for tiles
 
         Yields
         ------
         Dict[str, int]
-          mapping of dimension name to index
+            Mapping of dimension name to index
 
         """
-        for hyb in numpy.arange(self.shape['hyb']):
-            for ch in numpy.arange(self.shape['ch']):
-                if is_3d:
+        for hyb in numpy.arange(self.shape[Indices.HYB]):
+            for ch in numpy.arange(self.shape[Indices.CH]):
+                if is_volume:
                     yield {Indices.HYB: hyb, Indices.CH: ch}
                 else:
-                    for z in numpy.arange(self.shape['z']):
+                    for z in numpy.arange(self.shape[Indices.Z]):
                         yield {Indices.HYB: hyb, Indices.CH: ch, Indices.Z: z}
 
     def _iter_tiles(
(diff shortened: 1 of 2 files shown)


starfish/image/_stack.py
```diff
@@ -1,6 +1,8 @@
 import collections
 import os
 import typing
+from typing import Iterable, Mapping, Iterator, Union
```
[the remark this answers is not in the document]
ambrosejcarr: Agree. I added this in as a subtle suggestion to provoke conversation (It worked! :))

I prefer the way I did it here. My rationale is: 
1. Most important: these typing examples can get really long, particularly in cases with dependent return values (`Union`, `Optional`, etc) which can be annoying to code while considering line length limits. Anything we can do to shorten the signature is good. 
2. Because typing examples are used in prescribed locations, I don't think there is a significant reduction in clarity.

ambrosejcarr:
Thanks for the comments, will fix all. 

We should also standardize our docstrings (I see both google and ReST, and I've written mine here in numpydoc). 

I like numpydoc because it's (1) no less readable than the alternatives (2) what most of the libraries we import use. 

Perhaps @dganguli can chime in about what he thinks would best meet user expectations. I'll sweep the codebase and update the docs + typing based on the conclusions here.

tests/test_apply.py
```diff
@@ -0,0 +1,58 @@
...
+
+def multiply(array, value):
+    return array * value
+
+
+class TestSetSliceAPI(unittest.TestCase):
+    NUM_HYB = 2
+    NUM_CH = 3
+    NUM_Z = 4
+    HEIGHT = 30
```
dganguli: Y, X instead of HEIGHT, WIDTH?

starfish/image/_stack.py
```diff
@@ -155,15 +156,92 @@ def _build_slice_list(
...
+
+        Yields
+        ------
+        Dict[str, int]
+          mapping of dimension name to index
+
+        """
+        for hyb in numpy.arange(self.shape['hyb']):
+            for ch in numpy.arange(self.shape['ch']):
+                if is_3d:
```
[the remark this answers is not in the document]
dganguli: oh i'm sorry -- this does make sense maybe? if it's 3D, you want to return volumes, not 'tiles'. if it is not 3D, then the iteration over z is effectively a noop? if i'm correct, that's sort of confusing -- maybe at the least add a comment?

dganguli:
re: standardization.

for typing -- i have a preference to the less verbose standard. e.g., not having to type typing.* all the time. 

for docstrings -- i don't care as long as:
1. it works with sphinx
2. pycharm can autogenerate docsrings (that's why i use reSt.

if numpy_docs does both then let's do it!

starfish/image/_stack.py
```diff
@@ -155,15 +156,92 @@ def _build_slice_list(
...
+        Yields
+        ------
+        numpy.ndarray
+          numpy array that corresponds to provided indices
+        """
+        for inds in indices:
+            array, axes = self.get_slice(inds)
+            yield array
+
+    def apply(self, func, is_3d=False, **kwargs):
```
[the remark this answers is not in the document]
ambrosejcarr: I think this needs a groupby overhaul, which will solve this problem, but take a but of work.

I'll compile the list of questions this work has raised about the API for tiles and volumes.

starfish/image/_stack.py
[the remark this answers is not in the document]
ambrosejcarr: That's right. I'll add a comment.

starfish/image/_stack.py
```diff
@@ -155,15 +156,92 @@ def _build_slice_list(
 
         return tuple(slice_list), axes
 
+    def _iter_indices(self, is_3d: bool=False) -> Iterator[Mapping[Indices, int]]:
```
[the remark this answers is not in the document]
ambrosejcarr: I like this much more, although I think someone suggested is_3d instead of that. Are we ok with `is_volume` despite some of our image stacks representing focal volumes, not true tissue volumes?

starfish/image/_stack.py
```diff
@@ -155,8 +156,85 @@ def _build_slice_list(
...
+          if True, yield indices necessary to extract volumes from self, else return
+          indices for tiles
+
+        Yields
+        ------
+        Dict[str, int]
+          mapping of dimension name to index
+
+        """
+        for hyb in numpy.arange(self.shape['hyb']):
```
ttung: Indices.HYB

starfish/image/_stack.py
```diff
@@ -155,8 +156,85 @@ def _build_slice_list(
...
+            for ch in numpy.arange(self.shape['ch']):
```
ttung: Indices.CH

starfish/image/_stack.py
```diff
@@ -155,8 +156,85 @@ def _build_slice_list(
...
+            array, axes = self.get_slice(inds)
+            yield array
+
+    def apply(self, func, is_3d=False, **kwargs):
+        """Apply func over all tiles or volumes in self
+
+        Parameters
+        ----------
+        func, Callable
+          function to apply. must expect a first argument which is a 2d or 3d numpy array (see is_3d) and return a
```
ttung: the same return shape is only if inplace is true?

starfish/image/_stack.py
```diff
@@ -155,8 +156,85 @@ def _build_slice_list(
...
+        results = mapfunc(applyfunc, tiles)
+
+        for r, inds in zip(results, indices):
+            self.set_slice(inds, r)
+
+        # TODO: ambrosejcarr implement inplace=False
+
     @property
-    def raw_shape(self) -> Optional[list]:
+    def raw_shape(self) -> Optional[Tuple]:
```
ttung: if you want to be super pedantic, it should be: Optional[Tuple[int]]

tests/test_apply.py
ttung: @ambrosejcarr

PR #160, opened 2018-05-12
implement white tophat and gaussian filters

ambrosejcarr:
- Early PR looking for feedback
- Implements Filter as a pipeline component, utilizing apply functionality. 
- Specifically, implements `white_tophat` (for ISS notebook) and `gaussian` for a generalization test. 

I made this PR to support implementation of the `allen_smFISH` filter methods. 

There are a few `TODO` notes where I'm interested in feedback on implementation or have some questions about what the best way to proceed is.


starfish/filters.py
```diff
@@ -142,17 +115,3 @@ def bin_thresh(img, thresh):
...
-    """
-    Performs white top hat filtering of an image to enhance spots
-    :param img: Image to filter. :type numpy array
-    :param disk_size: Radius of the disk-shaped structuring element.  :type int
-    :return: Filtered image, same shape as input :type ndarray
-    """
-    selem = disk(disk_size)
-    min_filt = minimum_filter(img, footprint=selem)
-    max_filt = maximum_filter(min_filt, footprint=selem)
-    res = img - max_filt
```
dganguli: doesn't this need a res = WhiteTopHatFilter.filter(img, disk_size) ?

dganguli: oh sorry -- looks like you're actually nuking this in favor for the new python api, but are leaving gaussian_high_pass in the old python api?

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -0,0 +1,49 @@
+from skimage.filters import gaussian
+import numpy
+
+from starfish.munge import swap
+from ._base import FilterAlgorithmBase
+
+
+# todo how do we make these methods available in the API?
+# the starfish import is a bit kludgy because it means we need to:
+# starfish.GaussianFilter.filter()
```
dganguli: agree -- i don't love this API

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -0,0 +1,49 @@
...
+class GaussianFilter(FilterAlgorithmBase):
```
dganguli: This is called GaussianFilter, but only does a lowpass filter. I see two options, GaussianFilter does either high-pass or low-pass, and takes in an argument like filter_type. OR, we rename this to GaussianLowPassFilter. We make a new class called GaussianHighPassFilter. I think I prefer the former -- what do you think?

starfish/pipeline/filter/white_tophat.py
```diff
@@ -0,0 +1,34 @@
+from ._base import FilterAlgorithmBase
+
+
+class WhiteTophatFilter(FilterAlgorithmBase):
```
dganguli: I think having Filter in the name is extraneous. WhiteTophatFilter -> WhiteTophat? Same goes for Gaussian filter(s)

starfish/pipeline/filter/_base.py
```diff
@@ -0,0 +1,31 @@
+# todo can we abstract this a single time? It feels unnecessary to have multiple definitions of this class.
```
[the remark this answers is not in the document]
ambrosejcarr: I spoke with @dganguli about this and we think a generic "run" method could replace the fourth (which would be extensively overridden by subclasses). #167

starfish/pipeline/filter/_base.py
```diff
@@ -0,0 +1,31 @@
...
+        https://docs.python.org/3/reference/lexical_analysis.html#identifiers
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        """Adds the arguments for the algorithm."""
+        raise NotImplementedError()
+
+    # @ttung classmethod means we can call this from the API without calling __init__
```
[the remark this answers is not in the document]
ambrosejcarr: It appears to pass the tests despite this method taking parameters, unless I'm misunderstanding.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: Clarified language. Basically, I'm wondering how our API for function calls within starfish should look, and I think the current version has a bit too much indirection. The above `run` command might help simplify things a little bit.

starfish/pipeline/filter/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if args.filter_algorithm_class is None or print_help:
+            cls.filter_group.print_help()
+            cls.filter_group.exit(status=2)
+
+        from starfish.io import Stack
+        import numpy
+        from starfish.image import Indices
+
+        s = Stack()
```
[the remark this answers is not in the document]
ambrosejcarr: Should this be true for all pipeline components, or just images?

starfish/filters.py
```diff
@@ -1,35 +1,8 @@
 import numpy as np
-from scipy.ndimage.filters import maximum_filter, minimum_filter
 from skimage import restoration
-from skimage.filters import gaussian
+from starfish.pipeline.filter.gaussian_low_pass import GaussianLowPass
```
ttung: PEP8 import ordering.

```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

starfish/filters.py
```diff
@@ -38,7 +11,7 @@ def gaussian_high_pass(img, sigma):
     :param sigma: Standard deviation of gaussian kernel :type int
     :return: Filtered image, same shape as input :type ndarray
     """
-    blurred = gaussian_low_pass(img, sigma)
+    blurred = GaussianLowPass.filter(img, sigma)
```
ttung: I believe you want to make it:

`GaussianLowPass(sigma).filter(img)`

starfish/pipeline/filter/__init__.py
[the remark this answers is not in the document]
ttung: It should be.  If you don't want to do this conversion step, that's okay, but it's something we'll want to do eventually.

starfish/pipeline/filter/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+        s.read(args.input)
+        instance = args.filter_algorithm_class.from_cli_args(args)
+        # TODO: ambrosejcarr could use a better way of passing arguments more robust to argument collisions
+        s.image.apply(instance.filter, **vars(args))
+
+        # TODO: ambrosejcarr from here on is a horrible hack that's specific to ISS, and still gets called by
```
ttung: What's the horrible hack?

starfish/pipeline/filter/_base.py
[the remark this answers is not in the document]
ttung: Well, you don't call `FilterAlgorithmBase.filter`, right...?

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -0,0 +1,58 @@
...
+        return "gaussian_low_pass"
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        group_parser.add_argument(
+            "--sigma", default=1, type=int, help="size of morphological masking disk")
+
+    @classmethod
+    def filter(cls, image, sigma, *args, **kwargs):
+        """Applies a gaussian low pass filter to an image. Implementation simply calls skimage.filters.gaussian
```
ttung: btw, what do you think of:

```
"""
Applies aaatsntoehu snoteh
```

vs
```
"""Applies aaatsntoehu snoteh
```

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -0,0 +1,58 @@
...
+        image : np.ndarray
+            input image to be filtered
+        sigma : int
+            standard deviation of gaussian kernel
+        kwargs : dict
+            Not used. Enables passing of arbitrary arguments from the CLI.
+
+        Returns
+        -------
+        blurred : np.ndarray
```
ttung: what's blurred?

starfish/pipeline/filter/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+        from starfish.io import Stack
+        import numpy
+        from starfish.image import Indices
+
+        s = Stack()
+        s.read(args.input)
+        instance = args.filter_algorithm_class.from_cli_args(args)
+        # TODO: ambrosejcarr could use a better way of passing arguments more robust to argument collisions
+        s.image.apply(instance.filter, **vars(args))
```
ttung: can we use this strategy to pass arguments to `from_cli_args`?

ttung: i'm going to try...

starfish/filters.py
ambrosejcarr: Yep, that's consistent with our conversation. Will fix.

starfish/filters.py
ambrosejcarr: > You should put a blank line between each group of imports.

Oops. Will fix.

starfish/pipeline/filter/__init__.py
ambrosejcarr: Any attempt to call a CLI recipe that is not specific to ISS will fail. There is a lot more logic in `_cli` than we want. 

Ideally we'd have 
```
i = instance(parameters)
i.run(stack)
```

And nothing else. _Maybe_ some code to save outputs.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: Apparently both are supported by numpydoc. No strong feelings.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: blurred is the output, of type np.ndarray. (gaussian low pass blurs an image).

starfish/pipeline/filter/__init__.py
ambrosejcarr: Yea I think it should work.

starfish/pipeline/filter/_base.py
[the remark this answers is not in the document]
ambrosejcarr: Well, this can _inherit_ that method and call it. However, given our conversation, and the `instance(parameters).run(stack)` pattern, we'll never call a classmethod. So, this is moot.

starfish/pipeline/filter/gaussian_low_pass.py
ttung: I think this is kind of weird.  It's not like the return value is named 'blurred'.  I would almost prefer that it be a textual explanation for what is being returned.

starfish/pipeline/filter/__init__.py
ttung: It's not free, please see https://github.com/spacetx/starfish/pull/173 for upsides and downsides.

starfish/pipeline/filter/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+from starfish.pipeline.pipelinecomponent import PipelineComponent
+from starfish.util.argparse import FsExistsType
+from . import _base
+from . import gaussian_low_pass
+from . import white_tophat
+
+
+class Filter(PipelineComponent):
+
+    filter_group = None
```
ttung: I wonder if it's possible to type this.

starfish/pipeline/filter/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    def _cli(cls, args, print_help=False):
+        """Runs the filter component based on parsed arguments."""
+
+        if args.filter_algorithm_class is None or print_help:
+            cls.filter_group.print_help()
+            cls.filter_group.exit(status=2)
+
+        s = Stack()
+        s.read(args.input)
+        instance = args.filter_algorithm_class.from_cli_args(args)
```
ttung: need to rebase.

starfish/pipeline/filter/_base.py
```diff
@@ -0,0 +1,34 @@
...
+        https://docs.python.org/3/reference/lexical_analysis.html#identifiers
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        """Adds the arguments for the algorithm."""
+        raise NotImplementedError()
+
+    def filter(self, stack):
```
ttung: Is it possible to type the return value?

starfish/starfish.py
```diff
@@ -7,10 +7,13 @@
 import sys
 from pstats import Stats
 
+import numpy as np
```
ttung: I would discourage you from importing numpy at the top level.  Rationale: it makes starfish --help really really slow.

starfish/starfish.py
```diff
@@ -196,7 +156,14 @@ def segment(args, print_help=False):
     disk_size_markers = None
     disk_size_mask = None
 
-    seg = WatershedSegmenter(s.aux_dict['nuclei'], s.aux_dict[args.aux_image])
+    # make stain for segmentation (TODO ambrosejcarr: move this into watershed)
+    stain = np.mean(s.max_proj(Indices.CH, Indices.Z), axis=0)
+    stain = stain / stain.max()
+
+    # set data
+    s.set_aux('stain', stain)
```
ttung: this is unnecessary, as we don't save the stack.

starfish/pipeline/filter/__init__.py
ambrosejcarr: Yes, `argparse.ArgumentParser` apparently. Will add argparse types in future PR.

PR #169, opened 2018-05-16
unbreak spot detection

ttung:
pandas 0.23 (just released!) does not allow joins across differently typed objects.  melt apparently produces an integer column, but is typed as an object.  use infer_objects() to update the type.


ttung:
see https://github.com/pandas-dev/pandas/pull/18674/files

starfish/spots/gaussian.py
```diff
@@ -74,6 +74,7 @@ def _to_encoder_dataframe(self, bit_map_flag):
 
         res = pd.DataFrame(d)
```
ambrosejcarr: Please add a #TODO ambrosejcarr: explicitly specify data types.

PR #171, opened 2018-05-16
convert all use of typing module to `from typing import XXX`

ttung:


starfish/image/_stack.py
```diff
@@ -156,18 +156,18 @@ def _build_slice_list(
...
         return self._data.shape
 
     @property
-    def shape(self) -> typing.Optional[dict]:
+    def shape(self) -> Optional[dict]:
         if self._data is None:
             return None
 
-        result = collections.OrderedDict()
+        result = collections.OrderedDict()  # type: collections.OrderedDict[Any, str]
```
ambrosejcarr: Are these comments parsed without requiring python 3.6?

ttung: Yes, this is the 3.0+ compatible style.

PR #173, opened 2018-05-17
use **val(args) to call algorithm's constructor

ttung:
Upside: we can get rid of the `from_cli_args` call.  Downside: every algorithm's constructor is required to have `**kwargs`.


ttung:
Restacked on top of #174, and extended to all the pipeline components.

PR #179, opened 2018-05-17
Drop python < 3.6

ttung:
Converted existing _variable_ type annotations to use the new syntax.


ttung:
F-style formatting strings
Less weird Unicode issues
Better type hinting for variables

PR #183, opened 2018-05-17
Move Indices and Coordinates to starfish.constants

ttung:
Test: python -v -m starfish 2>&1 | egrep '^import' yields no fat libraries.


ambrosejcarr:
Is the problem here that importing Indices requires an import of Image? If that's the case, could we move Indices (constants) up to the main level instead of having to import it in weird places?

ttung:
Yes, I agree it would be better if Indices got moved.  However, it so very logically belongs with Image.  Can you think of a better hierarchy that splits ImageStack from Image? :)

ambrosejcarr:
I have no problem with constants being a top-level file. I see it as the place where all the constants for the repo go. We could just add a comment that describes where Indices is used.

ttung:
screen shot 2018-05-18 at 10 12 40 am
well this is disappointing.

PR #188, opened 2018-05-18
use metaclasses to ensure that _ensure_algorithms_setup() is called.

ttung:


ambrosejcarr:
Really really like this. 

There are some examples [here](http://avilpage.com/2017/06/auto-register-subclasess-without-metaclass.html) and [here](https://effectivepython.com/2015/02/02/register-class-existence-with-metaclasses/) that go through creating a class registry. It seems like there might be some idioms for this already that we could try to use here. 

Since `_ensure_algorithm_setup()` is only ever called here, I'd like to move it entirely inside the `metaclass` (perhaps via the registry idea) to make it as clear as possible that the method is only used in this one example.

ttung:
Updated to move the _ensure_algorithm_setup() into the metaclass.

PR #191, opened 2018-05-19
[easy] rename from_image_stack

ttung:
Also add from_url(..) to base class.

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -38,10 +38,10 @@ class ImageStack(ImageBase):
             self.set_slice(indices={Indices.HYB: h, Indices.CH: c, Indices.Z: zlayer}, data=tile.numpy_array)
 
     @classmethod
-    def from_image_stack(cls, image_stack_name_or_url, baseurl):
-        image_partition = Reader.parse_doc(image_stack_name_or_url, baseurl)
+    def from_url(cls, relativeurl, baseurl):
+        image_partition = Reader.parse_doc(relativeurl, baseurl)
 
-        return ImageStack(image_partition)
+        return cls(image_partition)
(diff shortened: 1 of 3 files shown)


starfish/image/_stack.py
```diff
@@ -38,8 +38,8 @@ def __init__(self, image_partition):
             self.set_slice(indices={Indices.HYB: h, Indices.CH: c, Indices.Z: zlayer}, data=tile.numpy_array)
 
     @classmethod
-    def from_image_stack(cls, image_stack_name_or_url, baseurl):
-        image_partition = Reader.parse_doc(image_stack_name_or_url, baseurl)
+    def from_url(cls, relativeurl, baseurl):
```
ambrosejcarr: What's the vision here? Since there's only one implementation, I'd leave this in the `ImageBase` until we have cause to move it to the more specialized class.

starfish/image/_stack.py
```diff
@@ -38,8 +38,8 @@ def __init__(self, image_partition):
...
+        image_partition = Reader.parse_doc(relativeurl, baseurl)
 
         return ImageStack(image_partition)
```
ambrosejcarr: I usually see `return cls(image_partition)` here instead of calling the object's formal name (else, why have an `@classmethod` tag to expose `cls` instead of an `@staticmethod`?)

starfish/image/_stack.py
ttung: the next step is to have a generic `from_url` inside ImageBase that builds a stack or a single image.

starfish/image/_stack.py
ttung: will fix.

PR #192, opened 2018-05-19
Add a single image implementation of ImageBase

ttung:
Depends on #191  and #188


starfish/image/_single.py
```diff
@@ -0,0 +1,48 @@
...
+        backend, name, baseurl = resolve_url(relativeurl, baseurl)
+        extension = os.path.splitext(relativeurl)[1].lstrip(".")
+        format = ImageFormat.find_by_extension(extension)
+        with backend.read_file_handle(name) as fh:
+            return SingleImage(format.reader_func(fh))
+
+    @property
+    def numpy_array(self):
+        data = self._data.view()
+        data.setflags(write=False)
```
ambrosejcarr: I'd like a comment here about what write=False means for how the class works (since the functionality is in an external repo). 

It makes me think changes to the `SingleImage` won't get written back to the backing file.

starfish/image/_single.py
```diff
@@ -0,0 +1,48 @@
...
+
+from ._base import ImageBase
+
+
+class SingleImage(ImageBase):
+    def __init__(self, data):
+        self._data = data
+
+    @classmethod
+    def from_url(cls, relativeurl: str, baseurl: str):
```
ambrosejcarr: This seems like it could be an API entrypoint. Can we get a docstring for this with an example?

starfish/image/_single.py
```diff
@@ -0,0 +1,48 @@
...
+        backend, name, baseurl = resolve_url(relativeurl, baseurl)
+        extension = os.path.splitext(relativeurl)[1].lstrip(".")
+        format = ImageFormat.find_by_extension(extension)
+        with backend.read_file_handle(name) as fh:
+            return SingleImage(format.reader_func(fh))
```
ambrosejcarr:
```python 
return cls(format.reader_func(fh))```

File: starfish/image/_single.py [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+    def numpy_array(self, data):
+        self._data = data.view()
+        data.setflags(write=False)
+
+    @property
+    def raw_shape(self) -> Tuple[int]:
+        return self._data.shape
+
+    @property
+    def shape(self) -> collections.OrderedDict:
```
ambrosejcarr: Does this return an empty dict because there are no `h, c, z` indices for a single image? If so I'd add a docstring that explains why this happens -- I could see it confusing users.

starfish/image/_single.py
```diff
@@ -0,0 +1,48 @@
...
+        return collections.OrderedDict()
+
+    def write(self, filepath):
```
ambrosejcarr: Here and elsewhere (maybe new PR?) I would either change this to `write_npy` or (better) add a parameter that allows the user to dictate the output type and make `.tif` the default. My expectation for a write command is to write the same format that I read (which in this case, is `.tif`, since that's what we read.

starfish/image/_single.py
```diff
@@ -0,0 +1,48 @@
...
+        import numpy
+        numpy.save(filepath, self._data)
+
+    def max_proj(self, *dims):
```
ambrosejcarr: I think that this should emit a warning to make users aware that they're calling a no-op function. It's pretty likely they're misunderstanding the API if they're calling this.

ttung: I'm not sure what the right approach is for this.  I can see the POV where someone calls max_proj on an ImageBase, not knowing if it's an ImageStack or an SingleImage.  In that case, the warning is just weird and annoying.

starfish/image/_single.py
ttung: @dganguli has told me in the past that TIF -> TIF does not work for starfish because the intermediate results are floating point.  is this no longer the case?

starfish/image/_single.py
ttung: #197

starfish/image/_single.py
ttung: #197

starfish/image/_single.py
ambrosejcarr: Isn't `ImageBase` abstract, and therefore this wouldn't happen?

ttung: Imagine that one builds some code to load an experiment, and flatten nuclei (across Z) using max_proj.  If it happens that the experiment is 2D, then the nuclei might just be a SimpleImage, and max_proj would be a noop and also not worthy of a warning.  If the experiment is 3D, then it would be an ImageStack, and max_proj would be meaningful.

Alternatively, we could enforce ban the notion of single standalone images.

PR #193, opened 2018-05-19
switch application code to SingleImage

ttung:
Depends on #192


starfish/io.py
```diff
@@ -38,10 +38,7 @@ def read(self, in_json_path_or_url):
     def _read_aux(self):
         for aux_key, aux_data in self.org['auxiliary_images'].items():
             name_or_url = aux_data['file']
-            img_format = ImageFormat[aux_data['tile_format']]
-            backend, name, _ = resolve_url(name_or_url, self.baseurl)
-            with backend.read_file_handle(name) as fh:
-                self.aux_dict[aux_key] = img_format.reader_func(fh)
+            self.aux_dict[aux_key] = SingleImage.from_url(name_or_url, self.baseurl)
```
ambrosejcarr: I think I'd like this to be `auxiliary_images`

PR #197, opened 2018-05-22
improved documentation for the base class

ttung:


starfish/image/_base.py
```diff
@@ -1,32 +1,77 @@
...
     def raw_shape(self) -> Tuple[int]:
-        """Retrieves the shape of the image data, as a list of the sizes of the indices."""
+        """
+        Returns the shape of the space that this image inhabits.  It does not include the dimensions of the image
+        itself.  For instance, if this is an X-Y image in a C-H-Y-X space, then the shape would include the dimensions C
+        and H.
+
+        Returns
+        -------
+        A list of the sizes of the indices.
```
ambrosejcarr: I think `list` could be confusing here. I like the numpydoc return codes. Here it would be: 
```
Returns
--------
Tuple[int] :
    A list of the sizes of the indices
```

It's potentially worth discussing how we want to do type annotations. The information in the docs is made somewhat irrelevant by having them in the function definitions. I suspect that the IDEs probably prefer one over the other, so there is a chance that we'd introduce errors where they don't line up with one another. 

As a dev, I like the function definitions to be typed. As a user, I think I'd probably prefer the docstrings. 🙃

ttung:
fixed confusing doc

PR #198, opened 2018-05-22
Small fixes to sample experiment generator

ttung:
1. Set the default image type to numpy. This is not really that important since each entry overrides the image type, but it's confusing otherwise.
2. Add extras to the top level experiment doc.


ttung:
> What are extras?

it's an explicitly uncontrolled space for people to store whatever data they want.

PR #202, opened 2018-05-23
add nuclei parameter

ambrosejcarr:
- Different pipelines call the nuclei different things. Adds flexibility for watershed to support this.


ttung:
Does it make sense to standardize the vocabulary?

ambrosejcarr:
Based on what brian said, I think this is safe to close.

PR #203, opened 2018-05-23
lift over residual filters to pipeline components

ambrosejcarr:
- Add `DeconvolvePSF` and `GaussianHighPass` filters (filters used in MERFISH notebooks)
- Drop `bin_erode` and `bin_dilate` which are unused
- Fix docstring in `GaussianLowPass`
- Update docstrings for all modified functions.


starfish/pipeline/filter/gaussian_high_pass.py
```diff
@@ -0,0 +1,75 @@
...
+        self.sigma = sigma
+
+    @classmethod
+    def get_algorithm_name(cls):
+        return "gaussian_high_pass"
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        group_parser.add_argument(
+            "--sigma", default=1, type=int, help="standard deviation of gaussian kernel")
```
ttung: question: does it make sense that there is a default for the CLI but not for the API?

starfish/pipeline/filter/richardson_lucy_deconvolution.py
```diff
@@ -0,0 +1,90 @@
...
+        Note that the default parameters are highly optimized for the MERFISH use case and that num_iter is a
+        very important parameter that requires careful optimization.
+
+        Parameters
+        ----------
+        num_iter : int
+            number of iterations to run
+        sigma : int
+            standard deviation of the gaussian kernel used to construct the point spread function
+        clip : bool (default = False)
```
ttung: This is inconsistent with the function signature.

starfish/pipeline/filter/richardson_lucy_deconvolution.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @classmethod
+    def add_arguments(cls, group_parser):
+        group_parser.add_argument('--num-iter', default=15, type=int, help='number of iterations to run')
+        group_parser.add_argument('--sigma', default=2, type=int, help='standard deviation of gaussian kernel')
+        group_parser.add_argument(
+            '--clip', action='set_true', help='(default False) if True, clip values below -1 and above 1')
+
+    @staticmethod
+    def richardson_lucy_deconv(img: np.ndarray, num_iter: int, psf: np.ndarray, clip=False):
```
ttung: clip: bool=False

ttung: Also you can do:

```
    def richardson_lucy_deconv(img: np.ndarray, num_iter: int, psf: np.ndarray, clip=False) -> np.ndarray:
```

starfish/pipeline/filter/util.py
```diff
@@ -0,0 +1,72 @@
...
+    np.ndarray :
+        Filtered image, same shape as input
+
+    """
+    selem = disk(disk_size)
+    res = binary_opening(img, selem)
+    return res
+
+
+def gaussian_kernel(shape: Tuple[int, int] = (3, 3), sigma=0.5):
```
ttung: PEP8 says it's:

`shape: Tuple[int, int]=(3, 3)`

starfish/pipeline/filter/gaussian_high_pass.py
ambrosejcarr: Nope. This problem may exist in more places. I'll got through and check. There are a few ways to deal with this, None of the ones i'm thinking of are very appealing. 

1. Set the defaults in both places. This can lead to discrepancies with we're not careful. 
2. Set in the API only, and reference the defaults in the text of the CLI. We're still typing things in two places here, although the risk is that the defaults aren't what we say they are, if the CLI changes and the help strings lag behind
3. Set the default values as class properties and reference those in both the API and the CLI (in cli, using string formatting, + references to the defaults)

(3) seems best, but adds a hoop to jump through for new developers... I wonder if we could generate a script to bootstrap a new pipeline component 🙃

starfish/pipeline/filter/richardson_lucy_deconvolution.py
ambrosejcarr: will fix.

PR #211, opened 2018-05-23
[auximg 7/13] [easy] wire up tile opener

ttung:
Missed this earlier.  Now the FOVs generate pretty names.

### DIFF
--- a/examples/build_sample_experiment.py
+++ b/examples/build_sample_experiment.py
@@ -21,7 +21,7 @@ AUX_IMAGE_NAMES = {
 }
 
 
-def tile_opener(toc_path, tile, ext):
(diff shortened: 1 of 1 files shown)


examples/build_sample_experiment.py
```diff
@@ -20,15 +20,15 @@
 }
 
 
-def tile_opener(toc_path, tile, ext):
+def tile_opener(toc_path, tile, file_ext):
     tile_basename = os.path.splitext(toc_path)[0]
     return open(
         "{}-Z{}-H{}-C{}.{}".format(
```
ambrosejcarr: Is this meant to be starfish specific, or just for this experiment? I ask because we haven't specified that tiles need to be saved with specific naming conventions in the spec.

PR #213, opened 2018-05-24
[auximg 9/13] Parameterizable image generation for sample sets.

ttung:
Introduce the notion of an ImageFetcher, whose job is to return a FetchedImage for a given FOV/HYB/CH/Z combination.  The goal is to allow us to reuse the logic in `build_sample_experiment.py` for `get_iss_data.py`

`write_experiment_json()` now accepts ImageFetchers for producing an image.  In the case of the `build_sample_experiment.py` script, these will just be configured with DumbImageFetchers, which fetch random data.

Depends on #210, #212

### DIFF
--- a/examples/build_sample_experiment.py
+++ b/examples/build_sample_experiment.py
@@ -68,7 +68,7 @@ class FetchedImage:
         raise NotImplementedError()
 
 
-class DumbFetchedImage(FetchedImage):
+class RandomNoiseImage(FetchedImage):
     """
     This is a simple implementation of :class:`.FetchedImage` that simply regenerates random data for the image.
     """
@@ -121,7 +121,7 @@ def build_image(fov_count, hyb_count, ch_count, z_count, default_shape=(1536, 10
         for z_ix in range(z_count):
             for hyb_ix in range(hyb_count):
                 for ch_ix in range(ch_count):
-                    image = DumbFetchedImage()
+                    image = RandomNoiseImage()
                     tile = Tile(
                         {
                             Coordinates.X: (0.0, 0.0001),
--- a/examples/build_sample_experiment.py
+++ b/examples/build_sample_experiment.py
@@ -68,6 +68,18 @@ class FetchedImage:
         raise NotImplementedError()
 
 
+class ImageFetcher:
+    """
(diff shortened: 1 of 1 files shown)


examples/build_sample_experiment.py
```diff
@@ -135,7 +145,15 @@ def build_image(fov_count, hyb_count, ch_count, z_count, default_shape=(1536, 10
     return collection
 
 
-def write_experiment_json(path, fov_count, hyb_dimensions, aux_name_to_dimensions):
+def write_experiment_json(
```
ambrosejcarr: Ohhhh, I think I finally get where this is all going. If you have a pile of images with correct names, are you working towards something that will build the `experiment.json` for a user?

examples/build_sample_experiment.py
```diff
@@ -149,14 +167,31 @@ def write_experiment_json(path, fov_count, hyb_dimensions, aux_name_to_dimension
         Dictionary mapping dimension name to dimension size for the hybridization image.
     aux_name_to_dimensions : Mapping[str, Mapping[str, int]]
         Dictionary mapping the auxiliary image type to dictionaries, which map from dimension name to dimension size.
+    hyb_image_fetcher : Optional[ImageFetcher]
```
ambrosejcarr: This and the two following comments could use a bit more verbosity -- What is an ImageFetcher and why would a user pass it? 

Other things I'm wondering here -- it looks like ImageFetcher looks for single images, but I want to build an experiment! so, how does this work?

examples/build_sample_experiment.py
```diff
@@ -88,7 +93,12 @@ def image_data_handle(self) -> IO:
         return output
 
 
-def build_image(fov_count, hyb_count, ch_count, z_count, default_shape=(1536, 1024)):
+class DumbImageFetcher(ImageFetcher):
```
ambrosejcarr: I think maybe `RandomNoiseImageFetcher` would be more descriptive.

examples/build_sample_experiment.py
ttung: Added more prose.

PR #214, opened 2018-05-24
[auximg 10/13] [easy] Move the common code for building an experiment to examples/support

ttung:
This should just be code movement (i.e., no actual logic changes).

Depends on #213


ttung:
This PR was super broken.  Updated to the correct setup now.

PR #215, opened 2018-05-24
[auximg 11/13] move default tile shape to be configured at the experiment level.

ttung:
This allows callers of `build_image(..)` to provide the tile shape.

Depends on #214

### DIFF
--- a/examples/support/__init__.py
+++ b/examples/support/__init__.py
@@ -107,7 +107,7 @@ class RandomNoiseImageFetcher(ImageFetcher):
         return RandomNoiseImage()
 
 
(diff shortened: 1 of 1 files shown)


examples/support/__init__.py
```diff
@@ -173,6 +174,8 @@ def write_experiment_json(
     postprocess_func : Optional[Callable[[dict], dict]]
         If provided, this is called with the experiment document for any postprocessing.  The callable should return
         what is to be written as the experiment document.
+    default_shape : Tuple[int, int]
```
ambrosejcarr: docstring should list the default value for an optional parameter (see [sections point 4](https://numpydoc.readthedocs.io/en/latest/format.html#sections))

ttung: Fixed.

PR #216, opened 2018-05-24
[auximg 12/13] switch aux images to using ImageStack

ttung:
There are two commits that make up this diff.  The first one will not work by itself, because it only changes the write path of get_iss_data.py to use the new examples/support module, and generate the new format.  The tests will obviously break because it expects a single image for each aux img.  The second one changes the consumers to use ImageStack for aux images, and unbreaks the tests.

I have kept the two changes separate for readability purposes, as it's an absurdly large change otherwise. :)

Depends on #208, #215

### DIFF
--- a/examples/get_iss_data.py
+++ b/examples/get_iss_data.py
@@ -1,15 +1,20 @@
 import argparse
+from typing import IO, Tuple
+
 import io
 import os
 import json
 import zipfile
 
 import requests
-from skimage.io import imread, imsave
-from slicedimage import ImageFormat, Tile, TileSet, Writer
+from slicedimage import ImageFormat
 
-from starfish.constants import Coordinates, Indices
+from starfish.constants import Indices
 from starfish.util.argparse import FsExistsType
+from examples.support import FetchedImage, ImageFetcher, write_experiment_json
+
+
+SHAPE = (980, 1330)
 
 
 def download(input_dir, url):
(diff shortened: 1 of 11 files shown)


examples/get_iss_data.py
```diff
@@ -133,14 +80,44 @@ def format_data(input_dir, output_dir, d):
...
+            },
+            'dots': {
+                Indices.HYB: 1,
+                Indices.CH: 1,
+                Indices.Z: 1,
+            }
+        },
+        hyb_image_fetcher=HybridizationImageFetcher(input_dir),
+        aux_image_fetcher={
+            'nuclei': AuxImageFetcher(os.path.join(input_dir, "DO", "c1.TIF")),
```
ambrosejcarr: Could you use keyword arguments here? Might make these parameter choices clearer to someone reading the code.

examples/get_iss_data.py
```diff
@@ -133,14 +80,44 @@ def format_data(input_dir, output_dir, d):
...
+        experiment_json_doc['codebook'] = "codebook.json"
+
+        # TODO: (ttung) remove the following unholy hacks.  this is because we want to point at a tileset rather than
+        # a collection.
+        experiment_json_doc['hybridization_images'] = "hybridization-fov_000.json"
+        experiment_json_doc['auxiliary_images']['nuclei'] = "nuclei-fov_000.json"
+        experiment_json_doc['auxiliary_images']['dots'] = "dots-fov_000.json"
+        return experiment_json_doc
+
+    write_experiment_json(
```
ambrosejcarr: This could use a comment I think, I don't understand why you're using these numbers (4, 4, 1). I can maybe understand the nuclei and dots images (although as deep mentioned, sometimes we'll have nuclei images across the whole z-plane).

starfish/io.py
```diff
@@ -79,13 +72,40 @@ def set_aux(self, key, img):
...
                 raise AttributeError(msg)
+            self.aux_dict[key].numpy_array = img
         else:
-            self.org['auxiliary_images'][key] = {
-                'file': key,
-                'tile_format': ImageFormat.NUMPY.name,
-                'tile_shape': img.shape,
-            }
-        self.aux_dict[key] = img
+            # TODO: (ttung) major hack alert
```
ambrosejcarr: Second comment to explain the hack, please :)

starfish/pipeline/filter/bandpass.py
```diff
@@ -68,4 +68,5 @@ def filter(self, stack) -> None:
 
         # apply to aux dict too:
         for k, val in stack.aux_dict.items():
-            stack.aux_dict[k] = self.bandpass(val, self.lshort, self.llong, self.threshold, self.truncate)
+            stack.aux_dict[k].numpy_array = self.bandpass(
```
ambrosejcarr: Ah cool, these filters can all be way simpler now. Since the entries in `aux_dict` are all `ImageStack` objects all you need to do is: 
```python
for image_stack in stack.aux_dict.values():
    image_stack.apply(bandpass_)
```

starfish/pipeline/filter/clip.py
```diff
@@ -74,4 +74,4 @@ def filter(self, stack) -> None:
 
         # apply to aux dict too:
         for k, val in stack.aux_dict.items():
-            stack.aux_dict[k] = self.clip(val, self.p_min, self.p_max)
+            stack.aux_dict[k].numpy_array = self.clip(val.numpy_array, self.p_min, self.p_max)
```
ambrosejcarr: same as bandpass.

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -70,4 +70,4 @@ def filter(self, stack) -> None:
 
         # apply to aux dict too:
         for k, val in stack.aux_dict.items():
-            stack.aux_dict[k] = self.low_pass(val, self.sigma)
+            stack.aux_dict[k].numpy_array = self.low_pass(val.numpy_array, self.sigma)
```
ambrosejcarr: same as bandpass

starfish/pipeline/filter/white_tophat.py
```diff
@@ -52,7 +52,6 @@ def white_tophat(image):
 
         stack.image.apply(white_tophat)
 
-        # TODO ambrosejcarr: port aux functionality into apply, I think we always want this
-        # apply to aux dict too:
-        for k, val in stack.aux_dict.items():
-            stack.aux_dict[k] = white_tophat(val)
+        # apply to aux dict too.
+        for aux_img in stack.aux_dict.values():
```
ambrosejcarr: Ah, you figured it out yourself. :)

examples/get_iss_data.py
ttung: will do.

examples/get_iss_data.py
ambrosejcarr: Yea, I meant:
```
'nuclei': AuxImageFetcher(os.path.join(
    input_dir, 
    whatever_this_param_is="DO", 
    i_am_probably_an_informative_param_name_too="c1.TIF")
), ...
```

ambrosejcarr: Man when did I do this PR? I must have been half asleep. Please ignore. I somehow thought that DO and c1.TIF were arguments to `AuxImageFetcher`, not paths to join.

ambrosejcarr: 🤦‍♂️

PR #217, opened 2018-05-24
Ajc vis z stack

ambrosejcarr:
- Adds a function adapted from the imagexd tutorials to visualize a z-stack inside the ipython notebook. 
- Adds REQUIREMENTS-NOTEBOOK.txt because ipywidgets is a weird thing to require for a pipeline. Open to comments here. 

Another possible spot for this function is in `Stack.image` - I'm agnostic about where it goes.


starfish/io.py
```diff
@@ -123,3 +124,65 @@ def un_squeeze(self, stack):
         new_shape = (self.image.num_hybs, self.image.num_chs, self.image.num_zlayers) + self.image.tile_shape
         res = stack.reshape(new_shape)
         return res
+
+    def show_z_stack(
+            self, h: int=1, c: int=0, cmap: str='gray', rescale: bool=False, figsize: Tuple[int, int]=(10, 10)):
```
ttung: do you want those defaults for h & c?

starfish/io.py
```diff
@@ -123,3 +124,65 @@ def un_squeeze(self, stack):
...
+            size of the figure in inches
+
+        Notes
+        -----
+        For this widget to function interactively in the notebook, after ipywidgets has been installed, the user must
+        register the widget with jupyter by typing the following command into the terminal:
+        jupyter nbextension enable --py widgetsnbextension
+
+        """
+        from skimage import exposure
```
ttung: Because we're now type-annotating, we can't really avoid importing the heavy libraries at load time.  There's no advantage to doing these lazy imports any more.

ambrosejcarr:
I made the changes you suggested and realized this could benefit from being a bit more general. I updated it using the get_stack API to pull out an arbitrary volume so users can page across hyb/channel in addition to z. Putting this back up for review again.

starfish/image/_stack.py
```diff
@@ -131,6 +133,82 @@ def set_slice(
         self._data[slice_list] = data
         self._data_needs_writeback = True
 
+    def show_stack(
+            self, indices: Mapping[Indices, Union[int, slice]], cmap: str='gray', rescale: bool=False,
```
ttung: My personal newlining strategy is (in order of preference):
1. one line for parameters if it fits.  
2. split by logical groups.  this is a bit subjective, but in this case, self and indices might belong on one line, and the rest on the second line.
3. one line per parameter. (reduces the amount of lines highlighted when parameters are changed.

Feel free to disregard if you feel strongly otherwise.

ttung: Also, unless cmap and figsize are very standard acronyms, I would go with color_map and figure_size.

starfish/io.py
```diff
@@ -1,5 +1,6 @@
 import json
 import os
+from typing import Tuple
```
ttung: This should be removed.

PR #221, opened 2018-05-26
Ajc fix merfish notebook

ambrosejcarr:
This PR Includes a few fixes necessary to recover the functionality of the MERFISH notebook after `bit` information was eliminated from the spec. 

- update MERFISH data to new spec
- upload new-spec MERFISH data to cloud and change notebook to localize it. 
- remove more vestigial `ind` and `val` in favor of `barcode_index` and `intensity` (per the spec)
- eliminate squeeze_map in favor of exposing a `tile_metadata` property on stack that produces the same output and doesn't require one to run `squeeze()` for the side effect. Added bonus: it exposes the data in `image._image_partition.tiles() extras`. 
- `squeeze(bit_map_flag=True)` used to mean "use the `bit` information in the old `org.json` to determine which Tile corresponds to which bit of a barcode". This is now True by default if `barcode_index` is in extras. Eventually this will be stored in the codebook. If this information is not present, the assay is probably single-plex (smFISH) and doesn't leverage this information. If there is no barcode index, this field is just filled sequentially to mirror how squeeze map used to work, but this functionality should be eliminated in future PRs in favor of exposing this information exclusively in either `extras` or the `codebook`. 

Sorry for the big diff.


notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -33,10 +33,10 @@
 
 # load in current directory so we can also grab the benchmark results later
 s = Stack()
-s.read('MERFISH/fov_001/org.json')
+s.read('http://czi.starfish.data.public.s3-website-us-east-1.amazonaws.com/MERFISH/fov_001/experiment.json')
```
ttung: Now that things are stabilized, you may want to move this to cloudfront.

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -33,10 +33,10 @@
...
 
 # data from one FOV correspond to 16 single plane images as shown here (see below for details)
-tile(s.squeeze());  
+# tile(s.squeeze());
```
ttung: just remove this?

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -63,7 +63,8 @@
 # EPY: END markdown
 
 # EPY: START code
-codebook = pd.read_csv('MERFISH/codebook.csv', dtype={'barcode': object})
+# TODO ambrosejcarr: update this to point at the s3 bucket
```
ttung: remove this TODO?

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -110,24 +111,31 @@
...
+    s.image.set_slice(indices, scaled)
 # EPY: END code
 
 # EPY: START code
-tile_lims(stack_blurred, 2, size=10);
+from scipy.stats import scoreatpercentile
+# EPY: END code
+
+# EPY: START code
+mp = s.max_proj('h', 'c', 'z')
```
ttung: Indices.HYB, Indices.CH, Indices.Z ?

ambrosejcarr: Good catch -- my tendency is to try to avoid our internals in notebooks, but this is important to keep them up to date. Will fix.

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
ambrosejcarr: Will fix so that it points at the bucket. 👍

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -30,15 +30,25 @@
 # EPY: ESCAPE %matplotlib inline
 
 from starfish.io import Stack
+from starfish.constants import Indices
```
ttung: sort pls.

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
```diff
@@ -43,9 +43,21 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
```
ttung: LOL what's this?

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -110,24 +120,31 @@
...
+# at least we can do it with get_slice and set_slice right now.
+
+for indices in s.image._iter_indices():
+    data = s.image.get_slice(indices)[0]
+    scaled = data / scale_factors[indices[Indices.HYB], indices[Indices.CH]]
+    s.image.set_slice(indices, scaled)
+# EPY: END code
+
+# EPY: START code
+from scipy.stats import scoreatpercentile
```
ttung: this is an odd snippet

starfish/io.py
```diff
@@ -94,33 +92,46 @@ def set_aux(self, key, img):
...
 
-    def squeeze(self, bit_map_flag=False):
-        first_dim = self.image.num_hybs * self.image.num_chs * self.image.num_zlayers
+    def squeeze(self) -> np.ndarray:
+        """return an array that is linear over categorical dimensions and z
+
+        Returns
+        -------
+        np.ndarray :
+            array of shape (num_hybs + num_channels + num_z_layers, x, y).
```
ttung: i think @dganguli  said we're doing hyb,ch,z,y,x.

or should be...?

starfish/io.py
```diff
@@ -94,33 +92,46 @@ def set_aux(self, key, img):
...
+            extras field for each tile in hybridization.json
+
+        """
+
+        data: defaultdict = defaultdict(list)
+        for tile in self.image._image_partition.tiles():
+            for k, v in tile.indices.items():
+                data[k].append(v)
+            for k, v in tile.extras.items():
+                data[k].append(v)
```
ttung: this is kind of weird in that if extras has the same key as indices, you'll mask data.

possibly worth it to consider raising an error in that situation.

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
ambrosejcarr: Oops. Will fix. :)

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
ambrosejcarr: Could you explain what part you find odd?

starfish/io.py
ambrosejcarr: Yep, that's what the `ImageStack.numpy_array` object is/should be. This is a weird function that I hope will go away as a result of #226.

starfish/io.py
ambrosejcarr: I didn't think about that. Will fix.

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
ttung: it's a cell with one line, and it's an import.  unless i missed something.

ambrosejcarr: Got it, I'll move it to the imports section. I was wondering if you were talking about the code I added.

PR #223, opened 2018-05-27
Ajc doc gather

ambrosejcarr:
- Add docstring to `munge.gather`
- Rename `gather -> melt` since that's what it's calling
- Clarify how args relate to `pd.melt`, that it wraps. 
- Use kwargs when calling melt to clarify outcome of calling the function. 

N.B.  `starfish/spots/gaussian.py` is unchanged because it's vestigial and removed in #222

### DIFF
--- a/starfish/munge.py
+++ b/starfish/munge.py
@@ -1,3 +1,5 @@
+from typing import Any, Iterable
+
 import numpy as np
 import pandas as pd
 import regional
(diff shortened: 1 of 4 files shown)


starfish/munge.py
```diff
@@ -30,12 +32,64 @@ def scale(stack, metric, clip=False):
     return list_to_stack(res)
 
 
-def gather(df, key, value, cols):
-    id_vars = [col for col in df.columns if col not in cols]
-    id_values = cols
-    var_name = key
-    value_name = value
-    return pd.melt(df, id_vars, id_values, var_name, value_name)
+def melt(df: pd.DataFrame, new_index_name: Any, new_value_name: Any, melt_columns: Iterable) -> pd.DataFrame:
```
ttung: melt_columns is not Iterable[str]?

(i actually don't know, and am too lazy to check)

starfish/munge.py
```diff
@@ -30,12 +32,64 @@ def scale(stack, metric, clip=False):
...
+
+    Columns in `melt_columns` are aggregated into a new column. Their identifiers are stored together in a column
+    whose index is `new_index_name`, and their values are stored together in a column whose name is `new_value_name`
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        The DataFrame to melt
+    new_index_name : Any
+        The name of the index column that will be created (typically a str)
```
ttung: weird, people have non-str column names?

starfish/munge.py
```diff
@@ -30,12 +32,64 @@ def scale(stack, metric, clip=False):
...
+
+    melted = pd.melt(
+        df,
+        id_vars=id_vars,
+        value_vars=melt_columns,
+        var_name=new_index_name,
+        value_name=new_value_name
+    )
+
+    # melt does not know the dtypes of the new columns, but pandas can normally guess them
```
ttung: so there isn't a better solution than this?

ambrosejcarr: I haven't been able to find one with my feeble googling. I did at least move it inside the function so the problem is bounded to this one function. 

Related: My RFC would blow away this function.

starfish/munge.py
ambrosejcarr: Yea, sometimes people will use integers. Pandas supports anything.

PR #224, opened 2018-05-28
Ajc pixel decoder pipeline component

ambrosejcarr:
- Exchange pixel encoder and decoder for a single pipeline component
- A bit more work will be needed in future PRs to exchange the current outputs for ones that match our specifications.


starfish/pipeline/features/pixels/__init__.py
```diff
@@ -0,0 +1,64 @@
+import argparse
+import os
+
+from starfish.io import Stack
+from starfish.pipeline.pipelinecomponent import PipelineComponent
+from starfish.util.argparse import FsExistsType
+from starfish.pipeline.features.pixels import _base
+
+
+class PixelSpotDetector(PipelineComponent):
```
ttung: Why this name?  It's not a pixel spot that's being detected, if I understand things correctly.  Why not just SpotDetector?

starfish/pipeline/features/pixels/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        if args.pixel_finder_algorithm_class is None or print_help:
+            cls.pixel_finder_group.print_help()
+            cls.pixel_finder_group.exit(status=2)
+
+        print('Detecting Pixels...')
+        s = Stack()
+        s.read(args.input)
+        instance = args.pixel_finder_algorithm_class(**vars(args))
+        # TODO does this work?
+        pixel_attributes, encoded_pixels = instance.find(s)
```
ttung: shouldn't this be returning spots_attributes, spots?

starfish/pipeline/features/pixels/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        path = os.path.join(args.output, 'pixels.geojson')
+        print(f"Writing | pixels geojson to: {path}")
+        pixel_attributes.save_geojson(path)
+
+        path = os.path.join(args.output, 'pixels.json')
+        print(f"Writing | spot_id | x | y | z | to: {path}")
+        pixel_attributes.save(path)
+
+        path = os.path.join(args.output, 'encoder_table.json')
+        print(f"Writing | spot_id | hyb | ch | val | to: {path}")
```
ttung: is this correct?

starfish/pipeline/features/pixels/__init__.py
ambrosejcarr: SpotDetector is a bit too general (we have a bunch of types)
I thought about this a bit. 

It could be `ConnectedPixelsSpotDetector` ... I suspect I'll have to think on this for a while.

starfish/pipeline/features/pixels/__init__.py
ambrosejcarr: It's not! I'll fix that in another PR, it affects a few more algorithms.

starfish/pipeline/features/pixels/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+
+from starfish.io import Stack
+from starfish.pipeline.pipelinecomponent import PipelineComponent
+from starfish.util.argparse import FsExistsType
+from starfish.pipeline.features.pixels import _base
+
+
+class PixelSpotDetector(PipelineComponent):
+
```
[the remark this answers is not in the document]
ambrosejcarr: Created #233 . This requires a bit of thought and affects most of the classes that subclass PipelineComponent. I'll do this in a future PR.

Correct me if I'm wrong, but most users won't interact with the objects in `__init__.py`, so these are primarily docs for developers, right? 

Generally, I've been lax on class docs though, so thank you for the reminder.

starfish/pipeline/features/pixels/__init__.py
ambrosejcarr: Yes. There are two steps here: 
1. Make pipeline component
2. Get outputs to be to-spec. 

I'm only addressing (1) in this PR, (2) depends on #226

starfish/pipeline/features/pixels/__init__.py
ambrosejcarr: It's also possible we will be able to merge the `pixels` and `spots` sub-packages and move everything up into features as a result of #226

starfish/pipeline/features/pixels/__init__.py
```diff
@@ -0,0 +1,64 @@
+import argparse
+import os
+
+from starfish.io import Stack
```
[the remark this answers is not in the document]
ambrosejcarr: Done. Thanks.

starfish/pipeline/features/pixels/_base.py
```diff
@@ -0,0 +1,7 @@
+from starfish.pipeline.algorithmbase import AlgorithmBase
+
+
+class PixelFinderAlgorithmBase(AlgorithmBase):
+    def find(self, stack):  # -> Tuple[SpotAttributes, EncodedSpots]:
```
ttung: Does this not work as a return type?

ambrosejcarr: Nope, needs #226, this does not return EncodedSpots or true SpotAttributes (although it returns something _like_ SpotAttributes). I've removed the comment.

PR #225, opened 2018-05-28
add classmethod constructor for Stack

ambrosejcarr:
- Simple PR to make stack construction a single line. 
- Later PRs can add error-handling related to `slicedimage` to this class. 
- Snuck in a spelling correction.

### DIFF
--- a/starfish/io.py
+++ b/starfish/io.py
@@ -18,7 +18,7 @@ class Stack:
         self.org = None
         self.image = None
 
-        # auxilary images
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
My thought process was: 
1. If Stack gets renamed to ImageDataSet or something similar eventually, then I'll feel happy to have this constructor. 
2. If it gets blown away, it was only a few lines of code I had to write :)

PR #228, opened 2018-05-30
RFC: Updated Contributing.md

ambrosejcarr:
What's still missing from this doc?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,11 +10,14 @@ A typical starfish run consists of running one or more image processing filter s
 The identified features are then decoded into the genes that they correspond to by mapping the fluorescence channel (and optionally hybridization round) using a codebook.
 Finally, the filtered data are segmented, identifying which cell each feature belongs to.
 
-Implementing a new module of an existing `pipeline_component` is a simple process. For example, to add another image filtering step, one would:
+# Creating a new algorithm for an existing `pipeline_component`:
+
+For example, to add a new image filter, one would:
 
 1. Create a new python file `new_filter.py` in the `starfish/pipeline/filter/` directory.
 
-2. Create a new python class `NewFilter(PipelineComponent)` in the `new_filter.py` class.
+2. Find the corresponding `AlgorithmBase` for your component. For filters, this is `FilterAlgorithmBase`, which is found in `starfish/pipeline/filter/_base.py`. 
+Import that base into `new_filter.py`, and have your new algorithm subclass it, e.g. create `class NewFilter(FilterAlgorithmBase)`
 
 3. Implement all required methods from the base class.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,18 +2,49 @@
...
+2. Create a new python class `NewFilter(PipelineComponent)` in the `new_filter.py` class.
+
+3. Implement all required methods from the base class.
+
+4. For the command line arguments to mimic the API, the arguments passed to the CLI must exactly match the names of the parameters for `NewFilter.__init__`, after dashes are automatically converted by argparse.
+For example, `--foo-bar` would convert to `foo_bar` and init must accept such an argument: `NewFilter.__init__(foo_bar, ..., **kwargs)`
+
+5. `NewFilter.__init__()` must have a `**kwargs` parameter to accept arbitrary CLI args.
+
+That's it! your pipeline component will automatically register and be available under `starfish filter` in the CLI.
```
ttung: I don't think this is true.  What you're describing is the outcome of a new FilterAlgorithm, not a new PipelineComponent.  Wiring up a new PipelineComponent is a bit more complicated...

ambrosejcarr: Ah true, I'll clean up this language. 

I _am_ describing a new Algorithm, and I made that decision deliberately because (1) I don't think we've missed many obvious pipeline components (2) I hope a user would probably contact us before writing a new pipeline component and (3) not describing how to wire up a new component will bias a user towards working in the existing workflow. 

Does this logic make sense to you?

CONTRIBUTING.md
```diff
@@ -2,18 +2,49 @@
...
-# Code of conduct
+# How can I contribute?
+Starfish is designed to specify pipeline recipes for image processing. To support this use, the library is composed as a series of `pipeline_component` modules.
+The objects that sub-class `pipeline_component` spawn a command-line interface that should identically track the API of the python library.
 
+A typical starfish run consists of running one or more image processing filter stages, and identifying features through either a spot- or pixel-based approach.
+The identified features are then decoded into the genes that they correspond to by mapping the fluorescence channel (and optionally hybridization round) using a codebook.
+Finally, the filtered data are segmented, identifying which cell each feature belongs to.
+
+Implementing a new module of an existing `pipeline_component` is a simple process. For example, to add another image filtering step, one would:
```
ttung: I see that the terminology of "module" came from the spec.  I would argue that we should ditch the term "module" as it is ridiculously overloaded.

I would just rephrase this as "Creating a new algorithm for an existing \`pipeline_component\` ..."

ttung: Also, "For example, to add a new image filter, one would:"

CONTRIBUTING.md
```diff
@@ -2,18 +2,49 @@
...
+
+1. Create a new python file `new_filter.py` in the `starfish/pipeline/filter/` directory.
+
+2. Create a new python class `NewFilter(PipelineComponent)` in the `new_filter.py` class.
```
ttung: This, I believe, is incorrect.  You would want to subclass `FilterAlgorithmBase` or something akin to that.

CONTRIBUTING.md
```diff
@@ -2,18 +2,49 @@
...
+
+3. Implement all required methods from the base class.
+
+4. For the command line arguments to mimic the API, the arguments passed to the CLI must exactly match the names of the parameters for `NewFilter.__init__`, after dashes are automatically converted by argparse.
+For example, `--foo-bar` would convert to `foo_bar` and init must accept such an argument: `NewFilter.__init__(foo_bar, ..., **kwargs)`
+
+5. `NewFilter.__init__()` must have a `**kwargs` parameter to accept arbitrary CLI args.
+
```
ttung: starfish/pipeline/filter/__init__.py needs to import the new filter as well.

ambrosejcarr: Ah, good catch.

PR #199, opened 2018-05-22
add type hints

ambrosejcarr:
- Fixes some minor problems with my previous PRs related to typing
- Adds hints for uninitialized argparse.ArgumentParser objects to PipelineComponents
- reassign `id` -> `id_` to avoid clobbering builtin `id` method
- Assigned an errant TODO

PR #230, opened 2018-05-30
[not for merging] Fix NotImplemented property on ValidatedTable

ambrosejcarr:
I thought this might be a good approach to fix @ttung 's comment [here](https://github.com/spacetx/starfish/pull/199#discussion_r191898497), but I get this weird error about Callables. :\

$ make mypy
mypy --ignore-missing-imports starfish tests examples
starfish/pipeline/features/_validated_table.py:72: error: Argument "required_fields" to "ValidatedTable" has incompatible type "Callable[[ValidatedTable], Set[Any]]"; expected "Set[Any]"
make: *** [mypy] Error 1

### DIFF
--- a/starfish/pipeline/features/_validated_table.py
+++ b/starfish/pipeline/features/_validated_table.py
@@ -2,9 +2,6 @@ import pandas as pd
 
 
 class ValidatedTable:
-
-    required_fields = NotImplemented
-
     """
     This base class defines common methods for the json outputs of the starfish package, each of which is stored
     internally as a pandas dataframe.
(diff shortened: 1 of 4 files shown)


ttung:
It's not possible afaik to do a @property on a classmethod.  You could just make it an abstract method.

ambrosejcarr:
Ah, that's what I had forgotten. Thanks.

ambrosejcarr:
I like your existing solution better than making it an abstract method. I don't like mixing `collections.abc` with  `NotimplementedError`, and prefer the latter.

Issue #231, opened 2018-06-01
Support verbosity + 3d in all apply loops


ambrosejcarr:
Change the baseclasses to enable pipeline components to receive the following boolean flags:
- [ ] `verbose`
- [ ] `is_volume`

ambrosejcarr:
Closed by #365

PR #232, opened 2018-06-01
Ajc local max peak finder

ambrosejcarr:
This PR is a work-in-progress which I'm putting this up because I'd **like comments on the notebook** prior to giving it to Brian. Please don't feel the need to review the code. 

In terms of code and functionality changes, this PR:

1. Adds an `allen_smFISH` notebook that mimics the allen_smFISH pipeline through peak finding. It's final output is to produce `SpotAttributes` tables for each channel that match the specification. It does not currently decode, although this would be easy to write a simple function for it, since smFISH has simple decoding. 
2. Puts the `allen_smFISH` data onto s3 and fetches it from the notebook
3. Adds a `LocalMaxPeakFinder`
4. Adds 3d support for low-pass gaussian filtering. 
5. Adds a verbose tag to apply for all functions used in `allen_smFISH` that provides updates on how much of the apply loop is complete
6. Adds the ability to visualize called spots on top of the pageable `show_stack` output. The spots show on the relevant z-plane, since this pipeline does spot calling in 3d. Note that the widget does not load in github, you'll need to follow the (updated) installation instructions for the widget, then run the notebook locally to see how that works. 
7. Adds missing documentation to many objects that are exposed in the notebook. 
8. Fixes a few dumb untested errors that I introduced when trying to make mypy happy. 

Things that should be improved: 
- This PR cuts a few corners on type hinting and currently fails mypy.
- Working through the notebook shows that I need better support for user-defined dynamic range adjustment (see the TODO in the notebook, I'll do this tomorrow, it should not be a hard change). 
- I thought this would be less work, and therefore didn't think ahead and parcel out the commits. As a result, there's a big diff here. Sorry Tony. 

@ttung please review the documentation I added to `starfish.image._stack.ImageStack`. I am not 100% confident everything there is accurate.


notebooks/allen_smFISH.py
```diff
@@ -0,0 +1,201 @@
...
+# We're ready now to load the experiment into starfish (This experiment is big, it takes a few minutes):
+# EPY: END markdown
+
+# EPY: START code
+s = Stack()
+s.read(experiment_json)
+
+# The allen's data is uint, but starfish loads it as floats. Cast it directly to uint16, which does
+# not cause any loss of precision. 
+s.image._data = s.image._data.astype(np.uint16)
```
ttung: i don't think this is the case any more.

notebooks/allen_smFISH.py
```diff
@@ -0,0 +1,201 @@
...
+# EPY: START markdown
+# Below, because spot finding is so slow when single-plex, we'll pilot this on a max projection to show that the parameters work. Here's what trackpy.locate, which we wrap, produces for a z-projection of channel 1. To do use our plotting methods on z-projections we have to expose some of the starfish internals, which will be improved upon. 
+# EPY: END markdown
+
+# EPY: START code
+from showit import image
+from trackpy import locate
+
+# grab a section from the tensor. 
+ch1 = s.max_proj(('z'))[0, 1]
```
ttung: Is it intentional that you're not using `Indices`?

starfish/image/_base.py
```diff
@@ -70,7 +70,7 @@ def write(self, filepath: str) -> None:
         """
         raise NotImplementedError()
 
-    def max_proj(self, *dims: Sequence[Indices]) -> ndarray:
+    def max_proj(self, *dims: Indices) -> ndarray:
```
ttung: thanks.

starfish/image/_stack.py
```diff
@@ -446,7 +548,20 @@ def tile_opener(tileset_path, tile, ext):
             pretty=True,
             tile_opener=tile_opener)
 
-    def max_proj(self, *dims):
+    def max_proj(self, *dims: Indices) -> numpy.ndarray:
+        """return a max projection over one or more axis of the image tensor
+
+        Parameters
+        ----------
+        dims : Tuple[Indices]
```
ttung: is this right?  Is it considered a tuple?

starfish/pipeline/features/spots/detector/local_max_peak_finder.py
```diff
@@ -0,0 +1,143 @@
...
+        self.is_volume = is_volume
+        self.verbose = verbose
+
+    # # TODO ambrosejcarr: make this generalize to smFISH methods
+    # def encode(self, spot_attributes: SpotAttributes):
+    #     spot_table = spot_attributes.data
+    #     spot_table['barcode_index'] = np.ones(spot_table.shape[0])
+
+    def find_attributes(self, image):
+        attributes = locate(
```
ttung: does this require image to be a specific format?

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -1,24 +1,30 @@
...
 class GaussianLowPass(FilterAlgorithmBase):
 
-    def __init__(self, sigma, **kwargs) -> None:
+    def __init__(self, sigma, is_volume: bool=False, verbose=False, **kwargs) -> None:
         """Multi-dimensional low-pass gaussian filter.
 
         Parameters
         ----------
-        sigma : float
+        sigma : Union[float, Tuple[float]]
```
ttung: type can go into the params?

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -1,24 +1,30 @@
...
             Standard deviation for Gaussian kernel.
+        is_volume : bool
+            If True, 3d (z, y, x) volumes will be passed to self.low_pass. Otherwise, 2d tiles will be passed
```
ttung: very confused here as `low_pass` does not seem to accept an is_volume parameter.

starfish/viz.py
```diff
@@ -1,10 +1,10 @@
 import numpy as np
-from showit import image, tile
 
 from .stats import im_describe, stack_describe
 
 
 def tile_lims(stack, num_std, bar=True, size=20):
```
ttung: are these functions still used?

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: Ah, so apply accepts it (and I have an issue out to expand this across all functions, #231). 

I can try to clarify the documentation here. Basically it provides 3d support by determining whether we apply gaussian_low_pass over planes or volumes.

starfish/viz.py
ambrosejcarr: They might still be in some notebooks. I think they deserve some modernization, but probably shouldn't be removed yet.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: Not sure I understand the question, unless you mean it literally, in which case: It can, Pycharm will check it. I doubt mypy does though, because this would throw an error. 

 I need to figure out a better solution for 3d algorithm support, this duck typing is giving me headaches -- particularly in cases where the data is a 3d volume but the nuclei or dots image is not.

starfish/image/_stack.py
ambrosejcarr: Mypy says it should just be documented as Indices, so I will change it to that. However, yes, it will be treated as a tuple by python.

starfish/pipeline/features/spots/detector/local_max_peak_finder.py
ambrosejcarr: It interacts with the parameters, but no, I don't think so. I'll add a docstring that gives some details.

notebooks/allen_smFISH.py
ambrosejcarr: In the sense that I'm being stupid because I find it more intuitive to use the string when I'm writing, I guess. 

But you should keep reminding me -- much better to use Indices :)

notebooks/allen_smFISH.py
ambrosejcarr: You're right. I've removed it. Thanks.

notebooks/allen_smFISH.py
```diff
@@ -0,0 +1,187 @@
...
+# EPY: START markdown
+# Below, because spot finding is so slow when single-plex, we'll pilot this on a max projection to show that the parameters work. Here's what trackpy.locate, which we wrap, produces for a z-projection of channel 1. To do use our plotting methods on z-projections we have to expose some of the starfish internals, which will be improved upon. 
+# EPY: END markdown
+
+# EPY: START code
+from showit import image
+from trackpy import locate
+
+# grab a section from the tensor. 
+ch1 = s.max_proj((Indices.Z))[0, 1]
```
ttung: Why is the argument in double parens?

notebooks/allen_smFISH.py
```diff
@@ -0,0 +1,187 @@
...
+# plot the z-projection
+image(ch1, size=20, clim=(15, 52))
+
+# draw called spots on top as red circles
+# scale radius plots the red circle at scale_radius * spot radius
+s.image._show_spots(results, ax=plt.gca(), scale_radius=7)
+# EPY: END code
+
+# EPY: START markdown
+# Below spot finding is on the _volumes_ for each channel. This will take about `11m30s`
```
ttung: if this takes 11m30s, running this in travis will likely trigger timeouts unless the verbose=True triggers some output frequently enough.  may be worth trying out.

starfish/image/_stack.py
```diff
@@ -196,10 +223,17 @@ def show_stack(
...
         color_map : str (default = 'gray')
             string id of a matplotlib colormap
-        rescale : bool (default = True)
-            if True, rescale the data to exclude high and low-value outliers (see skimage.exposure.rescale_intensity)
         figure_size : Tuple[int, int] (default = (10, 10))
             size of the figure in inches
+        show_spots : Optional[SpotAttributes]
+            [Preliminary functionality] if provided, should be a SpotAttribute table that corresponds
+            to the volume being displayed. This will be paired automatically in the future.
+        rescale : bool (default = True)
```
ttung: Hm... why is the default of this True?

starfish/image/_stack.py
```diff
@@ -196,10 +223,17 @@ def show_stack(
...
+            if True, rescale the data to exclude high and low-value outliers (see skimage.exposure.rescale_intensity).
+        p_min: float
+            clip values below this intensity percentile. If provided, overrides rescale, above. (default = None)
```
ttung: If rescale defaulted to False, then this could throw a ValueError if both are set.

starfish/image/_stack.py
```diff
@@ -242,22 +276,72 @@ def show_stack(
...
+    @staticmethod
+    def _show_spots(result_df, ax, z=None, size=1, z_dist=1.5, scale_radius=5) -> None:
+        """function to plot spot finding results on top of any image as hollow red circles
+
+        called spots are colored by category
+
+        Parameters:
+        -----------
+        img : np.ndarray
+            2-d image
```
ttung: Can we make it a project convention to be specific about the acceptable dtypes for ndarrays?

starfish/image/_stack.py
```diff
@@ -335,29 +419,38 @@ def apply(self, func, is_volume=False, **kwargs):
...
+        if verbose:
+            tiles = tqdm(self._iter_tiles(indices))
+        else:
+            tiles = self._iter_tiles(indices)
 
+        applyfunc: Callable = partial(func, **kwargs)
         results = mapfunc(applyfunc, tiles)
 
+        if not in_place:
+            return list(results)
```
ttung: is this a useful result since you lose the indices of the tiles?

starfish/pipeline/filter/gaussian_low_pass.py
ttung: Like why isn't it:

`def __init__(self, sigma: Union[float, Tuple[float]], is_volume: bool=False, verbose=False, **kwargs) -> None:`

starfish/image/_stack.py
ambrosejcarr: YES. However we'll need to go learn about what they are by testing.

notebooks/allen_smFISH.py
ambrosejcarr: Got it, you mean the on-master runs, right? Sounds like we should knock the stack size down for testing.

starfish/image/_stack.py
ambrosejcarr: Will fix.

starfish/image/_stack.py
ambrosejcarr: Good idea. Will change.

starfish/image/_stack.py
ttung: Well, this is your viz code.  Seems like you should have an idea of what you'll accept :)

ambrosejcarr: Here I can add it -- it'll take any numerical array. But other filters are way more picky.

ttung: Really?  What if rescale=False?

ambrosejcarr: Yea, all these scaling factors are about dynamic ranges when you plot them as a colormap. 

i.e. if you have one value that's 2 billion and the rest are 0-100, you're going to get a black image. 

however, if you clip at the top at 99.5%, then you'll get a nice image where the range is 0-100.

ambrosejcarr: So if you screw things up here, you just get an ugly plot -- it'll still "work", though.

starfish/image/_stack.py
ambrosejcarr: Good catch, will fix.

PR #239, opened 2018-06-04
[TST] migrate to pytest

ambrosejcarr:
- migrate to pytest
- turn off F811 which complains about redefinition of parameters -- we import fixtures in pytest, which get used as parameters, which triggers this warning in every testing file
- update all tests except ISS to use pytest framework
- add merfish testing data and merfish tests that mirror the merfish notebook (just exercising code paths, no scientific validation yet)
- create test fixtures for datasets so they're downloaded once, but can be used multiple times
- remove some ill-fated type hinting in the gaussian filter modules that have bad ducktyping and play poorly with mypy. 

Potential problem: not clear if code coverage is being calculated in the same way as last time, but that trigger appears to only happen on PR so we'll see!


ambrosejcarr:
@ttung I've fixed the coverage issue, rebased relatively recently, and I think this is ready for review now. I'll rebase again before merge to reconcile the changes to `tile_metadata` based on how you decide to go in your PR.

starfish/test/full_pipelines/cli/test_profiler.py
```diff
@@ -0,0 +1,30 @@
...
+        "--noop",
+    ]
+    env = os.environ.copy()
+    if sys.version_info < (3, 0):
+        # this is a hideous hack because argparse is not consistent between 2.7 and 3.x about whether the subcommand
+        # is required.  TODO ttung: this entire hack can be dropped if we deprecate python2.7.
+        cmdline.append("noop")
+        env[PROFILER_NOOP_ENVVAR] = ""
+    if cmdline[0] == 'starfish':
+        coverage_cmdline = [
```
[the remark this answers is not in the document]
ambrosejcarr: I deleted a bunch of stuff and everything still seems to work, so looks like no. 👍

REQUIREMENTS-DEV.txt
```diff
@@ -1,5 +1,7 @@
 coverage
```
ttung: This can be removed, I would imagine (I suppose pytest-cov would bring it in anyway, but we no longer depend on it directly).

ambrosejcarr: You're correct. Sorry for missing this. 

There is no coverage data reported because this PR is currently based off a feature branch. When it's swapped to master the coverage should be reported. It _is_ uploaded. The results for this commit are: https://codecov.io/github/spacetx/starfish/commit/6bdd68df06017276dcbff48df65ba6c210519c6f

For F811 to be skipped only in testing files, I found [a package](https://github.com/snoack/flake8-per-file-ignores/blob/master/test/setup.cfg) for doing it but I haven't been able to get it to work. 

You can also call # noqa: F811 line-by-line but I thought that could confuse contributors. If you prefer this I can make a follow-up PR with the change.

ambrosejcarr:
Oh also, from looking at this PR, it seems like codecov is editing its report up top instead of posting new ones.

ambrosejcarr:
Finally, I'm not sure how to extricate the coverage hack. I'd encourage you to test it out and see after this is merged and see if it can get removed.

ambrosejcarr:
Some more learning about coverage: looks like when master is failing the coverage comparison doesn't run because codecov wants that report to form its comparison. I'm retesting after adding the skip file, and praying my changes didn't break the iss test. :)

PR #241, opened 2018-06-05
Convert to float64 before white top hat filter

ttung:
White top hat filter can generate negative values, and if it's an uint type, it will wrap around.  Raise the threshold for spot detection.

Depends on #242


ttung:
After more thought, I realized this is the wrong solution.  :(

PR #245, opened 2018-06-05
Clip bottom range of white-top-hat filter to 0

ttung:
Take the minimum of image,max_filtered to ensure we don't get a negative result.  This allows us to avoid the cast to float64.

This is a better solution to the problem #241 was attempting to solve.

Depends on #242

### DIFF
--- a/starfish/pipeline/filter/white_tophat.py
+++ b/starfish/pipeline/filter/white_tophat.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from ._base import FilterAlgorithmBase
 
 
@@ -44,10 +46,12 @@ class WhiteTophat(FilterAlgorithmBase):
         from skimage.morphology import disk
 
         def white_tophat(image):
+            if image.dtype.kind != "u":
+                raise TypeError("images should be stored in an unsigned integer array")
             structuring_element = disk(self.disk_size)
             min_filtered = minimum_filter(image, footprint=structuring_element)
             max_filtered = maximum_filter(min_filtered, footprint=structuring_element)
-            filtered_image = image - max_filtered
+            filtered_image = image - np.minimum(image, max_filtered)
             return filtered_image
 
         stack.image.apply(white_tophat)


starfish/pipeline/filter/white_tophat.py
```diff
@@ -47,7 +49,7 @@ def white_tophat(image):
             structuring_element = disk(self.disk_size)
             min_filtered = minimum_filter(image, footprint=structuring_element)
             max_filtered = maximum_filter(min_filtered, footprint=structuring_element)
-            filtered_image = image - max_filtered
+            filtered_image = image - np.minimum(image, max_filtered)
```
ambrosejcarr: I would have expected this to be np.maximum(image - max_filtered, 0) or np.clip(image - max_filtered, 0)

As programmed I feel like weird things could happen if `image` is smaller than `max_filtered`, but non-zero.

ttung: I don't think this works because image - max_filtered produces wrapped-around values.  none of the values are actually below 0.  we can compare the outputs to the inputs to detect underflow, but i think that's considerably more complicated.

ambrosejcarr: Oh I see. That's.. super scary. So, basically, you're seeing something like `7 - 10 = 65533` ? 

In that case, I'm pretty confident that the result should be zero, not 7.

ttung: well, it definitely shouldn't be 65533. :)

ambrosejcarr: Hehe. Yes, true. I'm not sure what the best solution is here. I could image casting one of these to a normal int might trigger a conversion. You can probably think up a better solution, but this would work: 
```
>>> x = np.uint(5)
>>> y = np.uint(7)
>>> np.maximum(x.astype(int) - y, 0).astype(y.dtype)
0
```

ttung: What's the concern here?  The only case where I think this could go pretty horribly wrong is if max_filtered has a negative value.  We could solve that by:

1. `filtered_image = image - np.clip(np.minimum(image, max_filtered), 0, None)`
2. asserting that max_filtered has an unsigned dtype.

If desired, we can assert that image has an unsigned dtype as well.

ambrosejcarr: I'm dumb. 🤦‍♂️  Approved.

ttung: I thought a bit about how to support floating point in its various ranges (-inf, +inf), (-1, +1), (0, +1), and I've concluded there's not a generic way to support any of them.  Doing so would require that we include the range in the data structures.  It's really only beneficial if we need a lot more precision in midtones (favoring (-inf, +inf), (-1, +1)) or shadows (0, +1).

I propose that we mandate unsigned integer types until we see a pressing need for more precision in those ranges.

PR #247, opened 2018-06-06
Fix ImageStack initialization

ttung:
1. Fix the initialization of default_tile_shape.  If the value is unset, we should just copy the unset value (None) rather than try to construct an empty tuple.
2. If we default_tile_shape is not set, grab the shape from the first tile that has the value set.
3. Bump to slicedimage 0.0.0-dev21 to pick up bugfixes https://github.com/spacetx/slicedimage/pull/14, https://github.com/spacetx/slicedimage/pull/15, and https://github.com/spacetx/slicedimage/pull/16


starfish/image/_stack.py
```diff
@@ -46,6 +46,10 @@ def __init__(self, image_partition):
                     raise TypeError("All tiles should have the same kind of dtype")
             if dtype.itemsize > max_size:
                 max_size = dtype.itemsize
+            if self._tile_shape is None:
+                self._tile_shape = tile.tile_shape
+            elif tile.tile_shape is not None and self._tile_shape != tile.tile_shape:
```
csweaver: Do you have to check if it is not none here since you already checked if it is none previously?

starfish/image/_stack.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, image_partition):
             self._num_zlayers = image_partition.get_dimension_shape(Indices.Z)
         else:
             self._num_zlayers = 1
-        self._tile_shape = tuple(image_partition.default_tile_shape)
+        self._tile_shape = image_partition.default_tile_shape
```
csweaver: What type is an image_partition?

starfish/image/_stack.py
ttung: Yup, we don't need the check.  Good call.

starfish/image/_stack.py
ttung: It's a TileSet.

starfish/image/_stack.py
ttung: Actually, I'm wrong.  First check is `self._tile_shape`, and the second check is `tile.tile_shape`

csweaver: oops missed that they were different, my fault

starfish/image/_stack.py
csweaver: Can you point me to the tileset object definition? (Just for my own learning, I don't have any concerns I want to address with the code, feel free to merge)

ambrosejcarr: @csweaver It lives here: https://github.com/spacetx/slicedimage/blob/master/slicedimage/_tileset.py

PR #248, opened 2018-06-06
Fix ISS notebook.

ttung:
Depends on #246 and #247


notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
```diff
@@ -13,18 +13,19 @@
...
 from showit import image, tile
 import pprint
 # EPY: ESCAPE %matplotlib inline
 
 from starfish.io import Stack
+from starfish.constants import Indices
 
 s = Stack()
-s.read('https://dmf0bdeheu4zf.cloudfront.net/ISS/fov_001/experiment.json')
+s.read('https://dmf0bdeheu4zf.cloudfront.net/20180606/ISS/fov_001/experiment.json')
```
ambrosejcarr: This versioning system works for me. I've swapped mine over.

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
```diff
@@ -13,18 +13,19 @@
...
 # s.squeeze() simply converts the 4D tensor H*C*X*Y into a list of len(H*C) image planes for rendering by 'tile'
-tile(s.squeeze());  
+tile(s.squeeze());
```
ambrosejcarr: How do extra spaces get into the .py files? Did I do this? :|

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
```diff
@@ -76,49 +77,41 @@
...
-# ## Filter and scale raw data 
+# ## Filter and scale raw data
 # 
-# Now apply the white top hat filter to both the spots image and the individual channels. White top had enhances white spots on a black background. 
+# Now apply the white top hat filter to both the spots image and the individual channels. White top had enhances white spots on a black background.
 # EPY: END markdown
 
 # EPY: START code
-from starfish.filters import white_top_hat 
+from starfish.pipeline.filter import Filter
```
ambrosejcarr: This import differs from the way we import the gaussian spot detector. Should we be consistent?

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
ttung: Automated tooling.

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
ttung: Yes.  What's your preference?

ambrosejcarr: I think I prefer the way you've done it here, because it reinforces that all methods need to be imported in `__init__.py` (You'll notice that many currently are _not_ -- I didn't realize we needed to do that when I was changing some of them over, and so the CLI will not include them). 

.. but the API still works if you just incorporate the methods, so I think this is a more robust way to enforce API == CLI.

ttung: I fixed the gaussian spot instance.

PR #250, opened 2018-06-06
Fix extraction of tile metadata

ttung:
There are two issues with how tile metadata is extracted.

1. Any two tiles that contain the same extras field will erroneously trigger the error where a key is found in tile indices and extras.  This is because we create a new key in 'k' the first time we see the extra, and the second instance will find it in data.
2. If the tiles don't have the same set of extras, the columns in data won't be the same length.

This approach gathers all the keys for indices and all the keys for extras.  If the keys overlap at all, then flag the warning.

Then as we go through each tile, we try to get the value.  If it's not present, we add `None` to the column.

Depends on #246


ttung:
> I think your solution is fine as well, although I personally find mine a bit more readable. You can pick. :)

As I mentioned in #239, this doesn't cover the case where the extras or the indices are not identical in which keys they contain.  I don't know if such a check belongs here or elsewhere.

ambrosejcarr:
I think your reasoning is good. This is a fine place to program the check for now, I think.

ttung:
I was going to add unit tests, which was why this wasn't landed yet. :)

Anyway, #264 addresses that (and a bug!)

PR #252, opened 2018-06-07
Validate that the notebooks can run successfully

ttung:
If *.py.skip exists, then that notebook is skipped.

See https://travis-ci.org/spacetx/starfish/builds/389071632 for a run when .travis.yml is set to look for a different branch name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,14 @@ python:
 - 3.6
 dist: trusty
 install:
-- pip install -r REQUIREMENTS-DEV.txt
+- pip install -r REQUIREMENTS-DEV.txt -r REQUIREMENTS-NOTEBOOK.txt
 - pip install -e .
 - pip freeze
 script:
+- export MPLBACKEND=Agg
+- export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
 - set -eo pipefail
 - make all
+- if [ "$BRANCH" == "master" ]; then make run_notebooks; fi
 after_success:
 - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


ttung:
> Is there any way you could add a log of some kind that points to which notebook is broken? that could save us a lot of time when we bust stuff.

Not sure what you mean.  The travis logs _should_ say.

ambrosejcarr:
Ok, great. Does it fail immediately, or run all the notebooks and then fail? This is something that tripped us up in green -- we've found it's best to test everything and report all the failures.

joshmoore:
Follow-up: with gh-596 we could also include the notebooks in an "allow fail' section of .travis.yml. That being said, is there a timeline for reviewing and removing:

```
$ ls notebooks/py/*.skip
notebooks/py/allen_smFISH.py.skip  notebooks/py/Starfish_simulation.py.skip
```
?

ttung:
@shanaxel42 with travis caching, can `allen_smFISH` run successfully on travis?  or is it silent too long?

shanaxel42:
Looks like we hit a travis memory error https://travis-ci.org/spacetx/starfish/jobs/432579202

joshmoore:
@shanaxel42 : you should be able to get the same functionality as in https://github.com/spacetx/starfish/commit/f0c92ab4a1f3 by pushing a branch prefixed with `master-` or `merge-`. If `testing-` is a better name, happy to add it.

ambrosejcarr:
>  on a 7.5GB instance...

Notebooks running simultaneously on the same instance, or on separate instances?

ttung:
Separate instances.

ttung:
There's something definitely fishy going on.

ambrosejcarr:
```
x      y      z    ch  r   dtype 
2048 * 2048 * 32 * 4 * 1 * 4 = 2147483648 bytes
= 21.5 gb
```

The data used to be uint16, and at half the above payload, would have fit in a laptop's memory.

PR #261, opened 2018-06-08
[ENH] Enable 3-d support for gaussian high-pass

ambrosejcarr:
- Makes `gaussian_high_pass` and `gaussian_low_pass` consistent
- Fixes typing through use of `Number`
- Adds 3D support for gaussian_high_pass (`osmFISH` needs it)
- Adds a test for new functionality

### DIFF
--- a/starfish/pipeline/filter/gaussian_high_pass.py
+++ b/starfish/pipeline/filter/gaussian_high_pass.py
@@ -1,9 +1,12 @@
 import argparse
 from functools import partial
-from typing import Callable, Optional
+from numbers import Number
+from typing import Callable, Union, Tuple, Optional
 
 import numpy as np
+from skimage import img_as_uint
 
+from starfish.errors import DataFormatWarning
 from starfish.image import ImageStack
 from starfish.pipeline.filter.gaussian_low_pass import GaussianLowPass
 from ._base import FilterAlgorithmBase
@@ -11,47 +14,70 @@ from ._base import FilterAlgorithmBase
 
 class GaussianHighPass(FilterAlgorithmBase):
 
-    def __init__(self, sigma, **kwargs) -> None:
+    def __init__(
+            self, sigma: Union[Number, Tuple[Number]], is_volume: bool=False, verbose: bool=False, **kwargs
+    ) -> None:
         """Gaussian high pass filter
 
         Parameters
         ----------
-        sigma : int (default = 1)
+        sigma : Union[Number, Tuple[Number]]
             standard deviation of gaussian kernel
+        is_volume : bool
+            If True, 3d (z, y, x) volumes will be filtered, otherwise, filter 2d tiles independently.
+        verbose : bool
+            if True, report on filtering progress (default = False)
 
         """
+        if isinstance(sigma, tuple):
+            message = ("if passing an anisotropic kernel, the dimensionality must match the data shape ({shape}), not "
+                       "{passed_shape}")
+            if is_volume and len(sigma) != 3:
+                raise ValueError(message.format(shape=3, passed_shape=len(sigma)))
+            if not is_volume and len(sigma) != 2:
+                raise ValueError(message.format(shape=2, passed_shape=len(sigma)))
+
         self.sigma = sigma
+        self.is_volume = is_volume
+        self.verbose = verbose
 
     @classmethod
     def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
         group_parser.add_argument(
-            "--sigma", default=1, type=int, help="standard deviation of gaussian kernel")
+            "--sigma", type=float, help="standard deviation of gaussian kernel")
+        group_parser.add_argument(
+            "--is-volume", action="store_true", help="indicates that the image stack should be filtered in 3d")
 
     @staticmethod
-    def gaussian_high_pass(img: np.ndarray, sigma) -> np.ndarray:
+    def high_pass(image: np.ndarray, sigma: Union[Number, Tuple[Number]]) -> np.ndarray:
         """
         Applies a gaussian high pass filter to an image
 
         Parameters
         ----------
-        img : numpy.ndarray
-            Image to filter
-        sigma : Union[float, int]
+        image : numpy.ndarray[np.uint32]
+            2-d or 3-d image data
+        sigma : Union[Number, Tuple[Number]]
             Standard deviation of gaussian kernel
 
         Returns
         -------
-        numpy.ndarray :
-            Filtered image, same shape as input
+        np.ndarray :
+            Standard deviation of the Gaussian kernel that will be applied. If a float, an isotropic kernel will be
+            assumed, otherwise the dimensions of the kernel give (z, y, x)
 
         """
-        blurred: np.ndarray = GaussianLowPass.low_pass(img, sigma)
+        if image.dtype != np.uint16:
+            DataFormatWarning('gaussian filters currently only support uint16 images. Image data will be converted.')
+            image = img_as_uint(image)
 
-        over_flow_ind: np.ndarray[bool] = img < blurred
-        res: np.ndarray = img - blurred
-        res[over_flow_ind] = 0
+        blurred: np.ndarray = GaussianLowPass.low_pass(image, sigma)
 
-        return res
+        over_flow_ind: np.ndarray[bool] = image < blurred
+        filtered: np.ndarray = image - blurred
+        filtered[over_flow_ind] = 0
+
+        return filtered
 
     def filter(self, stack: ImageStack, in_place: bool=True) -> Optional[ImageStack]:
         """Perform filtering of an image stack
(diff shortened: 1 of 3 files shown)


starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -29,41 +32,44 @@ def __init__(self, sigma, is_volume: bool=False, verbose=False, **kwargs) -> Non
...
-    def get_algorithm_name(cls):
+    def get_algorithm_name(cls) -> str:
         return "gaussian_low_pass"
 
     @classmethod
-    def add_arguments(cls, group_parser):
+    def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
         group_parser.add_argument(
-            "--sigma", default=1, type=int, help="standard deviation of gaussian kernel")
+            "--sigma", default=1, type=float, help="standard deviation of gaussian kernel")
```
ttung: how does this work with volumes?

ttung: Also, I don't see any CLI parameterization for is_volume

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -29,41 +32,44 @@ def __init__(self, sigma, is_volume: bool=False, verbose=False, **kwargs) -> Non
...
+            2-d or 3-d image data
+        sigma : Union[Number, Tuple[Number]]
             Standard deviation of the Gaussian kernel that will be applied. If a float, an isotropic kernel will be
             assumed, otherwise the dimensions of the kernel give (z, x, y)
 
         Returns
         -------
         np.ndarray :
-            Blurred data in same shape as input image.
+            Blurred data in same shape as input image, converted to np.uint16 dtype.
```
ttung: Should we constrain the input types? At the very least, I would imagine that the skimage limitations should apply here.

starfish/pipeline/filter/gaussian_high_pass.py
```diff
@@ -29,33 +38,33 @@ def get_algorithm_name(cls) -> str:
...
+            Filtered image, same shape as input, converted to np.uint16 dtype.
 
         """
-        blurred: np.ndarray = GaussianLowPass.low_pass(img, sigma)
+        blurred: np.ndarray = GaussianLowPass.low_pass(image, sigma)
 
-        over_flow_ind: np.ndarray[bool] = img < blurred
-        res: np.ndarray = img - blurred
-        res[over_flow_ind] = 0
+        over_flow_ind: np.ndarray[bool] = image < blurred
```
ttung: I don't think this approach works for (-1, +1) ranged data.

starfish/pipeline/filter/gaussian_high_pass.py
```diff
@@ -68,9 +77,9 @@ def filter(self, stack: Stack) -> None:
 
         """
 
-        high_pass: Callable = partial(self.gaussian_high_pass, sigma=self.sigma)
-        stack.image.apply(high_pass)
+        high_pass: Callable = partial(self.high_pass, sigma=self.sigma)
```
ttung: Idea: verify that if self.sigma is a tuple, that the length is equal to the shape of the image.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: if a scalar is passed, it is broadcast across all axes of the volume, so valid inputs are a scalar or a tuple of length == len(array.shape)

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: There's some craziness here. I'm just documenting it in this PR but I've added a todo to look into why this is here and hopefully remove it.

starfish/pipeline/filter/gaussian_high_pass.py
ambrosejcarr: You're right, it won't, however the gets cast to uint16 in `low_pass`, so that's not a problem here (although the _casting_ is potentially a big problem).

ambrosejcarr:
I've looked into the casting, which was definitely wrong. For the short term, a good solution is to warn the users that we only support uint16 image data here and properly cast to it using the `skimage` functionality. 

I've added some clipping to low_pass to prevent some crazy wrap-around behavior that was previously present, and verified that if passing a tuple, that it matches the expected image dimensions. 

Finally, I added CLI args.

starfish/pipeline/filter/gaussian_high_pass.py
ambrosejcarr: ... and now that casting is explicit and won't break the results.

starfish/pipeline/filter/gaussian_high_pass.py
ambrosejcarr: Done.

starfish/pipeline/filter/gaussian_low_pass.py
ambrosejcarr: We can safely cast the image types, added a warning when that happens.

starfish/pipeline/filter/gaussian_high_pass.py
```diff
@@ -1,26 +1,45 @@
...
+    ) -> None:
         """Gaussian high pass filter
 
         Parameters
         ----------
-        sigma : int (default = 1)
+        sigma : Union[Number, Tuple[Number]]
             standard deviation of gaussian kernel
+        is_volume : bool
+            If True, 3d (z, y, x) volumes will be filtered, otherwise, filter 2d tiles independently.
```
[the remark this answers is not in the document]
ambrosejcarr: This speaks to a general need to figure out what it means for 3d volumes to be first class. I suspect some approaches might do some weirds when presented with a (1, y, x).

I like `filter_by_2d_tile` as a flag with 3d as default but that could potentially be a lot of work to implement. I think the right thing here is to make an issue. What do you think?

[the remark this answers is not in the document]
ttung: Cool that sounds good.

starfish/pipeline/filter/gaussian_high_pass.py
```diff
@@ -29,33 +48,40 @@ def get_algorithm_name(cls) -> str:
...
-            Filtered image, same shape as input
+        np.ndarray :
+            Standard deviation of the Gaussian kernel that will be applied. If a float, an isotropic kernel will be
+            assumed, otherwise the dimensions of the kernel give (z, y, x)
 
         """
-        blurred: np.ndarray = GaussianLowPass.low_pass(img, sigma)
+        if image.dtype != np.uint16:
+            DataFormatWarning('gaussian filters currently only support uint16 images. Image data will be converted.')
+            image = img_as_uint(image)
```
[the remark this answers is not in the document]
ambrosejcarr: Yea, `skimage` takes care of it so long as it's in a standard format (those two will be autodetected).

starfish/pipeline/filter/gaussian_low_pass.py
```diff
@@ -1,69 +1,92 @@
...
         """
+        if image.dtype != np.uint16:
+            DataFormatWarning('gaussian filters only support uint16 images. Image data will be converted')
+            image = img_as_uint(image)
+
         blurred = gaussian(
             image, sigma=sigma, output=None, cval=0, multichannel=True, preserve_range=True, truncate=4.0)
 
-        blurred = blurred.astype(numpy.uint16)
+        blurred = blurred.clip(0).astype(np.uint16)
```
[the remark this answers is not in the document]
ambrosejcarr: The filter converts the data to float, and can produce values below zero because of how the gaussian convolution works. Those are not physiological and therefore not desirable, so we clip them.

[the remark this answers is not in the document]
ambrosejcarr: I should also say: simple conversion to numpy.uint16 is a bug: the data wraps around. Very bad.

starfish/pipeline/filter/gaussian_high_pass.py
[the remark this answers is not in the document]
ttung: I think the only downside of this approach is that the params/cmd line args would change, but I think we can live with that.

starfish/pipeline/filter/gaussian_high_pass.py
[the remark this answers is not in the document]
ttung: I don't see how they can be autodetected.  There is no way to differentiate between an image that is entirely between midpoint grey and white, but in range (-1, +1), and an image that's (0, 1).

starfish/pipeline/filter/gaussian_low_pass.py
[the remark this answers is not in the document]
ttung: Ohhh.

ambrosejcarr:
> I'm going to approve this because it's been pending forever, but I think we need to have a discussion about data types.

100% agree. I'm thinking a decorator that uses skimage to try to convert whatever users give us to uint16 might be the way to go. 

I'll rebase + merge this next time I have good internet.

starfish/pipeline/filter/gaussian_high_pass.py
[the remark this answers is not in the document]
ambrosejcarr: That's an interesting point, which provokes a question -- our image analysis is mostly about spot detection (bright against dark background). Do we think it matters that much that we can't differentiate these different types?

PR #262, opened 2018-06-08
migrate test_slicedimage_dtype.py to pytest framework

ttung:


starfish/test/image/test_slicedimage_dtype.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+NUM_HYB = 2
+NUM_CH = 2
+NUM_Z = 2
+HEIGHT = 10
+WIDTH = 10
+
+
+def make_stack(dtype: numpy.number, corner_dtype: numpy.number):
```
ambrosejcarr: Could this be combined with the [synthetic_stack](https://github.com/spacetx/starfish/blob/master/starfish/test/dataset_fixtures.py#L31) fixture? 

If not, it's probably still a good idea to make it a fixture and move it to that file so that other tests could load this and test against it. 

N.B.: If confusing, can see how the fixtures work by checking `test_apply.py`, which consumes the `synthetic_stack` tile function you'd written previously and that I converted to a fixture.

starfish/test/image/test_slicedimage_dtype.py
```diff
@@ -0,0 +1,116 @@
...
+        stack = make_stack(numpy.int32, numpy.int8)
+        assert len(w) == 1
+        assert issubclass(w[0].category, DataFormatWarning)
+    expected = numpy.empty(
+        (NUM_HYB,
+         NUM_CH,
+         NUM_Z,
+         HEIGHT,
+         WIDTH), dtype=numpy.int32)
+    expected.fill(16777216)
```
ambrosejcarr: Love this. Great tests. 👍

starfish/test/image/test_slicedimage_dtype.py
```diff
@@ -0,0 +1,116 @@
...
+         WIDTH), dtype=numpy.int32)
+    expected[0, 0, 0] = corner
+    assert stack.numpy_array.all() == expected.all()
+
+
+def test_float_type_promotion():
+    with warnings.catch_warnings(record=True) as w:
+        stack = make_stack(numpy.float64, numpy.float32)
+        assert len(w) == 1
+        assert issubclass(w[0].category, DataFormatWarning)
```
ambrosejcarr: 👍

starfish/test/image/test_slicedimage_dtype.py
ttung: #268

PR #264, opened 2018-06-08
Add some basic tile metadata tests.

ttung:
Also fixed a bug where we were joining the keys blindly, but they might not necessarily be strings.

Depends on #263, #271

### DIFF
--- a/starfish/test/image/test_imagestack_metadata.py
+++ b/starfish/test/image/test_imagestack_metadata.py
@@ -1,0 +1,67 @@
+from typing import Any
+
+import pytest
+
+from starfish.constants import Indices
+from starfish.test.dataset_fixtures import DEFAULT_NUM_HYB, DEFAULT_NUM_CH, DEFAULT_NUM_Z, synthetic_stack_factory
+
+
+def test_metadata(synthetic_stack_factory):
+    """
+    Normal situation where all the tiles have uniform keys for both indices and extras.
+    """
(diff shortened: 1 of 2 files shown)


starfish/test/image/test_imagestack_metadata.py
```diff
@@ -0,0 +1,154 @@
...
+from starfish.image import ImageStack
+
+NUM_HYB = 2
+NUM_CH = 2
+NUM_Z = 2
+HEIGHT = 10
+WIDTH = 10
+
+
+def make_stack(indices: Sequence[dict], extras: Sequence[dict]):
```
ambrosejcarr: I suspect this could be the only stack-creation object we need (it could replace the other ones scattered around in testing). After we make an empty stack we can just fill part of it with a certain value or cast it to a different dtype for _most_ of the tests. 

Generally, I think we need some basic stack creation routines in starfish that we can use for this (filling the stack with ones) but also for creating the metadata from files (filling the stack with data). I think they could then be used across the repo for all our stack creation. 

What're your thoughts?

ttung: Really?  This doesn't allow for the parameterization of data.  I think the `synthetic_stack` merged with the one in `examples/support` is closest to the Ultimate Generator of Data™.

ttung: In any case, I've refactored this one to use the `synthetic_stack` one. :)

PR #266, opened 2018-06-09
[easy] Turn synthetic_stack into a factory.

ttung:
The model is https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures

### DIFF
--- a/starfish/test/dataset_fixtures.py
+++ b/starfish/test/dataset_fixtures.py
@@ -1,3 +1,5 @@
+from typing import Any, Callable
+
 import numpy as np
 import pytest
 from slicedimage import Tile, TileSet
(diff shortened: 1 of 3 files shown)


starfish/test/dataset_fixtures.py
```diff
@@ -28,49 +30,51 @@ def merfish_stack() -> Stack:
 
 
 @pytest.fixture(scope='session')
-def synthetic_stack() -> ImageStack:
-    """generate a synthetic ImageStack
+def synthetic_stack_factory():
+    def synthetic_stack()-> ImageStack:
```
ambrosejcarr: Love this whole PR chain. I'm still wondering if this synthetic_stack object is more general than we're posing it here -- I have to make stacks for lots of reasons, not just testing: 

1. testing
2. creation from org json
3. creation of synthetic data
4. maybe out-of-place operations
5. probably lots of others 

I'm wondering if this deserves an issue that requests we take all the code that takes tilesets and move it to a `format`, `data` , or `util` package.

I'm thinking of the thought experiment: say we moved away from `ImageStack` in the future, could we have the codebase in such a state that there's a single package we need to refactor to accomplish that? 

In any case, this is awesome work and I'll change my fixtures to match your pattern, thanks for finding that.

starfish/test/dataset_fixtures.py
```diff
@@ -28,49 +30,51 @@ def merfish_stack() -> Stack:
 
 
 @pytest.fixture(scope='session')
-def synthetic_stack() -> ImageStack:
-    """generate a synthetic ImageStack
+def synthetic_stack_factory():
```
ambrosejcarr: could use a one-liner docstring in case someone's not familiar with factories.

starfish/test/dataset_fixtures.py
ttung: I think we need to slurp in the code from examples/support/__init__.py, plus have a generalizable framework for handling coordinates.

ttung:
@ambrosejcarr I am curious why this needs to be a fixture.  I find the entire pytest fixture system a bit clumsy.

starfish/test/dataset_fixtures.py
ambrosejcarr: Makes sense to me.

PR #267, opened 2018-06-09
Allow tests to provide tile data or extras

ttung:


starfish/test/dataset_fixtures.py
```diff
@@ -29,9 +29,20 @@ def merfish_stack() -> Stack:
     return s
 
 
+def default_tile_data_provider(hyb: int, ch: int, z: int, height: int, width: int) -> np.ndarray:
```
ambrosejcarr: doc each of these new functions please, I imagine devs (me, heh) will want to use these. :)

PR #269, opened 2018-06-10
Nuke ImageBase and move all relevant docblocks to ImageStack

ttung:
Since there isn't going to be SingleImage any more, let's nuke ImageBase.

Test plan:
```
[czipa1osx186 (.venv)]:~/microscopy/starfish:tonytung-imagebase> hg grep ImageBase
[czipa1osx186 (.venv)]:~/microscopy/starfish:tonytung-imagebase>
```


starfish/image/_stack.py
```diff
@@ -115,18 +114,31 @@ def __init__(self, image_partition):
 
     @classmethod
     def from_url(cls, relativeurl, baseurl):
```
ambrosejcarr: I still get confused about how relativeurl + baseurl should be used here. Could you add an example (or more than one, if there are multiple common use cases)? I think it would help me not keep rewriting code that relates to ImageStack.

ttung: Updated.  Please let me know if it addresses your concern.

PR #270, opened 2018-06-10
Write relative paths correctly when writing experiment.json

ttung:
If stack.write(..) gets an absolute path, then experiment.json['hybridization'] is an absolute path and everything works.  However, if stack.write(..) gets a relative path, then experiment.json['hybridization'] gets the relative path twice -- once from the baseurl, and once from the written entry.

This resolves the bug by always treating everything as relative paths.


ambrosejcarr:
Could we write a test that would catch this if we screwed it up again in the future?

ttung:
> Could we write a test that would catch this if we screwed it up again in the future?

Uh, let's see how long Stack lasts. :)

PR #271, opened 2018-06-10
Allow parameterization of generated stacks (HYB/CH/Z)

ttung:


ttung:
> Here, the parameterization must be done manually by the tester, right?

Only if they need to.  There are sensible defaults.

ambrosejcarr:
Ok, that's what I meant. Thanks. 👍

PR #272, opened 2018-06-10
Fix the dtype test.

ttung:
1. Actually test the data rather than calling all().  I completely misunderstood what that function does.
2. Add a uint test (separate from int)
3. Fix how fixed-point scaling works.  It should multiply by the difference in the ranges, not the max.
4. Change how floating point scaling works (it does not interpolate the range any more).

Depends on #271

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -88,21 +88,11 @@ class ImageStack:
             zlayer = tile.indices.get(Indices.Z, 0)
             data = tile.numpy_array
             if max_size != data.dtype.itemsize:
-                # this source tile has a smaller data size than the other ones, though the same kind.  need to scale the
-                # data.
-                if data.dtype.kind == "f":
-                    # floating point can be done with numpy.interp.
-                    src_finfo = numpy.finfo(data.dtype)
-                    dst_finfo = numpy.finfo(self._data.dtype)
-                    data = numpy.interp(
-                        data,
-                        (src_finfo.min, src_finfo.max),
-                        (dst_finfo.min, dst_finfo.max))
-                else:
+                if data.dtype.kind == "i" or data.dtype.kind == "u":
                     # fixed point can be done with a simple multiply.
-                    src_max = numpy.iinfo(data.dtype).max
-                    dst_max = numpy.iinfo(self._data.dtype).max
-                    data = data * (dst_max / src_max)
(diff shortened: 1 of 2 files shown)


starfish/image/_stack.py
```diff
@@ -88,21 +88,11 @@ def __init__(self, image_partition):
...
-                        (dst_finfo.min, dst_finfo.max))
-                else:
+                if data.dtype.kind == "i" or data.dtype.kind == "u":
                     # fixed point can be done with a simple multiply.
-                    src_max = numpy.iinfo(data.dtype).max
-                    dst_max = numpy.iinfo(self._data.dtype).max
-                    data = data * (dst_max / src_max)
+                    src_range = numpy.iinfo(data.dtype).max - numpy.iinfo(data.dtype).min + 1
+                    dst_range = numpy.iinfo(self._data.dtype).max - numpy.iinfo(self._data.dtype).min + 1
+                    data = data * (dst_range / src_range)
```
ambrosejcarr: Is there any chance of this multiplication doing an implicit conversion to float (e.g.nif the `dst_range / src_range` is a float instead of an int)

starfish/test/image/test_slicedimage_dtype.py
```diff
@@ -63,38 +72,56 @@ def test_int_type_promotion(synthetic_stack_factory):
...
          WIDTH), dtype=numpy.int32)
-    expected.fill(16777216)
-    corner = numpy.ones(
+    corner = numpy.empty(
         (HEIGHT,
          WIDTH), dtype=numpy.int32)
+    corner.fill(16777216)
     expected[0, 0, 0] = corner
-    assert stack.numpy_array.all() == expected.all()
+    assert numpy.array_equal(stack.numpy_array, expected)
```
ambrosejcarr: if you wanted to use `.all()` here I think it would work to use it as follow, although I think `array_equal` is more expressive: 

`(stack.numpy_array == expected).all()`
or
`np.all(stack.numpy_array == expected)`

starfish/image/_stack.py
ttung: hmmmm, I don't think that's possible.  python integer gets upgraded to some > 64bit integer automagically.

>>> type(2 ** 64 + 2)
<class 'int'>

PR #273, opened 2018-06-10
Remove the indirection for the synthetic stack factory

ttung:
1. Made it a simple function.
2. Remove the now unnecessary deepcopy calls.

### DIFF
--- a/starfish/test/dataset_fixtures.py
+++ b/starfish/test/dataset_fixtures.py
@@ -50,58 +50,51 @@ DEFAULT_HEIGHT = 30
 DEFAULT_WIDTH = 20
 
 
-@pytest.fixture(scope='session')
-def synthetic_stack_factory():
+def synthetic_stack(
+        num_hyb: int=DEFAULT_NUM_HYB,
+        num_ch: int=DEFAULT_NUM_CH,
+        num_z: int=DEFAULT_NUM_Z,
+        tile_height: int=DEFAULT_HEIGHT,
+        tile_width: int=DEFAULT_WIDTH,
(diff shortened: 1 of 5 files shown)


starfish/test/dataset_fixtures.py
```diff
@@ -50,58 +50,51 @@ def default_tile_extras_provider(hyb: int, ch: int, z: int) -> Any:
 DEFAULT_WIDTH = 20
 
 
-@pytest.fixture(scope='session')
-def synthetic_stack_factory():
-    """
-    Inject this factory, which is a method to produce image stacks.
+def synthetic_stack(
```
ambrosejcarr: Now that it's a function, can we make these global-scope-inherited `Indices` into parameters with default values so that this can be called elsewhere?

Also please take a look at my synthetic data PR sometime soon -- using that as a `tile_data_provider` is probably the way we want to go for more robust testing of downstream components that need things like spots to be evaluated.

ttung: > Now that it's a function, how would you feel about making these global-scope-inherited Indices into parameters with default values so that this can be called elsewhere?

Not sure I follow here.  Can you explain?

ambrosejcarr: Sorry, I was suggesting that synthetic stack could take as parameters `NUM_HYB, NUM_CH, ...`. Do you think that would be useful?

ttung: it does...?

ttung: https://github.com/spacetx/starfish/blob/f0d545a631cb95640729faa9e2765f24ee1921e4/starfish/test/dataset_fixtures.py#L59

PR #274, opened 2018-06-11
Add helper function to more easily construct ImageStacks from a URL or a path

ttung:


ttung:
> Cool, I like it. Checking -- do you see this as a user-facing function, or do you envision us nudging people towards your nascent Stack replacement?

I would say this is more likely an user-facing function.

PR #275, opened 2018-06-11
convert codebook retrieval to use jsonpathrw

ttung:
I did this a long time ago in #96, but I forgot the reason I did it by the time the code was being reviewed.  Now I remember: this is to generalize the retrieval of data fields when converting away from `Stack`.

### DIFF
--- a/REQUIREMENTS-DEV.txt
+++ b/REQUIREMENTS-DEV.txt
@@ -1,4 +1,5 @@
 flake8
+jsonpath_rw
 pytest-cov>=2.5.1
 pytest-xdist
 mypy
--- a/starfish/test/full_pipelines/cli/test_iss.py
+++ b/starfish/test/full_pipelines/cli/test_iss.py
@@ -6,15 +6,27 @@ import subprocess
 import sys
 import tempfile
 import unittest
+from typing import Sequence
+
+import jsonpath_rw
 
 from starfish.util import clock
 
 
-def get_codebook(tempdir):
-    json_filepath = os.path.join(tempdir, "formatted", "experiment.json")
+def get_jsonpath_from_file(json_filepath_components: Sequence[str], jsonpath: str):
+    """
+    Given a series of filepath components, join them to find a file <FILE>.  Open that file, and locate a specific value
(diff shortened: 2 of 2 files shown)


starfish/test/full_pipelines/api/test_iss.py
```diff
@@ -6,15 +6,27 @@
...
-def get_codebook(tempdir):
-    json_filepath = os.path.join(tempdir, "formatted", "experiment.json")
+def get_jsonpath_from_file(json_filepath_components: Sequence[str], jsonpath: str):
+    """
+    Given a series of filepath components, join them to find a file <FILE>.  Open that file, and locate a specific value
+    in the json structure <PATH>.  Join the directory path of <FILE> and <PATH> and return that.
+
+    For example, if json_path_components is ["/tmp", "formatted", "experiment.json"] and jsonpath is
+    "$['hybridization']", this method will open /tmp/formatted/experiment.json, decode that as a json document, and
+    locate the value of the key 'hybridization'.  It will return /tmp/formatted/XXX, where XXX is the value of the key.
```
ambrosejcarr: This seems more complex than simply making a user pass the codebook location separately as a string. Could you explain more why you're doing it this way?

starfish/test/full_pipelines/api/test_iss.py
```diff
@@ -6,15 +6,27 @@
...
 
 
-def get_codebook(tempdir):
-    json_filepath = os.path.join(tempdir, "formatted", "experiment.json")
+def get_jsonpath_from_file(json_filepath_components: Sequence[str], jsonpath: str):
+    """
+    Given a series of filepath components, join them to find a file <FILE>.  Open that file, and locate a specific value
+    in the json structure <PATH>.  Join the directory path of <FILE> and <PATH> and return that.
+
+    For example, if json_path_components is ["/tmp", "formatted", "experiment.json"] and jsonpath is
```
ambrosejcarr: `json_path_components` -> `json_filepath_components`

starfish/test/full_pipelines/api/test_iss.py
ambrosejcarr: Nevermind.. looking at the next PR now.

PR #284, opened 2018-06-14
BUG: [easy] update notebook encoding

ambrosejcarr:
The synthetic notebook got pushed to master without updating the encoding. Here's an update that un-breaks the tests.


ttung:
How does this not fail the .ipynb <-> .py test?

PR #286, opened 2018-06-14
DEP: AlgorithmBase.get_algorithm_name

ambrosejcarr:
- `AlgorithmBase.get_algorithm_name` is no longer abstract
- Remove overriding methods. 
- All Algorithms are now referenced by class name
- Update tests and notebooks.


ttung:
I didn't read this, but I assume this is a mechanical refactor that will be a huge pain to carry around.

ambrosejcarr:
You are correct. Manual refactor. Seems like it doesn't break anything, and I went through the other notebooks and there don't seem to be any issues.

PR #293, opened 2018-06-18
removed the int type requirement in WhiteTophat filter

kevinyamauchi:
I removed the uint check for the input images for the WhiteTophat filter. It doesn't seem necessary for the Scipy implementation of the min/max filters and none of the other filters have a type requirement. Please let me know if I missed a type spec for images upstream. Thanks!

### DIFF
--- a/starfish/pipeline/filter/white_tophat.py
+++ b/starfish/pipeline/filter/white_tophat.py
@@ -42,8 +42,6 @@ class WhiteTophat(FilterAlgorithmBase):
         from skimage.morphology import disk
 
         def white_tophat(image):
-            if image.dtype.kind != "u":
-                raise TypeError("images should be stored in an unsigned integer array")
             structuring_element = disk(self.disk_size)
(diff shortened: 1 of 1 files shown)


ttung:
So the rationale for requiring uint types is that there is not a general algorithm that describes white_top_hat.  We can probably infer what range is in play for a float array, but even that's at best a guess.  For instance, if the array only contains values from (0,1), that could either mean that intensity ranges from 0%-100%, or the array is actually a (-1,1) array, and the data just happens to be intensity range 50%-100%..

ambrosejcarr:
@kevinyamauchi @ttung How would you guys feel about making a function decorator that:

1. checks that the data is in a valid skimage format
2. attempts to cast it to uint
3. when the function completes, revert the data to the original type

I think this would give us some flexibility to support skimage data types, but would let us internally code against uint16 data, which might make life easier for us.

dganguli:
@kevinyamauchi can you close this in favor of the decision we made in #316? @ttung will implement the outlined in that issue which will solve our dtype problems

PR #297, opened 2018-06-26
docker: add Dockerfile which runs starfish by default

joshmoore:
Happy to also enable building of this in .travis if useful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,40 @@
-## Development dockerfile for starfish, primarily
-## focused on the running of tests but the executables
-## are installed to ~starfish/.local/bin
+## Dockerfile for starfish
+##
+## Default entrypoint is the starfish script,
+## but can also be used to run pytests.
 ##
 ## Examples:
+## --------
 ##
-##   docker build -t starfish .
-##   docker run --rm starfish -h
-##   docker run --rm -it --entrypoint=bash starfish
-##   docker run -e TEST_ISS_KEEP_DATA=true starfish -vsxk iss
+##   (1) Pull the centrally built image
+##   $ docker pull spacetx/starfish:latest
+##
+##   or, (2) build a local image named "spacetx/starfish"
+##   $ docker build -t spacetx/starfish .
+##
+##   (3) See the help for starfish
+##   $ docker run --rm spacetx/starfish -h
+##
+##   (4) Run starfish passing arguments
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
+## Development dockerfile for starfish, primarily
+## focused on the running of tests but the executables
+## are installed to ~starfish/.local/bin
+##
+## Examples:
+##
+##   docker build -t starfish .
```
ambrosejcarr: This is really great. How would you feel about adding a short explanation to each command? For those less familiar with docker, they might not know that these: (1) run help, (2) are equivalent to calling `starfish`, and (3) run verbose tests. 

Once this is built, what do you think about us pushing this up to quay or dockerhub and having it autobuild on new commits to master?

ambrosejcarr:
I haven't used travis to build Dockerfiles before -- what would that accomplish?

Dockerfile
joshmoore: More than happy to add the explanations. One question I had myself was whether you wanted the tests run by default or `starfish`. As for Dockerhub, by all means. Let me know to what extent you want me to create the org and/or repository.

joshmoore:
> I haven't used travis to build Dockerfiles before -- what would that accomplish?

If nothing else, it's testing the Dockerfile continues to build for all PRs. (Docker Hub will only tell you for well-defined branches). Additionally if interested, logic can be refactored out of `.travis.yml` in favor of `Dockerfile` so that users can make use of the steps locally.

Dockerfile
ambrosejcarr: I think `starfish` would be preferable. @ttung what do you think?

joshmoore:
Suggestions pushed.

ttung:
> Happy to also enable building of this in .travis if useful.

Does that mean .travis will build the docker image?  Or fire up docker and run things inside the docker instance?

joshmoore:
@ttung : both are an option. If you have a Dockerfile in the repo, I'd suggest at least building the image so that you know it's still working for all PRs.

joshmoore:
```
$ docker build -t spacetx/starfish .
...
Successfully tagged spacetx/starfish:latest
```

ttung:
lgtm.  i hope this doesn't add much to the build time. :)

joshmoore:
PR 285 was 6 minutes; this one 5. Apparently it's saving time... ;)

PR #302, opened 2018-07-05
RFC: Cleanup + Unification of starfish API + CLI 

### LINKED ISSUE #218: Research: CLI and API parameters are defined once and used for both calls.
Currently, starfish defines both API and CLI parameters, both of which can declare default paramters and help strings. 

We'd like to harmonize this, ideally by programming the information into the API and automatically generating the CLI. 

Tony thought it might be possible to do this with a decorator, but wants to experiment with how that could affect usability.

### LINKED ISSUE #292: Create a first draft of the Starfish API + CLI
Before we introduce the spaceTx community to starfish, we should attempt to standardize a coherent API that is consistent with the CLI. 

Currently, there are many methods that should be private that are exposed, which makes the API confusing. 

- pipeline components all have different execution methods (`filter`, `segment`, `run`, ...). These should be unified into one method (`run`, maybe?)
- many methods that are currently public should be made private, including several methods on the base classes
- object imports in `__init__.py` should be made clearer, and since we want to move towards an import pattern where we import the pipeline component, we should make all the implementing components private.

ambrosejcarr:
I've written up a doc that proposes a unification of the API and CLI: 

[https://docs.google.com/document/d/1l3f6sYhYcxanHDVhWtQzFrrz0--2YmNRXTivZ1gi4t4/edit#](https://docs.google.com/document/d/1l3f6sYhYcxanHDVhWtQzFrrz0--2YmNRXTivZ1gi4t4/edit#)

- The code in this PR contains an example of how we might adjust the CLI to use `click`, which is a small part of the proposal. See the doc for an explanation, examples, and the use patterns for this implementation. 

Possible implementation to fix #292 
Closes #218

### DIFF
--- a/click_starfish/click_starfish/__init__.py
+++ b/click_starfish/click_starfish/__init__.py
@@ -1,0 +1,1 @@
+from . import registration
--- a/click_starfish/click_starfish/_pipeline_component.py
+++ b/click_starfish/click_starfish/_pipeline_component.py
@@ -1,0 +1,2 @@
+class PipelineComponent:
+    pass
--- a/click_starfish/click_starfish/_util.py
+++ b/click_starfish/click_starfish/_util.py
@@ -1,0 +1,16 @@
+class Singleton(type):
+    """Metaclass to make TestComponent a singleton. Not strictly necessary."""
+
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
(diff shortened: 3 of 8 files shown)


ambrosejcarr:
I'm going to close this one for now. When I get to API/CLI synchronization I will re-open with a section added to the doc that has a proposal for the API update.

PR #303, opened 2018-07-06
Run the notebooks using the pip package on master

ttung:
This way, we can ensure that our users don't see broken behavior.


notebooks/subdir.mk
```diff
@@ -4,6 +4,7 @@ py_run_targets := $(addprefix run__, $(py_files))
 ipynb_validate_targets := $(addprefix validate__, $(ipynb_files))
 ipynb_regenerate_targets := $(addprefix regenerate__, $(addsuffix .ipynb, $(basename $(py_files))))
 py_regenerate_targets := $(addprefix regenerate__, $(addsuffix .py, $(basename $(ipynb_files))))
+PYTHON := python
```
ambrosejcarr: Why is this change necessary?

Issue #300, opened 2018-07-02
Errors when using example notebooks


ebouilhol:
Hi,
I tried a few notebooks and I can't rich the end of anyone I tried.
Let's focus on the allen_smFISH.ipynb here : 
At the step 5 

```
from starfish.pipeline.filter import Filter
s_clip = Filter.Clip(p_min=10, p_max=100, verbose=True)
s_clip.filter(s)
```

I get the following error : 

```
--------------------------------------------------------------------
AttributeError                     Traceback (most recent call last)
<ipython-input-16-6caa463fd5e6> in <module>()
      1 from starfish.pipeline.filter import Filter
      2 s_clip = Filter.Clip(p_min=10, p_max=100, verbose=True)
----> 3 s_clip.filter(s)

~/Documents/10_starfish/starfish/starfish/pipeline/filter/clip.py in filter(self, stack)
     75         """
     76         clip = partial(self.clip, p_min=self.p_min, p_max=self.p_max)
---> 77         stack.apply(clip, is_volume=self.is_volume, verbose=self.verbose)

AttributeError: 'Stack' object has no attribute 'apply'
```

I use python 3.6 on Ubuntu 18.04
I pulled the starfish repository on monday 02 July 11h.
I didn't change anything in the notebook's code, including the URL for the JSON, to be sure it don't come from my JSON.
(I get the same error if I try to launch it on other data)

Thanks for your help,

Emmanuel

ttung:
@ambrosejcarr do you know if the allen_SMFISH notebook is up-to-date?

ambrosejcarr:
I'll be able to take a look tomorrow on the plane. This issue suggests not. @ebouilhol sorry for the slow response, I'm out on vacation this week. 

I think as a part of "get the notebooks working again" I should write tests for each notebook similar to the iss one.

ambrosejcarr:
- Looks like when we extracted stack from filters the allen_smFISH notebook got missed. #307 provides a fix.
- MERFISH will take a bit longer to fix, as I'd like to finish #304 before wrangling the outputs. 
- [ISS](https://github.com/spacetx/starfish/blob/master/notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb) is currently working, if you want to look at a functional example @ebouilhol

ebouilhol:
> ISS is currently working, if you want to look at a functional example


Okay, thanks a lot, I switch on ISS then !

ambrosejcarr:
[...]

PR #307, opened 2018-07-08
BUG: fix allen_smFISH notebook and have it run on master

ambrosejcarr:
- Update notebook to pass ImageStack to PipelineComponents instead of Stack
- Fix a bug in the notebook where x and y coordinates were reversed when assigning the results of the spot calling on a max projection (trackpy.locate's order is `z, x, y` and `x, y`, not the skimage standard of `z, y, x`. 
- Fix a bug in stack._show_spots that would cause the function to fail if a 3d tensor was extracted with ImageStack.get_slice() -- the buggy code expected a tensor with dimensions >= 4.
- Remove the .skip file to cause it to run on master.


ambrosejcarr:
Closes #300

starfish/image/_stack.py
```diff
@@ -287,7 +288,14 @@ def show_stack(
         data, remaining_inds = self.get_slice(indices)
 
         # identify the dimensionality of data with all dimensions other than x, y linearized
-        n = numpy.dot(*data.shape[:-2])
+        if len(data.shape) == 3:
+            n = data.shape[0]
+        elif len(data.shape) > 3:
+            n = numpy.dot(*data.shape[:-2])
```
ttung: generically, can't this be replaced by:
`n = numpy.multiply(*data.shape[:2])`

Should handle all the possible dimension shapes.  Plus, don't we always expect the full Z,CH,HYB set for starfish?

ambrosejcarr: (1) Yea, I think we could replace dot with multiply.  

(2) No, we don't expect a 5d tensor. The user has the chance to select what they want to visualize in this case (output of `ImageStack.get_slice`). I made this fix because I realized that the original method 

Some examples: 
- all sections (ch, hyb, z) (5d tensor)
- just the tiles for a specific z-layer, across channels and hybs (4d tensor)
- just the tiles for a specific channel and hyb, but across z (3d tensor)

... I suppose it could also take a 2d tensor, but then this function is overkill (you can just use `imshow`)

ambrosejcarr: I think `np.product` is the correct choice, not multiply or dot.

ambrosejcarr: Good call, no -- we don't. Sorry, hadn't thought the change all the way through.

starfish/image/_stack.py
```diff
@@ -287,7 +288,12 @@ def show_stack(
         data, remaining_inds = self.get_slice(indices)
 
         # identify the dimensionality of data with all dimensions other than x, y linearized
-        n = numpy.dot(*data.shape[:-2])
+        if len(data.shape) >= 3:
+            n = numpy.product(data.shape[:-2])
+        else:
+            raise ValueError(
+                f'a stack with dimensionality >= 3 is required, the provided indexer produced a '
```
ttung: string quote style.

ambrosejcarr: Yep, these problems are all over the place, will fix all at once in #321

PR #315, opened 2018-07-10
[Easy] BUG: allow rescale to be False

ambrosejcarr:
- The default for rescale is False, which caused rescale to always run, because False is not None.

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -304,7 +304,7 @@ class ImageStack:
         if rescale and any((p_min, p_max)):
             raise ValueError('select one of rescale and p_min/p_max to rescale image, not both.')
 
-        elif rescale is not None:
+        elif rescale:
             print("Rescaling ...")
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
@ttung That's a good point, there are not many tests of the vis suite. 

To write them, I'd need to have a deeper understanding of how these vis scalings work, which would probably take me about 1/4 to 1/2 a day. Longer term it's definitely worth it, but my sense is that we have more important short term priorities. 

I'm tempted to make an issue and defer. What do you think?

ttung:
For `show()`, it doesn't seem like it's that necessary to build tests.  If it was easy, it would be nice to have something, but if it's not easy, then I guess we'll just leave it as is.

ambrosejcarr:
I think it's worth looking into how matplotlib writes tests. Maybe it's easy. I've never tried, so I don't know. :-)

Issue #48, opened 2018-02-07
Overhall stack API


dganguli:
The current Stack abstraction in io.stack.py conflates 'hybridization' images and 'auxilary' images. It's probably a good idea to separate these two under a common base class. This will:

1. Eliminate having to load (and potentially write) all 'hybridization' and 'auxilary' images when you only want to operate on, say, a single auxilary images

2. Eliminate the propagation of an 'org.json' file at each step of computation

3. Make the CLI more flexible and less tied to specific intermediary file formats

dganguli:
@ttung  can we add a 'linter' to make sure that peoples json files are well specified and point to real files? I'm running into this with collaborators self-made json files. This can also be a separate issue to file once the API experiences less churn as well.

ttung:
Once we standardize on the file format, we can do that.

ttung:
Also, would be great to build that tool that makes it easy to build a UI from a json-schema. :)

ttung:
https://github.com/mozilla-services/react-jsonschema-form might do.

ttung:
I think this is done.

PR #325, opened 2018-07-17
Remove write methods from Stack.

ttung:
Stacks are only read from the top of the experiment.

Connects to #48

### DIFF
--- a/starfish/io.py
+++ b/starfish/io.py
@@ -55,9 +55,6 @@ class Stack:
             self.org = json.load(fh)
 
         self.image = ImageStack.from_url(self.org['hybridization_images'], self.baseurl)
-        self._read_aux()
-
-    def _read_aux(self):
         for aux_key, aux_data in self.org['auxiliary_images'].items():
             self.auxiliary_images[aux_key] = ImageStack.from_url(aux_data, self.baseurl)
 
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
So basically, we read an experiment, and then to persist the results of filtering, we use `ImageStack` to write the results? If I'm wrong you can ignore the next paragraph. 

My one concern here is that we have no way of doing persistent "out of place" writes that, for example, save the raw data after applying filters. Do we have a larger plan to support that? I think users will probably want to be able to do that to, for example, save intermediates they could load up in FIJI or our nascent vis software.

ttung:
> My one concern here is that we have no way of doing persistent "out of place" writes that, for example, save the raw data after applying filters. Do we have a larger plan to support that? I think users will probably want to be able to do that to, for example, save intermediates they could load up in FIJI or our nascent vis software.

if you want to save an entire stack post filter, as a short-term hack, you can clobber over yourself.

experiment.json refers to hybridization.json.  `starfish filter -i hybridization.json -o hybridization.json`

I'm less certain of why one would need to save an entire experiment and load an experiment up in fiji though.  wouldn't one vizualize an ImageStack and not an experiment?  Alternatively, we can make the aux images a separate dimension of the same ImageStack (see thread @ https://spacetx.slack.com/archives/CAR91U354/p1531934898000413)

PR #333, opened 2018-07-21
[2/8] ENH: IntensityTable tracks the shape of the ImageStack it was created from

ambrosejcarr:
- Next PR enables direct conversion of ImageStack to IntensityTable by reshaping pixels. 
- This PR makes that reshaping process revertible by storing the shape of the ImageStack in the IntensityTable when it is created.

### DIFF
--- a/starfish/intensity_table.py
+++ b/starfish/intensity_table.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Union, Tuple
 
 import numpy as np
 import pandas as pd
@@ -75,7 +75,8 @@ class IntensityTable(xr.DataArray):
 
     @classmethod
     def empty_intensity_table(
-            cls, spot_attributes: pd.MultiIndex, n_ch: int, n_hyb: int
+            cls, spot_attributes: pd.MultiIndex, n_ch: int, n_hyb: int,
+            image_shape: Tuple[int, int, int]
     ) -> "IntensityTable":
         """Create an empty intensity table with pre-set axis whose values are zero
 
@@ -88,6 +89,8 @@ class IntensityTable(xr.DataArray):
             number of channels measured in the imaging experiment
         n_hyb : int
             number of hybridization rounds measured in the imaging experiment
+        image_shape : Tuple[int, int, int]
+            the shape (z, y, x) of the image from which features will be extracted
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


starfish/intensity_table.py
```diff
@@ -100,7 +103,14 @@ def empty_intensity_table(
         hyb_index = np.arange(n_hyb)
         data = np.zeros((spot_attributes.shape[0], n_ch, n_hyb))
         dims = (IntensityTable.Constants.FEATURES.value, Indices.CH.value, Indices.HYB.value)
-        return cls(data=data, coords=(spot_attributes, channel_index, hyb_index), dims=dims)
+        attrs = {'image_shape': image_shape}
+
+        intensity_table = cls(
+            data=data, coords=(spot_attributes, channel_index, hyb_index), dims=dims,
```
dganguli: hyb_index -> round_index

ambrosejcarr: You'll see lots of those -- the PR that makes these changes comes _after_ this one.

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -95,7 +96,9 @@ def _measure_spot_intensities(
         n_ch = stack.shape[Indices.CH]
         n_hyb = stack.shape[Indices.HYB]
         spot_attribute_index = dataframe_to_multiindex(spot_attributes)
-        intensity_table = IntensityTable.empty_intensity_table(spot_attribute_index, n_ch, n_hyb)
+        image_shape: Tuple[int, int, int] = stack.raw_shape[2:]
```
ttung: why do you want to call raw_shape?  if you have specific indices you care about (which you do), why not just call shape on the specific indices in the order you want it in?

specifically, if we rearrange the order of the axes in ImageStack, would this still work?

ttung:
Is this still needed if the IntensityTable -> ImageStack conversion method is removed?

ambrosejcarr:
@ttung I don't think it is. I'm going to make that adjustment first. Thanks for noticing!

starfish/pipeline/features/spots/detector/gaussian.py
ambrosejcarr: No, that would cause this to break. Good catch. We should be able to use the named shape.

ambrosejcarr:
@ttung Going to merge this so we can get the latter stuff in that @dganguli is depending on, but I think the comment above provides a reasonable strategy to excise this functionality later if we decide you'd rather go with the strategy that does not use `image_size`

PR #334, opened 2018-07-21
[3/8] ENH: add methods to reshape pixels to/from intensities

ambrosejcarr:
Add methods to:
- reshape ImageStack directly into an IntensityTable containing pixel intensities
- revert an IntensityTable containing pixel intensities back into an ImageStack

@ttung @dganguli There's a note at the bottom here questioning whether or not we want to support reshaping non-contiguous intensities (i.e. IntensityTables that contain _spots_) back into an ImageStack. My sense is no, since it would be a lossy reversion (we summarize spots as location, radius, intensity, not clear how to reconstruct original data). 

If we _don't_ support it, then I think having the method here is a bit weird, since it only works on one type of IntensityTable. Open to thoughts.

### DIFF
--- a/starfish/image/_stack.py
+++ b/starfish/image/_stack.py
@@ -1,25 +1,24 @@
 import collections
 import os
+import warnings
+from copy import deepcopy
 from functools import partial
 from itertools import product
 from typing import Any, Callable, Iterable, Iterator, List, Mapping, MutableSequence, Optional, Sequence, Tuple, Union
-import warnings
-from copy import deepcopy
 
 import numpy as np
 import pandas as pd
+from scipy.ndimage.filters import gaussian_filter
 from scipy.stats import scoreatpercentile
 from skimage import exposure
-from scipy.ndimage.filters import gaussian_filter
 from slicedimage import Reader, Writer, TileSet, Tile
 from slicedimage.io import resolve_path_or_url
 from tqdm import tqdm
 
 from starfish.constants import Coordinates, Indices
 from starfish.errors import DataFormatWarning
-from starfish.pipeline.features.spot_attributes import SpotAttributes
 from starfish.intensity_table import IntensityTable
-from starfish.munge import dataframe_to_multiindex
+from starfish.pipeline.features.spot_attributes import SpotAttributes
 
 
 class ImageStack:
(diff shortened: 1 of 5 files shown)


starfish/image/_stack.py
```diff
@@ -960,3 +961,91 @@ def from_numpy_array(cls, array: np.ndarray) -> "ImageStack":
...
+                intensities.sizes[Indices.CH],
+                intensities.sizes[Indices.HYB]])
+            data = data.transpose(4, 3, 0, 1, 2)
+            return ImageStack.from_numpy_array(data)
+
+        else:
+
+            raise NotImplementedError
+
+            # not implementing this right now, not clear it makes sense to.
```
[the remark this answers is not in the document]
ambrosejcarr: > perhaps this could be an interesting way to re-do ImageStack.show_spots?

How would this work? 

> It's unclear to me why you would want to go from pixel intensities -> an imagetack, other than for testing / simulation purposes

Great idea. Pending your explanation of the above, I'll nuke the method and move it to a testing module. 👍

starfish/image/_stack.py
```diff
@@ -960,3 +961,91 @@ def from_numpy_array(cls, array: np.ndarray) -> "ImageStack":
...
+        ----------
+        crop : Tuple[int, int, int]
+            number of pixels from the image borders (z, y, x) to ignore
+
+        Returns
+        -------
+
+        """
+        zmin, ymin, xmin = crop
+        zmax = self.shape['z'] - zmin
```
ttung: Do you want to use Indices.Z?

ttung: I also think this should be `zmax = self.shape['z'] - crop`

starfish/image/_stack.py
```diff
@@ -960,3 +961,91 @@ def from_numpy_array(cls, array: np.ndarray) -> "ImageStack":
...
+        Parameters
+        ----------
+        crop : Tuple[int, int, int]
+            number of pixels from the image borders (z, y, x) to ignore
+
+        Returns
+        -------
+
+        """
+        zmin, ymin, xmin = crop
```
ttung: Assert that size > 2 * crop?

starfish/image/_stack.py
```diff
@@ -960,3 +961,91 @@ def from_numpy_array(cls, array: np.ndarray) -> "ImageStack":
...
+        zmax = self.shape['z'] - zmin
+        ymax = self.shape['y'] - ymin
+        xmax = self.shape['x'] - xmin
+        data = self.numpy_array.transpose(2, 3, 4, 1, 0)  # (z, y, x, ch, hyb)
+        cropped_data = data[zmin:zmax, ymin:ymax, xmin:xmax, :, :]
+        intensity_data = cropped_data.reshape(-1, self.num_chs, self.num_hybs)  # (pixels, ch, hyb)
+        z = np.arange(zmin, zmax)
```
ttung: nit: i think it's better to set up *min, *max, and x/y/z before mucking around with the data.

starfish/image/_stack.py
```diff
@@ -960,3 +961,91 @@ def from_numpy_array(cls, array: np.ndarray) -> "ImageStack":
                     empty.set_slice({Indices.HYB: h, Indices.CH: c, Indices.Z: z}, view)
 
         return empty
+
+    def to_pixel_intensities(self, crop: Tuple[int, int, int]=(0, 0, 0)) -> IntensityTable:
```
ttung: Is this the right place for this method?  I kind of think this belongs more in IntensityTable than it does in ImageStack.

ambrosejcarr: Yea, that's where I originally had it. However, when I made the reversion method, that caused circular imports. So I put it here. 

Given that we're going to transfer the reversion method to the testing suite, we should be able to change this to `IntensityTable.from_image_Stack` 👍

starfish/image/_stack.py
ambrosejcarr: crop is a 3-tuple, so this doesn't work. I'm going to try this: 

```python
        crop_z, crop_y, crop_x = crop

        assert crop_z > 2 * self.shape[Coordinates.Z]
        assert crop_y > 2 * self.shape[Coordinates.Y]
        assert crop_x > 2 * self.shape[Coordinates.X]

        zmin = crop_z
        ymin = crop_y
        xmin = crop_x
        zmax = self.shape['z'] - crop_z
        ymax = self.shape['y'] - crop_y
        xmax = self.shape['x'] - crop_x
```

starfish/image/_stack.py
ambrosejcarr: Added a comment to make it clear that these are indices based on the mucked around data.

starfish/intensity_table.py
```diff
@@ -276,3 +278,54 @@ def synthetic_intensities(
         intensities[cls.Constants.GENE.value] = ('features', genes)
 
         return intensities
+
+    @classmethod
+    def from_image_stack(cls, image_stack, crop: Tuple[int, int, int]=(0, 0, 0)) -> "IntensityTable":
```
ttung: This sort of thing might bite us in the long run.  Is there a compelling reason to do a single crop variable that contains all the dimensions?  Why not pass in crop_x, crop_y, and crop_z individually?

ttung: Rationale: it is easy to mess up the order of the tuple.

ambrosejcarr: Got it, now I understand. That makes sense. I'll make this change.

ttung: If you want to be really daring, you could stick it after a '*' parameter, which would require callers to explicitly name the parameters. 

`from_image_stack(...., 0, 15, 12)` is less readable than `from_image_stack(...., crop_x=0, crop_y=15, crop_z=12)`

starfish/intensity_table.py
```diff
@@ -276,3 +278,58 @@ def synthetic_intensities(
...
         return intensities
+
+    @classmethod
+    def from_image_stack(cls, image_stack, crop_x: int=0, crop_y: int=0, crop_z: int=0) -> "IntensityTable":
+        """Generate an IntensityTable from all the pixels in the ImageStack
+
+        Parameters
+        ----------
+        crop_x : int
+            number of pixels to crop from x
```
ttung: This is potentially confusing because one could interpret this as "I will crop crop_x/2 pixels from the left and crop_x/2 pixels from the right, for a total of crop_x pixels."

starfish/test/image/test_imagestack.py
```diff
@@ -3,11 +3,14 @@
 
 from starfish.constants import Indices
 from starfish.image import ImageStack
+from starfish.intensity_table import IntensityTable
 # don't inspect pytest fixtures in pycharm
 # noinspection PyUnresolvedReferences
 from starfish.test.dataset_fixtures import (
     synthetic_intensity_table, synthetic_spot_pass_through_stack, loaded_codebook,
-    synthetic_dataset_with_truth_values, simple_codebook_json, simple_codebook_array)
+    synthetic_dataset_with_truth_values, simple_codebook_json, simple_codebook_array,
```
ttung: Stylistically, when we get to these long lists of imported symbols, I prefer to do one per line.

starfish/test/test_stack_intensity_table_reshaping.py
```diff
@@ -0,0 +1,45 @@
...
+    the created Imagestack is the same as the original
+    """
+    codebook, intensities, image = single_synthetic_spot
+    pixel_intensities = IntensityTable.from_image_stack(image, 0, 0, 0)
+    image_shape = (image.shape['z'], image.shape['y'], image.shape['x'])
+    image_from_pixels = pixel_intensities_to_imagestack(pixel_intensities, image_shape)
+    assert np.array_equal(image.numpy_array, image_from_pixels.numpy_array)
+
+
+def pixel_intensities_to_imagestack(intensities: IntensityTable, image_shape: Tuple[int, int, int]) -> "ImageStack":
```
ttung: No longer need the "s

starfish/test/test_stack_intensity_table_reshaping.py
```diff
@@ -0,0 +1,45 @@
...
+    """Re-create the pixel intensities from an IntensityTable
+
+    Parameters
+    ----------
+    intensities : IntensityTable
+        intensities to transform into an ImageStack
+    image_shape : Tuple[int, int, int]
```
ttung: this is under-documented.  z-y-x?

Issue #323, opened 2018-07-16
Move decode machinery to PipelineComponent


ambrosejcarr:
Right now it's in codebook, but that feels awkward. Extract it and make it back into one or more algorithms subclassing `PipelineComponent` and `DecoderAlgorithmBase`

ambrosejcarr:
Upon improving the generality of the normalization capabilities of `metric_decode`,  @dganguli has rescinded this request. Closing this.

PR #335, opened 2018-07-21
[4/8] ENH: update euclidean codebook with MERFISH filters and NearestNeighbor search

ambrosejcarr:
- Updates euclidean decode to work with the MERFISH requirements
- Swap out brute force search for NearestNeighbors
- add `min_intensity`, `max_distance`
- swap `np.nan` gene_ids to `None`. Both approaches are gross, but `np.nan` was being implicitly converted to string, so I find setting it as a string `None` more explicit. How to deal with `object` coordinate indices is on my list of things to talk to the xarray people about.

### DIFF
--- a/starfish/codebook.py
+++ b/starfish/codebook.py
@@ -1,16 +1,17 @@
-import functools
 import json
 import urllib.request
 import uuid
-from typing import Any, Dict, List, Optional, Sequence, Set, Union
+from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import numpy as np
 import pandas as pd
 import validators
 import xarray as xr
+from sklearn.neighbors import NearestNeighbors
 
 from starfish.constants import Indices, AugmentedEnum
 from starfish.intensity_table import IntensityTable
+from starfish.typing import Number
 
 
 class Codebook(xr.DataArray):
@@ -356,67 +357,134 @@ class Codebook(xr.DataArray):
         with open(filename, 'w') as f:
             json.dump(code_array, f)
 
-    def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
+    @staticmethod
+    def _normalize_features(array: Union["Codebook", IntensityTable], norm_order) \
+            -> Tuple[Union["Codebook", IntensityTable], np.ndarray]:
+        """unit normalize each feature of array
+
+        Parameters
+        ----------
+        array : Union[Codebook, IntensityTable]
+            codebook or intensity table containing (ch, r) features to normalize
+        norm_order : int
+            the norm to apply to each feature
+
+        See Also
+        --------
+        The available norms for this function can be found at the following link:
+        https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.linalg.norm.html
+
+        Returns
+        -------
+        Union[IntensityTable, Codebook] :
+            IntensityTable or Codebook containing normalized features
+        np.ndarray :
+            A 1 dimensional numpy array containing the feature norms
+
+        """
+        feature_traces = array.stack(traces=(Indices.CH.value, Indices.HYB.value))
+        norm = np.linalg.norm(feature_traces.values, ord=norm_order, axis=1)
+        array = array / norm[:, None, None]
+
+        # if a feature is all zero, the information should be spread across the channel
+        n = array.sizes[Indices.CH.value] * array.sizes[Indices.HYB.value]
+        partitioned_intensity = np.linalg.norm(np.full(n, fill_value=1 / n), ord=norm_order) / n
+        array = array.fillna(partitioned_intensity)
+
+        return array, norm
+
+    @staticmethod
+    def _approximate_nearest_code(
+            norm_codes: "Codebook", norm_intensities: xr.DataArray, metric: str,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """find the nearest code for each feature using the ball_tree approximate NN algorithm
+
+        Parameters
+        ----------
+        norm_codes : Codebook
+            codebook with each code normalized to unit length (sum = 1)
+        norm_intensities : IntensityTable
+            intensity table with each feature normalized to unit length (sum = 1)
+        metric : str
+            the sklearn metric string to pass to NearestNeighbors
+
+        Returns
+        -------
+        np.ndarray : metric_output
+            the output of metric applied to each feature closest code
+        np.ndarray : gene_ids
+            the gene that corresponds to each matched code
+
+        """
+        # TODO ambrosejcarr: see if features can be pre-masked to reduce NN calculations
+        linear_codes = norm_codes.stack(traces=(Indices.CH.value, Indices.HYB.value)).values
+        linear_features = norm_intensities.stack(
+            traces=(Indices.CH.value, Indices.HYB.value)).values
+
+        # reshape into traces
+        nn = NearestNeighbors(n_neighbors=1, algorithm='ball_tree', metric=metric).fit(linear_codes)
+        metric_output, indices = nn.kneighbors(linear_features)
+        gene_ids = np.ravel(norm_codes.indexes[Codebook.Constants.GENE.value].values[indices])
+
+        return np.ravel(metric_output), gene_ids
+
+    def metric_decode(
+            self, intensities: IntensityTable, max_distance: Number, min_intensity: Number,
+            norm: int, metric: str='euclidean'
+    ) -> IntensityTable:
         """Assign the closest gene by euclidean distance to each feature in an intensity table
+
+        Normalizes both the codes and the features to be unit vectors and finds the closest code
+        for each feature
 
         Parameters
         ----------
         intensities : IntensityTable
             features to be decoded
+        max_distance : Number
+            maximum distance between a feature and its closest code for which the coded gene will
+            be assigned.
+        min_intensity : Number
+            minimum intensity for a feature to receive a gene id
+        norm : int
+            the scipy.linalg norm to apply to normalize codes and intensities
+        metric : str
+            the sklearn metric string to pass to NearestNeighbors
+
+        See Also
+        --------
+        The available norms for this function can be found at the following link:
+        https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.linalg.norm.html
 
         Returns
         -------
         IntensityTable :
-            intensity table containing additional data variables for gene assignments and feature
-            qualities
+            intensity table containing additional data variables for gene assignments and metric
+            outputs
 
         """
-
-        def _min_euclidean_distance(observation: xr.DataArray, codes: Codebook) -> np.ndarray:
-            """find the code with the closest euclidean distance to observation
-
-            Parameters
-            ----------
-            observation : xr.DataArray
-                2-dimensional DataArray of shape (n_ch, n_hyb)
-            codes :
-                Codebook containing codes to compare to observation
-
-            Returns
-            -------
-            np.ndarray :
-                1-d vector containing the distance of each code to observation
-
-            """
-            squared_diff = (codes - observation) ** 2
-            code_distances = np.sqrt(squared_diff.sum((Indices.CH, Indices.HYB)))
-            # order of codes changes here (automated sorting on the reshaping?)
-            return code_distances
-
         # normalize both the intensities and the codebook
-        norm_intensities = intensities.groupby(IntensityTable.Constants.FEATURES.value).apply(
-            lambda x: x / x.sum())
-        norm_codes = self.groupby(Codebook.Constants.GENE.value).apply(lambda x: x / x.sum())
+        norm_intensities, norms = self._normalize_features(intensities, norm_order=norm)
+        norm_codes, _ = self._normalize_features(self, norm_order=norm)
 
-        # calculate pairwise euclidean distance between codes and features
-        func = functools.partial(_min_euclidean_distance, codes=norm_codes)
-        distances = norm_intensities.groupby(IntensityTable.Constants.FEATURES.value).apply(func)
+        # mask low intensity features
+        intensity_mask = np.where(norms < min_intensity)[0]
 
-        # calculate quality of each decoded spot
-        qualities = 1 - distances.min(Codebook.Constants.GENE.value)
-        qualities_index = pd.Index(qualities)
+        metric_outputs, gene_ids = self._approximate_nearest_code(
+            norm_codes, norm_intensities, metric=metric)
 
-        # identify genes associated with closest codes
-        closest_code_index = distances.argmin(Codebook.Constants.GENE.value)
-        gene_ids = distances.indexes[
-            Codebook.Constants.GENE.value].values[closest_code_index.values]
+        exceeds_distance = np.where(metric_outputs > max_distance)[0]
         gene_index = pd.Index(gene_ids)
+
+        # remove genes associated with distant codes
+        gene_index.values[exceeds_distance] = 'None'
+        gene_index.values[intensity_mask] = 'None'
 
         # set new values on the intensity table in-place
         intensities[IntensityTable.Constants.GENE.value] = (
             IntensityTable.Constants.FEATURES.value, gene_index)
-        intensities[IntensityTable.Constants.QUALITY.value] = (
-            IntensityTable.Constants.FEATURES.value, qualities_index)
+        intensities[IntensityTable.Constants.DISTANCE.value] = (
+            IntensityTable.Constants.FEATURES.value, metric_outputs)
 
         return intensities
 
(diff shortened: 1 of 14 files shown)


starfish/codebook.py
```diff
@@ -372,51 +426,26 @@ def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
...
-                2-dimensional DataArray of shape (n_ch, n_hyb)
-            codes :
-                Codebook containing codes to compare to observation
+        intensity_mask = intensities.mask_low_intensity_features(min_intensity)
 
-            Returns
-            -------
-            np.ndarray :
-                1-d vector containing the distance of each code to observation
+        # normalize both the intensities and the codebook
```
dganguli: from an offline convo -- i suspect this is one potential source of your MERFISH results not more closely matching the MERFISH results on master. in the MERFISH pipeline, codes and intensities are re-scaled by their L2 norms, not by their sums. In other words, these re-scaled values have unit _magnitudes_ not unit _sums_. The old code went something like this:

```    @staticmethod
    def _normalize_barcodes(codebook: pd.DataFrame) -> np.ndarray:
        # parse barcode into numpy array and normalize by l2_norm
        codes = np.array([np.array([int(d) for d in c]) for c in codebook.barcode])
        codes_l2_norm = np.linalg.norm(codes, axis=1, ord=2)
        weighted_codes = codes / codes_l2_norm[:, None]
        return weighted_codes

    # TODO ambrosejcarr fill in these values + add docs
    @staticmethod
    def _parse_pixel_traces(encoded: pd.DataFrame) -> Tuple[np.ndarray, np.ndarray]:
        # parse spots into pixel traces, normalize and filter
        df = encoded.loc[:, ['spot_id', 'barcode_index', 'intensity']]
        # TODO this assumes that bits are sorted [they are, currently]
        pixel_traces = df.pivot(index='spot_id', columns='barcode_index', values='intensity')
        pixel_traces = pixel_traces.values
        pixel_traces_l2_norm = np.linalg.norm(pixel_traces, axis=1, ord=2)
        ind = pixel_traces_l2_norm > 0
        pixel_traces[ind, :] = pixel_traces[ind, :] / pixel_traces_l2_norm[ind, None]
        return pixel_traces, pixel_traces_l2_norm
```

starfish/codebook.py
```diff
@@ -356,13 +358,65 @@ def to_json(self, filename: str) -> None:
...
+        linear_features = norm_intensities.values.reshape(norm_intensities.shape[0], -1)
+        # reshape into traces
+        nn = NearestNeighbors(n_neighbors=1, algorithm='ball_tree').fit(linear_codes)
+        distances, indices = nn.kneighbors(linear_features)
+        gene_ids = np.ravel(norm_codes.indexes[Codebook.Constants.GENE.value].values[indices])
+        qualities = 1 - np.ravel(distances)
+
+        return qualities, gene_ids
+
+    def decode_euclidean(
```
dganguli: I still think it's weird to have codebooks do the decoding. Is the plan to move this into a Decoder in a later PR, along with the other decode method? I also think the helper functions this uses (e.g., approximate_nearest_code, unit_normalize, etc.) should move out of the decoder as well.

starfish/codebook.py
dganguli: in fact, you can generalize the above code to do 'unit_normalize' by changing the 'ord' parameter in np.linalg.norm to 1. hence the generalization is incurred by writing a method called norm, which exposes a parameter ord, which is then imply passed to np.linalg.norm.

starfish/intensity_table.py
```diff
@@ -276,3 +276,14 @@ def synthetic_intensities(
         intensities[cls.Constants.GENE.value] = ('features', genes)
 
         return intensities
+
+    def mask_low_intensity_features(self, intensity_threshold):
```
dganguli: this is another potential source of variation between MERFISH results on this branch and master. the MERFISH decoder in master discards pixel vectors with an _L2 norm_ lower than a specified threshold: 

```pixel_traces, pixel_traces_l2_norm = self._parse_pixel_traces(encoded)
        local_magnitude = np.reshape(pixel_traces_l2_norm, img_size)
        decoded_img[local_magnitude < self.magnitude_threshold] = 0
```

whereas the code in this branch filters out pixel vectors below a specified _mean intensity_

dganguli: i do, however like the masking approach (to ignore these vectors) rather than the set things to zero and do all the nearest neighbor calculations anyways.

starfish/codebook.py
ambrosejcarr: Nice, thanks for the comments. I'll follow up on this and see if it decreases the discrepancy (my plan was to attack that in person this week)

starfish/codebook.py
ambrosejcarr: Yep, that's still the plan. I made #323 but figured MERFISH was more important to prioritize. There's quite a bit of tech debt that I need to work through from the switch over to xarray based objects.

starfish/intensity_table.py
ambrosejcarr: Unfortunately as coded it does the nearest neighbor calculations so there is no efficiency gain... but I bet I might be able to pre-mask the way you suggest.

starfish/codebook.py
```diff
@@ -356,13 +358,65 @@ def to_json(self, filename: str) -> None:
         with open(filename, 'w') as f:
             json.dump(code_array, f)
 
-    def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
+    @staticmethod
+    def _unit_normalize(array: xr.DataArray) -> xr.DataArray:
+        """unit normalize each feature of array"""
+        total_intensities = array.sum(dim=[Indices.CH.value, Indices.HYB.value])
+        array = array / total_intensities
```
ttung: What if the array is all zeros?  :)

starfish/codebook.py
```diff
@@ -356,13 +358,65 @@ def to_json(self, filename: str) -> None:
...
+
+        # if a code is all zero, the unit intensity should be spread across the channel
```
ttung: this comment is weird because this is used for both normalizing the intensity table and the codebook.  does the same policy apply to both?  guessing yes, given the nature of the code.

starfish/codebook.py
```diff
@@ -425,7 +454,7 @@ def decode_per_hyb_max(self, intensities: IntensityTable) -> IntensityTable:
 
         Notes
         -----
-        If no code matches the per-channel max of a feature, it will be assigned np.nan instead
+        If no code matches the per-channel max of a feature, it will be assigned 'None' instead
```
ttung: personally, I think this is a bit odd.  `None` is a specific pythonic concept, but this is a string "None".

starfish/codebook.py
```diff
@@ -372,51 +426,26 @@ def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
...
-
-        # calculate quality of each decoded spot
-        qualities = 1 - distances.min(Codebook.Constants.GENE.value)
-        qualities_index = pd.Index(qualities)
-
-        # identify genes associated with closest codes
-        closest_code_index = distances.argmin(Codebook.Constants.GENE.value)
-        gene_ids = distances.indexes[
-            Codebook.Constants.GENE.value].values[closest_code_index.values]
+        exceeds_distance = np.where(qualities < 1 - max_distance)[0]
```
ttung: So this filtering is still happening after all the calculations.  I guess this is fine, but this is yet another case that demonstrates why numpy needs a futures system. :)

starfish/test/dataset_fixtures.py
```diff
@@ -123,7 +127,7 @@ def loaded_codebook(simple_codebook_json):
 
 @pytest.fixture(scope='function')
 def euclidean_decoded_intensities(small_intensity_table, loaded_codebook):
-    decoded_intensities = loaded_codebook.decode_euclidean(small_intensity_table)
+    decoded_intensities = loaded_codebook.decode_euclidean(small_intensity_table, max_distance=0)
```
ttung: if max_distance==0, doesn't that mean everything is pruned from the results?  if so, can we assert that the decoded intensity table is empty?

starfish/codebook.py
```diff
@@ -1,17 +1,19 @@
...
 import numpy as np
 import pandas as pd
 import validators
 import xarray as xr
+from sklearn.neighbors import NearestNeighbors
 
 from starfish.constants import Indices, AugmentedEnum
 from starfish.intensity_table import IntensityTable
 
+Number = Union[int, float]
```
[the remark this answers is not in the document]
ambrosejcarr: Yep. It will be. I've made a `starfish.typing` module and moved it there.

starfish/codebook.py
ambrosejcarr: The function will propagate np.NaN throughout the array. The lines below your comment deal with the result. For unit normalization, which I mistakenly thought that MERFISH was using, this is a really easy fix -- I ensure the feature equals one by filling each (ch, r) combination with the correct fraction of `1`: 

```
        # if a code is all zero, the unit intensity should be spread across the channel
        partitioned_intensity = 1 / np.product(array.shape[1:])
        array = array.fillna(partitioned_intensity)
```

I will need to think about what to do in the 2-norm case.

starfish/codebook.py
ambrosejcarr: So I really _really_ want to remove this altogether and I'm debating mapping all these values to ints internally. I'm going to wait to do something until I've poked at the xarray devs a bit because the string index handling is something that I don't like and I'm hoping there is a solution I'm not seeing. 

For now, I propose replacing `None`  with a constant in Features (maybe `no_target`?). I'll edit that into #340 when this stack gets merged.

starfish/test/dataset_fixtures.py
ambrosejcarr: Close, it means all the matches need to be perfect, so we can't make that assertion.

ttung: Ah!

Is it worth asserting on any of the resulting IntensityTables?   approx size?  i don't know what else is easily assertable...

ambrosejcarr: Added size assertion. 👍

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        np.ndarray :
+            A 1 dimensional numpy array containing the feature norms
+
+        """
+        feature_traces = array.stack(traces=(Indices.CH.value, Indices.HYB.value))
+        norm = np.linalg.norm(feature_traces.values, ord=norm, axis=1)
+        array = array / norm[:, None, None]
+
+        # if a feature is all zero, the information should be spread across the channel
+        partitioned_intensity = 1 / np.product(array.shape[1:])
```
dganguli: this only works for an L1 norm. to generalize, you need something like ```np.linalg.norm(np.ones(N)*(1/N),ord=norm)``` where N is the length of your pixel trace

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        Returns
+        -------
+        np.ndarray : qualities
+            the quality of each feature's match to its closest code (equal to 1 - distance)
+        np.ndarray : gene_ids
+            the gene that corresponds to each matched code
+
+        """
+        # TODO ambrosejcarr: see if features can be pre-masked to reduce NN calculations
+        linear_codes = norm_codes.values.reshape(norm_codes.shape[0], -1)
```
dganguli: these reshape codes should be done with xarray.stack

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        linear_features = norm_intensities.values.reshape(norm_intensities.shape[0], -1)
+        # reshape into traces
+        nn = NearestNeighbors(n_neighbors=1, algorithm='ball_tree').fit(linear_codes)
```
dganguli: this should expose the sklearn supported distance metrics.

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        distances, indices = nn.kneighbors(linear_features)
+        gene_ids = np.ravel(norm_codes.indexes[Codebook.Constants.GENE.value].values[indices])
+        qualities = 1 - np.ravel(distances)
```
dganguli: i vote we just report whatever distances is

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        np.ndarray : gene_ids
+            the gene that corresponds to each matched code
+
+        """
+        # TODO ambrosejcarr: see if features can be pre-masked to reduce NN calculations
+        linear_codes = norm_codes.values.reshape(norm_codes.shape[0], -1)
+        linear_features = norm_intensities.values.reshape(norm_intensities.shape[0], -1)
+        # reshape into traces
+        nn = NearestNeighbors(n_neighbors=1, algorithm='ball_tree').fit(linear_codes)
+        distances, indices = nn.kneighbors(linear_features)
```
dganguli: i think this should be renamed to metric_values or something that generalizes to distance and other metrics supported by nn.kneighbors.

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        gene_ids = np.ravel(norm_codes.indexes[Codebook.Constants.GENE.value].values[indices])
+        qualities = 1 - np.ravel(distances)
+
+        return qualities, gene_ids
+
+    def decode_euclidean(
+            self, intensities: IntensityTable, max_distance: Number, min_intensity: Number=0,
```
dganguli: this should be a tuple of (min, max) and we should filter on distances that are outside of this range.

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+            norm: int=2, return_norms: bool=False,
```
dganguli: we should move away from exposing defaults -- methods are too variable for now to expose meaningful defaults imo

starfish/codebook.py
```diff
@@ -371,52 +460,27 @@ def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
             qualities
 
         """
+        # normalize both the intensities and the codebook
+        norm_intensities, norms = self._normalize_features(intensities, norm=norm)
```
dganguli: from an offline convo w/ @kevinyamauchi -- expose norm as a tuple? if it's (None, None) then don't normalize. If it's (2, 1) then you can normalize features and codebook w/ L2 and L1 norm respectively -- even thought that would be weird.

starfish/codebook.py
```diff
@@ -371,52 +460,27 @@ def decode_euclidean(self, intensities: IntensityTable) -> IntensityTable:
...
-
-        # calculate quality of each decoded spot
-        qualities = 1 - distances.min(Codebook.Constants.GENE.value)
-        qualities_index = pd.Index(qualities)
-
-        # identify genes associated with closest codes
-        closest_code_index = distances.argmin(Codebook.Constants.GENE.value)
-        gene_ids = distances.indexes[
-            Codebook.Constants.GENE.value].values[closest_code_index.values]
+        exceeds_distance = np.where(qualities < 1 - max_distance)[0]
```
dganguli: careful about signs and inequality directions after you make he switch from quality to distance.

starfish/intensity_table.py
```diff
@@ -333,3 +333,14 @@ def from_image_stack(cls, image_stack, crop_x: int=0, crop_y: int=0, crop_z: int
...
 
         return IntensityTable.from_spot_data(intensity_data, spot_attributes, image_size)
+
+    def mask_low_intensity_features(self, intensity_threshold):
+        """return the indices of features that have average intensity below intensity_threshold"""
+        mask = np.where(
+            self.mean([Indices.CH.value, Indices.HYB.value]).values < intensity_threshold)[0]
+        return mask
+
+    def mask_small_features(self, size_threshold):
```
dganguli: size_threshold should be a tuple that exposes min and max sizes

starfish/codebook.py
```diff
@@ -356,13 +357,101 @@ def to_json(self, filename: str) -> None:
...
+        -------
+        Union[IntensityTable, Codebook] :
+            IntensityTable or Codebook containing normalized features
+        np.ndarray :
+            A 1 dimensional numpy array containing the feature norms
+
+        """
+        feature_traces = array.stack(traces=(Indices.CH.value, Indices.HYB.value))
+        norm = np.linalg.norm(feature_traces.values, ord=norm, axis=1)
+        array = array / norm[:, None, None]
```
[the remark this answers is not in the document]
ambrosejcarr: Nope. :-)

starfish/codebook.py
ambrosejcarr: I think perhaps this was supposed to refer to the size threshold, not the distance or intensity ones.

starfish/codebook.py
ambrosejcarr: I think this deserves a bit more thought, I don't find that API very intuitive.

starfish/codebook.py
ambrosejcarr: Spoke with Deep and I think #347 describes a solution for this.

Issue #305, opened 2018-07-08
ENH: Update LocalMaxPeakFinder to generate IntensityTable


ambrosejcarr:
Currently it generates SpotAttributes and EncoderTable. Update this method.

ambrosejcarr:
@berl I've been thinking about what it will take to update `LocalMaxPeakFinder` to use the `Codebook` and `IntensityTable` objects and generalize across other assays in the `starfish` codebase.  

You had previously mentioned that it was important to do per-channel detection, thus ignoring the intensities of other channels at the same coordinate and protecting against crowding. I wanted to confirm that the above request is still important for you. If that's the case, and I were designing this module for smFISH, the best way to support this would be to concatenate the features found in each channel into one long IntensityTable. 

However, this does not generalize well to experiments with multiple hybridization rounds, or for experiments that want to compare across channels. If I were designing this for generality, I would max project over channels and hybridization rounds to get all the spots, then determine their intensities across all the 2d image tiles. 

In the "generalized" approach above, we could still easily do per-channel decoding by masking out other channels. However, there is a risk of creating hybrid spots across channels (e.g. where two spots overlap imperfectly, presumably an example of local crowding). In these cases the centroid, eccentricity, and radius of the spot would all be measured incorrectly, which could lead to improper filtering. 

Perhaps worth mentioning:`skimage` upgraded `blob_log` earlier this year to support 3d spot detection, so that could *also* be used to do 3d detection. It might be worth investigating the extent to which these two approaches are interchangeable. 

Tagging @dganguli for his perspective.

berl:
> You had previously mentioned that it was important to do per-channel detection, thus ignoring the intensities of other channels at the same coordinate and protecting against crowding. I wanted to confirm that the above request is still important for you

definitely still important for me and probably also important for Simone and any other non-barcoded assay. also, even the barcoded folks (MERFISH, SeqFISH) as published have used non-barcoded and barcoded assays in combination to span a larger dynamic range of expression. 

> If I were designing this for generality, I would max project over channels and hybridization rounds to get all the spots, then determine their intensities across all the 2d image tiles.

this will work for extremely sparse data, like what's shown in the recent CODEX preprint.  But it wouldn't work for multiround smFISH, at least not with many of the markers we use routinely. So what's the solution?

>  If that's the case, and I were designing this module for smFISH, the best way to support this would be to concatenate the features found in each channel into one long IntensityTable.
[...]

PR #336, opened 2018-07-21
[5/8] ENH: adapt connected-components pixel merging to IntensityTable

ambrosejcarr:
- Adds connected components pixel merging to IntensityTable

I've left some notes in the method string that discuss how we might implement this for non-contiguous methods. If the decision is to not support this, I'll remove that text prior to merging.

### DIFF
--- a/starfish/intensity_table.py
+++ b/starfish/intensity_table.py
@@ -1,12 +1,12 @@
-from typing import Union, Tuple
 from itertools import product
+from typing import Union, Tuple
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 
-from starfish.munge import dataframe_to_multiindex
 from starfish.constants import Indices, AugmentedEnum
+from starfish.munge import dataframe_to_multiindex
 
 
 class IntensityTable(xr.DataArray):
@@ -345,3 +345,7 @@ class IntensityTable(xr.DataArray):
         mask = np.where(self.coords.features[self.SpotAttributes.RADIUS.value] < min_size)[0]
         mask |= np.where(self.coords.features[self.SpotAttributes.RADIUS.value] > max_size)[0]
         return mask
+
+    def _intensities_from_regions(self, props, reduce_op='max') -> "IntensityTable":
+        """turn regions back into intensities by reducing over the labeled area"""
+        raise NotImplementedError
--- a/starfish/pipeline/features/pixels/combine_adjacent_features.py
+++ b/starfish/pipeline/features/pixels/combine_adjacent_features.py
@@ -1,0 +1,129 @@
+from typing import List, Tuple, Union
+
+import numpy as np
+import pandas as pd
+from skimage.measure import regionprops, label
+
+from starfish.munge import dataframe_to_multiindex
+from starfish.constants import Indices
+from starfish.intensity_table import IntensityTable
+
+Number = Union[int, float]
+
+
+def combine_adjacent_features(
+        intensities: IntensityTable, min_area: Number, max_area: Number, connectivity: int=2,
+) -> Tuple[IntensityTable, List, np.ndarray, np.ndarray]:
+    """
+
+    Parameters
+    ----------
+    intensities : IntensityTable
+        pixel-valued IntensityTable
+    min_area : Number
+        discard spots whose total pixel area is less than this value
+    max_area : Number
+        discard spots whose total pixel area is greater than this value
+    connectivity : int
+        the connectivity allowed when constructing connected components. Intuitively it can be
+        thought of as the number of square steps allowed between pixels that are considered
+        connected. 1 enforces adjacent pixels. 2 (default) allows diagonal pixels in the same
+        plane. 3 allows diagonal pixels in 3d.
+
+    Returns
+    -------
+    IntensityTable :
+        spot intensities
+    List[Object] :
+        region properties for each spot (see skimage.measure.regionprops)
+    np.ndarray[int] :
+        label image wherein each connected component (spot) is coded with a different
+        integer
+    np.ndarray[int] :
+        decoded image wherein each target is coded as a different integer. Intended for visualization
+
+    """
+    # None needs to map to zero, non-none needs to map to something else.
+    int_to_gene = dict(
+        zip(range(1, np.iinfo(np.int).max),
+            set(intensities.gene_name.values) - {'None'}))
+    int_to_gene[0] = 'None'
+    gene_to_int = {v: k for (k, v) in int_to_gene.items()}
+
+    # map genes to ints
+    gene_list = [gene_to_int[g] for g in intensities.coords[intensities.Constants.GENE.value].values]
+    gene_array = np.array(gene_list)
+
+    # reverses the linearization that was used to construct the IntensityTable from the ImageStack
+    decoded_image: np.ndarray = gene_array.reshape(intensities.attrs['image_shape'])
+
+    # label each pixel according to its component
+    label_image: np.ndarray = label(decoded_image, connectivity=connectivity)
+
+    # calculate the mean value grouped by (ch, round) across all pixels of a connected component
+    pixel_labels = label_image.reshape(-1)
+    intensities['spot_id'] = (IntensityTable.Constants.FEATURES.value, pixel_labels)
+    mean_pixel_traces = intensities.groupby('spot_id').mean(IntensityTable.Constants.FEATURES.value)
+    mean_distances = intensities[IntensityTable.Constants.DISTANCE.value].groupby('spot_id').mean(
+        IntensityTable.Constants.FEATURES.value)
+    mean_pixel_traces[IntensityTable.Constants.DISTANCE.value] = (
+        'spot_id',
+        np.ravel(mean_distances)
+    )
+
+    # "Zero" spot id is background in the label_image. Remove it.
+    # TODO ambrosejcarr: can we replace zero with nan above to get around this?
+    mean_pixel_traces = mean_pixel_traces.loc[mean_pixel_traces['spot_id'] > 0]
+    props: List = regionprops(np.squeeze(label_image))
+
+    # calculate spots and drop ones that fail the area threshold
+    spots = []
+    passes_filter = np.ones_like(props, dtype=np.bool)  # default is passing (ones)
+    for i, spot_property in enumerate(props):
+        # TODO ambrosejcarr: add tests for min/max area when fixing masking
+        if min_area <= spot_property.area < max_area:
+
+            # because of the above skimage issue, we need to support both 2d and 3d properties
+            if len(spot_property.centroid) == 3:
+                spot_attrs = {
+                    'z': int(spot_property.centroid[0]),
+                    'y': int(spot_property.centroid[1]),
+                    'x': int(spot_property.centroid[2])
+                }
+            else:  # data is 2d
+                spot_attrs = {
+                    'z': 0,
+                    'y': int(spot_property.centroid[0]),
+                    'x': int(spot_property.centroid[1])
+                }
+
+            # we're back to 3d or fake-3d here
+            gene_index = decoded_image[spot_attrs['z'], spot_attrs['y'], spot_attrs['x']]
+            spot_attrs['gene_name'] = int_to_gene[gene_index]
+            spot_attrs['radius'] = spot_property.equivalent_diameter / 2
+            spots.append(spot_attrs)
+        else:
+            passes_filter[i] = 0  # did not pass radius filter
+
+    # filter intensities
+    mean_pixel_traces = mean_pixel_traces.loc[passes_filter]
+
+    # now I need to make an IntensityTable from this thing.
+    spots_df = pd.DataFrame(spots)
+    spots_df[IntensityTable.Constants.DISTANCE.value] = \
+        mean_pixel_traces[IntensityTable.Constants.DISTANCE.value]
+
+    # create new indexes for the output IntensityTable
+    spots_index = dataframe_to_multiindex(spots_df)
+    channel_index = mean_pixel_traces.indexes[Indices.CH]
+    round_index = mean_pixel_traces.indexes[Indices.HYB]
+
+    # create the output IntensityTable
+    dims = (IntensityTable.Constants.FEATURES.value, Indices.CH.value, Indices.HYB.value)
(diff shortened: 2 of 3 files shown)


starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+        int_to_gene[0] = 'None'
+        gene_to_int = {v: k for (k, v) in int_to_gene.items()}
+
+        # map genes to ints
+        gene_list = [gene_to_int[g] for g in self.coords[self.Constants.GENE.value].values]
+        gene_array = np.array(gene_list)
+        decoded_image = gene_array.reshape(self.attrs['image_shape'])  # reverse linearization
+
+        # label the image and extract max intensity across each feature
+        label_image = label(decoded_image, connectivity=2)
```
dganguli: expose connectivity as a default parameter that defaults to 2?

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+                raise NotImplementedError
+
+        Returns
+        -------
+        IntensityTable :
+            spot intensities
+        List[
+
+        """
+        if not assume_contiguous:
```
dganguli: yeah i don't think we should support this now.

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+            tiles: List[np.ndarray] = []
+            for prop in props:
+                # extract the relevant area of the table and apply the reduce function
+                raise NotImplementedError
+
+        Returns
+        -------
+        IntensityTable :
+            spot intensities
+        List[
```
dganguli: forgot to finish all the things this returns?

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+        channel_index = [0]
+        round_index = [0]
+
+        # Right now we've eliminated channels and hybs from this tensor, which is not great
+        # but this is therefore a features x 1 x 1 tensor.
+        intensities = np.array(intensities).reshape(-1, 1, 1)
+        dims = (IntensityTable.Constants.FEATURES.value, Indices.CH.value, Indices.HYB.value)
+        attrs = {'image_shape': self.attrs['image_shape']}
+
+        intensity_table = IntensityTable(
```
dganguli: doesn't this need to extend a SpotAttributesTable somewhere?

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+        gene_to_int = {v: k for (k, v) in int_to_gene.items()}
+
+        # map genes to ints
+        gene_list = [gene_to_int[g] for g in self.coords[self.Constants.GENE.value].values]
+        gene_array = np.array(gene_list)
+        decoded_image = gene_array.reshape(self.attrs['image_shape'])  # reverse linearization
+
+        # label the image and extract max intensity across each feature
+        label_image = label(decoded_image, connectivity=2)
+        intensities = self.max(dim=[Indices.CH, Indices.HYB])  # todo can use mean here too
```
dganguli: interesting. if i'm reading this right, for each 'pixel' you record it's maximum intensity across C and H. then, as pixel vectors that decode to the same spot are merged into a 'spot', we record the max of these max intensities across pixels in a spot.  I think the max of the max of a bunch of L2 normalized vectors is an odd statistic to report. the max of the mean, however, feels more natural. perhaps we can expose an option here to report the max or the mean that defaults to the mean to be parallel with the GaussianSpotDetector? How do we handle this in the LocalMaxPeakFinder?

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+
+        # now I need to make an IntensityTable from this thing.
+        spots_df = pd.DataFrame(spots)
+
+        # construct coordinate indices
+        spots_index = dataframe_to_multiindex(spots_df)
+        channel_index = [0]
+        round_index = [0]
+
+        # Right now we've eliminated channels and hybs from this tensor, which is not great
```
dganguli: Yeah. This is indeed weird. Here's how Jeff handles it. For each 'spot' in this table, he records what he calls the 'mean pixel trace'. In other words, IntensityTable.group_by('spot_id').mean(# across C and H ) to return a C X H vector for each spot_id. I've found this extremely useful for inspecting decoded results. Analagously, for the input IntensityTable (before combining adjacent features) you could record a spot_id value for each pixel. In fact, from this table, you could easily do the group_by operation described above and it's just a matter of dragging a SpotAttributes object around for the ride.

starfish/intensity_table.py
ambrosejcarr: Yikes, sorry.

starfish/intensity_table.py
ambrosejcarr: Awesome, this is exactly what I wanted this to output but I couldn't think of a way to get it coded quickly and wanted to get it out for review (which I think has been helpful). 

I'll make this change.

starfish/intensity_table.py
ambrosejcarr: > interesting. if i'm reading this right, for each 'pixel' you record it's maximum intensity across C and H. then, as pixel vectors that decode to the same spot are merged into a 'spot', we record the max of these max intensities across pixels in a spot.

You read this correctly. 

> perhaps we can expose an option here to report the max or the mean that defaults to the mean to be parallel with the GaussianSpotDetector? How do we handle this in the LocalMaxPeakFinder?

LocalMaxPeakFinder reports the max intensity of the spot, but it's not general to multiplex assays (see discussion in #305). It might be a good idea to hash out exactly what API we want to expose for spot finders.

starfish/intensity_table.py
ambrosejcarr: Discussed in slack -- it does not, that will be deprecated upon completion of #305. 👍

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
...
+        props_passing_filter = []
+
+        # we're dropping spots that fail filters
+        for spot_property in props:
+            if spot_property.area >= area_threshold:
+
+                # because of the above skimage issue, we need to support both 2d and 3d properties
+                if len(spot_property.centroid) == 3:
+                    spot_attrs = {
+                        'z': int(spot_property.centroid[0]),
```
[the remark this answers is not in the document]
ambrosejcarr: Yea. I went through and cleaned up all of these in the last PR (#340) -- sorry, it makes for awkward reviewing. 

I'll be extra thorough there, promise!

starfish/intensity_table.py
```diff
@@ -291,3 +293,123 @@ def mask_small_features(self, size_threshold):
     def _intensities_from_regions(self, props, reduce_op='max') -> "IntensityTable":
         """turn regions back into intensities by reducing over the labeled area"""
         raise NotImplementedError
+
+    def combine_adjacent_features(
+            self, area_threshold, assume_contiguous=True
+    ) -> Tuple["IntensityTable", List, np.ndarray, np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
ambrosejcarr: +10 million. 

I vote we upgrade as soon as our dependencies declare that they're supporting it.

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+
+from starfish.munge import dataframe_to_multiindex
+from starfish.constants import Indices
+from starfish.intensity_table import IntensityTable
+
+Number = Union[int, float]
+
+
+def combine_adjacent_features(
+        intensities: IntensityTable, area_threshold: Number, connectivity: int=2,
```
dganguli: area_threshold is now a tuple as per previous PR

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+    gene_array = np.array(gene_list)
+
+    # reverses the linearization that was used to construct the IntensityTable from the ImageStack
+    decoded_image: np.ndarray = gene_array.reshape(intensities.attrs['image_shape'])
+
+    # label each pixel according to its component
+    label_image: np.ndarray = label(decoded_image, connectivity=connectivity)
+
+    # calculate the mean value grouped by (ch, round) across all pixels of a connected component
+    pixel_labels = label_image.reshape(-1)
```
dganguli: possible to do this w/o numpy and w/ xarray instead? perhaps add a TODO

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+    intensities['spot_id'] = (IntensityTable.Constants.FEATURES.value, pixel_labels)
+    mean_pixel_traces = intensities.groupby('spot_id').mean('features')
+
+    # "Zero" spot id is background in the label_image. Remove it.
+    mean_pixel_traces = mean_pixel_traces.loc[mean_pixel_traces['spot_id'] > 0]
```
dganguli: is there a simpler way to do this? perhaps up above

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+    props: List = regionprops(np.squeeze(label_image))
+
+    # calculate spots and drop ones that fail the area threshold
+    spots = []
+    passes_filter = np.ones_like(props, dtype=np.bool)  # default is passing (ones)
+    for i, spot_property in enumerate(props):
+        if spot_property.area >= area_threshold:
```
dganguli: area_threshold is a tuple.

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+            # we're back to 3d or fake-3d here
+            gene_index = decoded_image[spot_attrs['z'], spot_attrs['y'], spot_attrs['x']]
+            spot_attrs['gene_name'] = int_to_gene[gene_index]
+            spot_attrs['radius'] = spot_property.equivalent_diameter / 2
+            spots.append(spot_attrs)
+        else:
+            passes_filter[i] = 0  # did not pass radius filter
+
+    # filter intensities
+    mean_pixel_traces = mean_pixel_traces.loc[passes_filter]
```
dganguli: store passes_filter, but don't filter out mean_pixel_traces

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+                }
+            else:  # data is 2d
+                spot_attrs = {
+                    'z': 0,
+                    'y': int(spot_property.centroid[0]),
+                    'x': int(spot_property.centroid[1])
+                }
+
+            # we're back to 3d or fake-3d here
+            gene_index = decoded_image[spot_attrs['z'], spot_attrs['y'], spot_attrs['x']]
```
dganguli: i want everything un-filtered, with a boolean that tells me what passed the filters. i also want qualities (or distances)

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+    mean_pixel_traces = intensities.groupby('spot_id').mean('features')
+
+    # "Zero" spot id is background in the label_image. Remove it.
+    mean_pixel_traces = mean_pixel_traces.loc[mean_pixel_traces['spot_id'] > 0]
+    props: List = regionprops(np.squeeze(label_image))
+
+    # calculate spots and drop ones that fail the area threshold
+    spots = []
+    passes_filter = np.ones_like(props, dtype=np.bool)  # default is passing (ones)
+    for i, spot_property in enumerate(props):
```
dganguli: need to confirm that order of props matches order of x-array spots

starfish/pipeline/features/pixels/combine_adjacent_features.py
```diff
@@ -0,0 +1,117 @@
...
+from starfish.munge import dataframe_to_multiindex
+from starfish.constants import Indices
+from starfish.intensity_table import IntensityTable
+
+Number = Union[int, float]
+
+
+def combine_adjacent_features(
+        intensities: IntensityTable, area_threshold: Number, connectivity: int=2,
+) -> Tuple["IntensityTable", List, np.ndarray, np.ndarray]:
```
ttung: Remove "s.

ttung: Also, this still has too many return values.

ttung: Oh, I see, you changed it in #337

starfish/pipeline/features/pixels/combine_adjacent_features.py
ambrosejcarr: Added TODO.

starfish/pipeline/features/pixels/combine_adjacent_features.py
ambrosejcarr: Would need to make an xarray from the numpy array that skimage outputs, so I think we're screwed there unless we can make some of these skimage functions accept xarray objects.

starfish/pipeline/features/pixels/combine_adjacent_features.py
ambrosejcarr: The underlying changes to logic required to do this consistently are adequately significant to warrant doing it in a subsequent PR. #347

starfish/pipeline/features/pixels/combine_adjacent_features.py
ambrosejcarr: #347 for un-filtered, distances have been added back.

starfish/pipeline/features/pixels/combine_adjacent_features.py
ambrosejcarr: Ah, sorry. I should have changed it here. But yes, this is fixed.

PR #339, opened 2018-07-21
[8/8] DEP: remove unused EncodedSpots and DecodedSpots

ambrosejcarr:
These objects are no longer used. SpotAttributes can also be removed when the Allen notebook is upgraded (see #305).


starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,12 +1,8 @@
...
-
 from starfish.image import ImageStack
 from starfish.pipeline.algorithmbase import AlgorithmBase
-from starfish.pipeline.features.spot_attributes import SpotAttributes
-from starfish.pipeline.features.encoded_spots import EncodedSpots
 
 
 class SpotFinderAlgorithmBase(AlgorithmBase):
-    def find(self, hybridization_image: ImageStack) -> Tuple[SpotAttributes, EncodedSpots]:
+    def find(self, hybridization_image: ImageStack):
```
ttung: Is this supposed to return IntensityTable once LocalMaxPeakFinder is updated?

ambrosejcarr: Yep, that's right.

ttung: Pls TODO: this so we don't forget.

PR #340, opened 2018-07-22
[REFACTOR, Easy]: reduce constant verbosity

ambrosejcarr:
This is actually pretty simple, despite the big diff. It implements the following changes: 

- [Commit 1] Implements the below class and uses it to replace the constants in `Codebook` and `IntensityTable`. 

```python
class Features:
    """
    contains constants relating to the codebook and feature (spot/pixel) 
    representations of the image data
    """

    AXIS = 'features'
    TARGET = 'gene_name'
    CODEWORD = 'codeword'
    CODE_VALUE = 'v'
    SPOT_RADIUS = 'radius'
    SPOT_QUALITY = 'quality'
    Z = 'z'
    Y = 'y'
    X = 'x'
```

- [Commit 2]: changes TARGET from 'gene_name' to 'target'

- [Commit 3]: changes Indices.HYB to Indices.ROUND and updates the notebooks, file specifications, and codebooks on s3

- [Commit 4]: removes all vestiges of "hybridization round" or "gene_name" from the codebase in favor of "imaging_round" and "target"


ambrosejcarr:
No notebooks that ran previously are affected negatively by this change. 

It will need to wait on the MERFISH stack, since I coded this on top of it to avoid a refactor.

ambrosejcarr:
I wanted to raise the somewhat confusing state of `X, Y,` and `Z` indices. 

As it stands in this PR, we have access to the following constants: 
```python
class Coordinates(AugmentedEnum):
    Z = 'z'
    Y = 'y'
    X = 'x'

class Indices(AugmentedEnum):
    ROUND = 'r'
    CH = 'c'
    Z = 'z'
```
to which I have added: 
```python
class Features:
    """
    contains constants relating to the codebook and feature (spot/pixel) representations of the
    image data
    """

    AXIS = 'features'
    TARGET = 'target'
    CODEWORD = 'codeword'
    CODE_VALUE = 'v'
    SPOT_RADIUS = 'radius'
    SPOT_QUALITY = 'quality'
    Z = 'z'
    Y = 'y'
    X = 'x'
```
While we were discussing the spec, we decided that coordinates and indices were different things -- coordinates refer to relative locations in tissues, and are in units of distance, whereas indices deal with the tensors and in the case of `X` and `Y`, are in units of pixels, while `Z` is a categorical offset. 

It's my understanding that prior to this PR, we lacked a constant for `X` and `Y` pixels, and that coordinates probably shouldn't be used outside of `ImageStack`, interactions with `slicedimage`, and maybe `Registration`, since we don't have any other objects that deal with physical space. 

That's why my earlier PRs use a lot of `'x'`, `'y'`, and `'z'` string indices. I patch that here with the Features object, but I'm wondering if we can come up with something better. 

... we all love discussing formats, right? 🙃

ambrosejcarr:
Offline decision on constants: we want separate objects for indices and coordinates to support x/y/z, x and y should be added to indices. I will add that functionality to this PR before merging, and will remove all vestiges of `.value`.

ttung:
> Offline decision on constants: we want separate objects for indices and coordinates to support x/y/z, x and y should be added to indices. I will add that functionality to this PR before merging, and will remove all vestiges of .value.

Before you do that, let me do my stringification of enums. :)

ambrosejcarr:
Cool, will leave that for a later PR then.

PR #343, opened 2018-07-25
remove .value when possible.

ttung:
All the interactions between the enum and xarray appear to still require the `.value`.  I didn't dive too deeply into figuring out why, however.

Depends on #342


ttung:
Nope, I figured out a solution.  Will implement it now. :)

Issue #316, opened 2018-07-10
data typing


ttung:
Do we use uint16 everywhere?
Do we force algorithms to declare what the data types for each input / output will be?
Wild west?

ttung:
See discussion starting around https://spacetx.slack.com/archives/CAR91U354/p1531262002000193

kevinyamauchi:
Hey! I wanted to follow up on this issue. I am implementing some filters and was wondering if you decided how image data type is going to be handled. Should we add a function decorator, as suggested by Ambrose? Or will there be some sort of type check upon ingestion and then an assumed data type throughout the pipeline. 

Sorry if I missed something - I'm trying to follow along with the Slack, but I'm a bit behind.

kevinyamauchi:
Sounds great. I will try to take a look at skimage's typing before tomorow's sync. Thanks!

ttung:
My proposal:

The internal representation in starfish should default to float (32 or 64 is definitely debatable).  To be added:

1. A pipeline component to format input data into float, with an optional param to manage non-linear mappings.
2. A decorator that a specific algorithm implementation can utilize, which will inspect the input data and converts it to the requested type, and the results back to float.  There are optional params to manage non-linear mappings in each direction.

Some examples of (2):

This will convert the internal data type (float) to np.uint32, and the results will be returned to float.  All conversions are linear.
```
@datatype(np.uint32)
def run(..):
```

This will convert the internal data type (float) to np.uint32, and the results will be returned to float.  Converting from float to np.uint32 will be done with the scaling algorithm provided by XXX, and converting from np.uint32 to float will be done linearly.
```
@datatype(np.uint32, in_scaling=XXX)
def run(..):
```

This will convert the internal data type (float) to np.uint32, and the results will be returned to float.  Converting from float to np.uint32 will be done with the scaling algorithm provided by XXX, and converting from np.uint32 to float will be done with the scaling algorithm provided by YYY.
```
@datatype(np.uint32, in_scaling=XXX, out_scaling=YYY)
def run(..):
```

kevinyamauchi:
Thanks Tony for the proposal and clear examples! I think the proposal looks good. I have a couple of comments below. I look forward to discussing further.

[...]

PR #344, opened 2018-07-25
added mean high pass filter

kevinyamauchi:
I added a mean high pass filter following the same pattern as the Gaussian high pass filter. The image typing is handled in the same manner as the Gaussian high pass filter (uses img_as_uint to convert) and will likely need to be updated when issue #316 is resolved. 

I did not add a test, as only some of the filters have tests. However, if the current test style for the Gaussian High Pass filter (just checking that the total intensity is reduced) is sufficient, I can add that.

### DIFF
--- a/starfish/pipeline/filter/mean_high_pass.py
+++ b/starfish/pipeline/filter/mean_high_pass.py
@@ -17,15 +17,15 @@ class MeanHighPass(FilterAlgorithmBase):
     def __init__(
             self, size: Union[Number, Tuple[Number]], is_volume: bool=False, verbose: bool=False, **kwargs
     ) -> None:
-        """Mean high pass filter. 
+        """Mean high pass filter.
 
-        The mean high pass filter reduces low spatial frequency features by subtracting a 
-        mean filtered image from the original image. The mean filter smooths an image by replacing 
-        each pixel's value with an average of the pixel values of the surrounding neighborhood. 
+        The mean high pass filter reduces low spatial frequency features by subtracting a
+        mean filtered image from the original image. The mean filter smooths an image by replacing
+        each pixel's value with an average of the pixel values of the surrounding neighborhood.
 
         The mean filter is also known as a uniform or box filter.
 
-        This is a pass through for the scipy.ndimage.filters.uniform_filter: 
+        This is a pass through for the scipy.ndimage.filters.uniform_filter:
         https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.uniform_filter.html
 
         Parameters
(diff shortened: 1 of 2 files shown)


starfish/pipeline/filter/mean_high_pass.py
```diff
@@ -0,0 +1,94 @@
+import argparse
```
ambrosejcarr: Good idea to use pycharm optimize imports here, it will catch unused ones and optimize their order. 

Here's what it comes up with for this module: 

```python
import argparse
from functools import partial
from numbers import Number
from typing import Callable, Optional

import numpy as np
from scipy.ndimage.filters import uniform_filter
from skimage import img_as_uint

from starfish.errors import DataFormatWarning
from ._base import FilterAlgorithmBase
```

starfish/pipeline/filter/mean_high_pass.py
```diff
@@ -0,0 +1,94 @@
...
+from scipy.ndimage.filters import uniform_filter
+from ._base import FilterAlgorithmBase
+
+
+class MeanHighPass(FilterAlgorithmBase):
+
+    def __init__(
+            self, size: Number, is_volume: bool=False, verbose: bool=False, **kwargs
+    ) -> None:
+        """Mean high pass filter
```
ambrosejcarr: I would add a block with the following to the documentation: 

```
See Also
--------
scipy.ndimage.filters.uniform_filter: 
https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.uniform_filter.html
```

Also as an imaging noob, I don't see the comparison between `MeanHighPass` <-> `UniformFilter` anywhere. Adding a few words of explanation here might be nice, or referencing the other filter in the docs to make the connection for a user. 

It could also be nice to add a little bit of explanation of what this filter is supposed to do or why to apply it.

starfish/pipeline/filter/mean_high_pass.py
```diff
@@ -0,0 +1,94 @@
...
+        size : Number
+            width of the kernel
+        is_volume : bool
+            If True, 3d (z, y, x) volumes will be filtered, otherwise, filter 2d tiles independently.
+        verbose : bool
+            if True, report on filtering progress (default = False)
+
+        """
+
+        self.size = size
```
ambrosejcarr: It looks like `uniform_filter` supports a tuple of ints in the case that you want an anisotropic kernel. That might be nice to support here similar to how it works in `gaussian_high_pass`.

kevinyamauchi:
Thanks for the feedback, Ambrose! I've implemented the requested changes.

PR #348, opened 2018-07-27
ENH: require pytest>=3.6.3

ambrosejcarr:
- Travis installs pytest 3.3 by default, which breaks a dependency on pytest-xdist.


ambrosejcarr:
necro 👍

PR #350, opened 2018-07-27
Swap out underlying numpy array with a xarray

ttung:
This is the minimal set of changes to get all the tests passing.

It's worth investigating whether we can/should remove a lot of the intermediate functions like max_proj / get_slice / set_slice.


starfish/image/_stack.py
```diff
@@ -83,9 +84,22 @@ def __init__(self, image_partition):
             if ix == axis_data.order
         ]
         shape.extend(self._tile_shape)
+        dims = [
+            axis_name.value
+            for axis_name, axis_data in ImageStack.AXES_DATA.items()
+            for ix in range(ImageStack.N_AXES)
+            if ix == axis_data.order
+        ]
+        dims.extend([Coordinates.Y.value, Coordinates.X.value])
```
ambrosejcarr: Coordinates. 🤢 (No change necessary, will fix in PR-to-be following your stringification of enums.)

starfish/image/_stack.py
```diff
@@ -83,9 +84,22 @@ def __init__(self, image_partition):
             if ix == axis_data.order
         ]
         shape.extend(self._tile_shape)
+        dims = [
```
ambrosejcarr: I find nested for loops in list comprehensions a bit confusing. I can never remember which way to read them. What are your thoughts on this vs creating the empty list and appending to it? 

Second, do we still need AXES_DATA with this change to ImageStack? It buys us optionality, right? Can we delegate access to the information in AXES_DATA to the xarray.dims attribute?

ttung: > I find nested for loops in list comprehensions a bit confusing. I can never remember which way to read them. What are your thoughts on this vs creating the empty list and appending to it?

i don't think it's confusing, but i can switch it.  rule for reading: for-loops from top to bottom.

> Second, do we still need AXES_DATA with this change to ImageStack? It buys us optionality, right? Can we delegate access to the information in AXES_DATA to the xarray.dims attribute?

yes, we still need it, because we still build the array in that order.  we can make it hidden if you think it is beneficial. (but in separate PR)

ttung:
Also, I think I did not notice your comment because you didn't request changes.  Would generally prefer request-changes so it shows up in https://github.com/pulls/review-requested :)

ttung:
Actually it doesn't.  Ha ha, i'm the worst.  I'm thinking of phabricator.  Well, I still prefer request changes anyway. :)

PR #354, opened 2018-07-30
ENH: add osmFISH notebook that loads and displays data

ambrosejcarr:
- Add osmFISH notebook
- Make the notebook run on master

### DIFF
--- a/notebooks/osmFISH.py
+++ b/notebooks/osmFISH.py
@@ -19,7 +19,7 @@ import os
 # EPY: END markdown
 
 # EPY: START code
-s = Stack.from_experiment_json('https://dmf0bdeheu4zf.cloudfront.net/20180722/osmFISH/fov_001/experiment.json')
+s = Stack.from_experiment_json('https://dmf0bdeheu4zf.cloudfront.net/20180802/osmFISH/fov_001/experiment.json')
 # EPY: END code
 
 # EPY: START markdown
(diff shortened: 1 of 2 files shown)


notebooks/osmFISH.py
```diff
@@ -0,0 +1,31 @@
...
+# coding: utf-8
+#
+# EPY: stripped_notebook: {"metadata": {"kernelspec": {"display_name": "Python 3", "language": "python", "name": "python3"}, "language_info": {"codemirror_mode": {"name": "ipython", "version": 3}, "file_extension": ".py", "mimetype": "text/x-python", "name": "python", "nbconvert_exporter": "python", "pygments_lexer": "ipython3", "version": "3.6.5"}}, "nbformat": 4, "nbformat_minor": 2}
+
+# EPY: START markdown
+# ## Loading the data into Starfish
+# EPY: END markdown
+
+# EPY: START code
+from starfish.io import Stack
```
dganguli: Perhaps outside of scope of this PR, but I think we should stop using Stack in favor of ImageStack.

ambrosejcarr: Interesting. We'd probably have to use `ImageStack.from_url()` and point it at the _hybridization_.json instead of the experiment.json. @ttung Would that work? 

@dganguli If we did it this way, it wouldn't read in the `aux_images`. Is that desirable?

ttung: I think we should rename Stack → Experiment

notebooks/osmFISH.py
```diff
@@ -0,0 +1,31 @@
...
+# EPY: START code
+s = Stack.from_experiment_json('https://dmf0bdeheu4zf.cloudfront.net/20180722/osmFISH/fov_001/experiment.json')
+# EPY: END code
+
+# EPY: START markdown
+# The below plot displays the z-volume for channel 0 of fov_001
+# EPY: END markdown
+
+# EPY: START code
+s.image.show_stack({Indices.CH: 0});
```
ttung: remove ;

PR #357, opened 2018-08-01
Unify spot detection framework

ambrosejcarr:
Prior to this PR, starfish implements two exclusive strategies for spot detection: 
1. Local max peak finding, which measures each channel independently, building a "sparse intensity table" where only one channel/round is measured. It therefore exclusively functions for smFISH assays, but does a great job of addressing crowding where spots from different channels may overlap. 
2. gaussian blob detection, which identifies spot locations in a 'dots image' and then measures the intensities of the blob locations in each channel and round, building a "dense" intensity table. This is great for building codes, but can produce odd results when multiple dot locations may overlap. 

This PR unifies the framework for spot detection, exposing a single `detect_spots` method that takes, optionally, a dots image, or enables the building of one from a max projection. In these cases, spots are measured across (round, ch) tiles and enable the building of codes. When no comparison image is provided, spots are identified independently in each (round, ch) and build a "sparse intensity table" amenable to smFISH decoding. In addition, `detect_spots` takes a peak finding method which accepts an `ImageStack` and emits a `SpotAttributes` table. Thus, it can accept both blog-log and trackpy-locate spot finding methods. 

This PR adds shared testing data that are used to test the three spot finding approaches (no reference, blobs reference, max projection reference) and both spot finding methods, guaranteeing that they act identically on noiseless test data. 

Finally, it moves some code from the `GaussianSpotDetector` that is now general across methods into `detect.py`. I have added notes throughout the code that explain the changes. 

The logical way to review this code is to start with `detect.py`, specifically with `detect_spots`. It may then make sense to look at the testing cases in `test_spot_detection.py` (renamed from `test_gaussian_spot_detector.py`, since it now generalizes). The changes to `GaussianSpotDetector` and `LocalMaxPeakFinder` are exclusively to support the model required by `detect.py`. 

Two things this PR does NOT address: 
1. Parameterization of mean/max/sum of a spot's intensity. Right now it is hard coded to take the max. 
2. Radius is treated differently in trackpy locate and blob-log. On noiseless data, they measure the same maximum spot intensities, however this may not hold on real image based transcriptomics data, and would not hold for sum or mean intensities.


starfish/pipeline/features/spot_attributes.py
```diff
@@ -11,7 +11,7 @@ class SpotAttributes(ValidatedTable):
     required_fields = {
         'x',  # spot x-coordinate
         'y',  # spot y-coordinate
-        'r',  # spot radius
```
ambrosejcarr: conflicts with `Features.round`

starfish/pipeline/features/spots/detector/__init__.py
```diff
@@ -23,6 +24,13 @@ def add_to_parser(cls, subparsers):
         spot_finder_group = subparsers.add_parser("detect_spots")
         spot_finder_group.add_argument("-i", "--input", type=FsExistsType(), required=True)
         spot_finder_group.add_argument("-o", "--output", required=True)
+        spot_finder_group.add_argument(
```
ambrosejcarr: Acceptance of a dots image is required for all spot finders, so I moved it from the Gaussian spot detector to the parent class.

starfish/pipeline/features/spots/detector/__init__.py
```diff
@@ -43,6 +51,11 @@ def _cli(cls, args, print_help=False):
 
         print('Detecting Spots ...')
         image_stack = ImageStack.from_path_or_url(args.input)
+        if args.blobs_stack is not None:
```
ambrosejcarr: Stacks are loaded in the `cli` code to allow the constructors of the algorithms to require an ImageStack (no strings pointing to filenames).

starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,9 +1,30 @@
...
+import xarray as xr
+import numpy as np
+
 from starfish.image import ImageStack
 from starfish.pipeline.algorithmbase import AlgorithmBase
+from starfish.pipeline.features.spot_attributes import SpotAttributes
+from starfish.intensity_table import IntensityTable
 
 
 class SpotFinderAlgorithmBase(AlgorithmBase):
```
ambrosejcarr: Adds typing and two methods that are general to all implementing algorithms: `image_to_spots`, which finds spots and emits a spot_attributes table, and `_get_measurement_function` which converts `max`, `mean` etc to the corresponding numpy function.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
+from functools import partial
```
ambrosejcarr: This module the focal point of the PR.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+
+from starfish.constants import Indices, Features
+from starfish.image import ImageStack
+from starfish.intensity_table import IntensityTable
+from starfish.munge import dataframe_to_multiindex
+from starfish.pipeline.features.spot_attributes import SpotAttributes
+from starfish.typing import Number
+
+
+def measure_spot_intensity(
```
ambrosejcarr: This has been moved from `GaussianSpotDetector` but aspects (lines 48-52) were moved from the spot detector into this specific method because (1) the bounding box (zmin/max, ymin/max, xmin/max) are not used outside this function, and (2) they're needed for the trackpy-locate based method, so this is the best place for that code.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+        spots[f'{v}_min'] = np.clip(spots[v] - inclusive_radius, 0, None)
+        spots[f'{v}_max'] = np.clip(spots[v] + inclusive_radius, None, max_size)
+
+    return spots[['z_min', 'z_max', 'y_min', 'y_max', 'x_min', 'x_max']].astype(int).apply(
+        fn,
+        axis=1
+    )
+
+
+def measure_spot_intensities(
```
ambrosejcarr: This is moved from GaussianSpotDetector.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+    for c, r in indices:
+        image, _ = data_image.get_slice({Indices.CH: c, Indices.ROUND: r})
+        blob_intensities: pd.Series = measure_spot_intensity(
+            image, spot_attributes, measurement_function)
+        intensity_table[:, c, r] = blob_intensities
+
+    return intensity_table
+
+
+def concatenate_spot_attributes_to_intensities(
```
ambrosejcarr: Previously there was no logic to stick together multiple SpotAttributes tables into a single IntensityTable. Now that we implement the LocalMaxPeakFinder approach, we needed one. This serves that role.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+
+    i = 0
+    for attrs, inds in spot_attributes:
+        for _, row in attrs.data.iterrows():
+            intensity_table[i, inds[Indices.CH], inds[Indices.ROUND]] = row['intensity']
+            i += 1
+
+    return intensity_table
+
+
```
ambrosejcarr: This method is the centerpiece of the PR, and provides the general framework and requirements for spot detection. Both `LocalMaxPeakFinder` and `GaussianSpotDetector` call this method from their `find` methods.

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -1,25 +1,29 @@
...
 from ._base import SpotFinderAlgorithmBase
 
 
 class GaussianSpotDetector(SpotFinderAlgorithmBase):
 
     def __init__(
-            self, min_sigma, max_sigma, num_sigma, threshold,
-            blobs_stack, overlap=0.5, measurement_type='max', is_volume: bool=True, **kwargs
-    ) -> None:
+            self, min_sigma: Number, max_sigma: Number, num_sigma: int, threshold: Number,
```
ambrosejcarr: Adding types, but also moving blobs stack to the `find` method. Since the spot finders already accept the `image_stack` this seemed more consistent. It also makes the parameterization of `detect_spots` simpler. This is replicated for the LocalMaxPeakFinder.

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -57,66 +61,34 @@ def __init__(
         self.threshold = threshold
         self.overlap = overlap
         self.is_volume = is_volume
-        if isinstance(blobs_stack, ImageStack):
```
ambrosejcarr: General code moved to `detect.py`

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -125,36 +97,38 @@ def _find_spot_locations(self) -> pd.DataFrame:
         fitted_blobs[Features.SPOT_RADIUS] = converted_radius
 
         # convert the array to int so it can be used to index
-        fitted_blobs = fitted_blobs.astype(int)
-
-        for v, max_size in zip(['z', 'y', 'x'], self.blobs_image.shape):
```
ambrosejcarr: moved to `measure_intensity`, since LocalMaxPeakFinder also needs this.

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -57,66 +61,34 @@ def __init__(
...
-        return blobs.apply(
-            fn,
-            axis=1
-        )
+        self.measurement_function = self._get_measurement_function(measurement_type)
 
-    def _measure_spot_intensities(
-            self, stack: ImageStack, spot_attributes: pd.DataFrame
-    ) -> IntensityTable:
+    def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> SpotAttributes:
```
ambrosejcarr: Rename this method, now required by all implementing algorithms. Calls `detect_spots`

starfish/pipeline/features/spots/detector/local_max_peak_finder.py
```diff
@@ -1,20 +1,24 @@
-from typing import List
+from typing import Union, Tuple
```
ambrosejcarr: Most of the code changes in this module are to unify this module with the general framework.

starfish/test/dataset_fixtures.py
```diff
@@ -208,14 +208,54 @@ def synthetic_single_spot_3d():
...
 def synthetic_two_spot_3d():
-    from scipy.ndimage.filters import gaussian_filter
     data = np.zeros((10, 100, 100), dtype=np.uint16)
     data[4, 10, 90] = 1000
     data[6, 90, 10] = 1000
     data = gaussian_filter(data, sigma=2)
     return data
 
 
+def synthetic_two_spot_3d_2round_2ch() -> ImageStack:
```
ambrosejcarr: This provides a very general multi-spot ImageStack to test all the spot-finding methods.

starfish/test/pipeline/test_gaussian_spot_detector.py
```diff
@@ -1,166 +0,0 @@
-import numpy as np
```
ambrosejcarr: The top of this got renamed.

starfish/test/pipeline/test_gaussian_spot_detector.py
```diff
@@ -1,166 +0,0 @@
...
-    """
-    verify that 3d blob log works, even when the third dimension is too small to support the
-    observed standard deviation
-    """
-    data = synthetic_single_spot_2d.reshape(1, *synthetic_single_spot_2d.shape)
-    result = blob_log(data, min_sigma=2, max_sigma=4, num_sigma=10, threshold=0)
-    assert np.array_equal(result, np.array([[0, 10, 90, 2]]))
-
-
-def single_spot_detector(stack):
```
ambrosejcarr: From here down is deleted and replaced by tests that are general across the two spot detectors.

starfish/test/pipeline/test_spot_detection.py
```diff
@@ -0,0 +1,148 @@
...
+def test_blob_log_3d_flat(synthetic_single_spot_2d):
+    """
+    Tests that skimage.feature.blob_log is acting as we expect on pseudo-3d data where the
+    z-axis has size 1
+    """
+    data = synthetic_single_spot_2d.reshape(1, *synthetic_single_spot_2d.shape)
+    result = blob_log(data, min_sigma=2, max_sigma=4, num_sigma=10, threshold=0)
+    assert np.array_equal(result, np.array([[0, 10, 90, 2]]))
+
+
```
ambrosejcarr: From here down is the unified testing framework for both spot detectors.

starfish/pipeline/features/spot_attributes.py
dganguli: this may break notebooks that try to refer to a spot's radius

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -38,16 +42,16 @@ def __init__(
...
+            If two spots have more than this fraction of overlap, the spots are combined
+            (default = 0.5)
         measurement_type : str ['max', 'mean']
             name of the function used to calculate the intensity for each identified spot area
 
         Notes
         -----
-        This spot detector is very sensitive to the threshold that is selected, and the threshold is defined as an
-        absolute value -- therefore it must be adjusted depending on the datatype of the passed image.
+        This spot detector is very sensitive to the threshold that is selected, and the threshold
```
dganguli: I think Cellprofiler has a good solution to this problem. Internally, it takes the analogue of blobs_stack, and scales it to lie in [0, 1]. As such, your threshold is now a number between [0, 1] as well -- which makes understanding this parameter across varying dynamic ranges more interpretable. I propose we follow this precedent. We either check whether or not blobs_stack is in [0,1] (and threshold as well) and throw an exception if not. OR we simply do the transformation ourselves on blobs_stack and throw an exception if threshold is not in [0,1]. What do you think?

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -125,36 +97,38 @@ def _find_spot_locations(self) -> pd.DataFrame:
...
             3d tensor containing the intensity of spots across channels and imaging rounds
 
         """
-        spot_attributes = self._find_spot_locations()
-        intensity_table = self._measure_spot_intensities(image_stack, spot_attributes)
+
+        intensity_table = detect_spots(
+            data_stack=data_stack,
+            spot_finding_method=self.image_to_spots,
+            reference_image=blobs_image,
```
dganguli: The find method in this detector is hard-coded to accept a 'blobs' image. This does not leverage the flexiblity of the detect_spots method to accept an argument where reference_image=None and reference_from_max_projection=True. This is a very common use case, and I think we should allow for this easily in the GaussianSpotDetector, especially if we go with the convention that the blobs_image should be normalized to [0,1] a-priori.

starfish/pipeline/features/spots/detector/gaussian.py
```diff
@@ -1,25 +1,29 @@
...
 class GaussianSpotDetector(SpotFinderAlgorithmBase):
 
     def __init__(
-            self, min_sigma, max_sigma, num_sigma, threshold,
-            blobs_stack, overlap=0.5, measurement_type='max', is_volume: bool=True, **kwargs
-    ) -> None:
+            self, min_sigma: Number, max_sigma: Number, num_sigma: int, threshold: Number,
+            overlap=0.5, measurement_type='max', is_volume: bool=True, **kwargs) \
+            -> None:
         """Multi-dimensional gaussian spot detector
```
dganguli: Add -- "... this is a wrapper for skimage.blahblahblah blah" or something like that, analagous to peak local max finder.

starfish/pipeline/features/spots/detector/__init__.py
dganguli: i think this breaks the gaussian spot detector of blobs_stack is none. according to the typing, the find method expects an xarray or ndarray

starfish/pipeline/features/spots/detector/detect.py
dganguli: this is great! love it!

starfish/pipeline/features/spots/detector/local_max_peak_finder.py
```diff
@@ -112,26 +119,29 @@ def find_attributes(self, image: np.ndarray) -> SpotAttributes:
...
+            If provided, spots will be found in the blobs image, and intensities will be measured
+            across hybs and channels. Otherwise, spots are measured independently for each channel
+            and round.
 
-        # TODO ambrosejcarr: this is where development stopped; spot_attributes is correct, but translating
-        # spot_attributes into an encoder_table is tricky without first implementing the new codebook. Do that first.
-        # create an encoded table
-        # encoded_spots = self.encode(spot_attributes.data)
+        """
+        intensity_table = detect_spots(
```
dganguli: i think this should also allow the user the ability to compute the blobs image as the max projection, thus fully leveraging detect_spots flexibility.

dganguli:
One last thought -- ConnectedComponentsSpotDetector can also be a 'thing' in this module that extends SpotFinderAlgorithmBase. It's image_to_spots function binarizes the image, runs a connected components algorithm, filters out identified regions that don't meet an area or intensity threshold, and otherwise operates the same as the other two spot detectors. This is only slightly different than how it's used in pixel based readout in features.pixels. 

Outside the scope of this PR, but it seems to me worth revisiting how to frame the PixelSpotDetector now that you are well versed in the ways of spot detection :)

starfish/pipeline/features/spot_attributes.py
ambrosejcarr: I'll make sure there aren't any missed references. 👍

starfish/pipeline/features/spots/detector/gaussian.py
ambrosejcarr: I agree. I think this change would ideally follow after Tony's solution for dtypes, since it sounds like the goal here would be consistent. I expect he'll probably write a few basic functions to check that the images are in the right dtype, and that we could leverage them here.

starfish/pipeline/features/spots/detector/gaussian.py
ambrosejcarr: Good catch, the typing should be Optional[...] to allow for the None case. 

You'll notice later in the PR that the tests do verify that the functionality you desire there works, so we just need to update the type 👍

starfish/pipeline/features/spots/detector/local_max_peak_finder.py
dganguli: still missing the pass thru of ```reference_image_from_max_projection``` here from constructor to detect_spots

ambrosejcarr:
I've verified that the ISS notebook yields identical results. Made an issue to track the idea of a connected components spot detector.

starfish/pipeline/features/spots/detector/__init__.py
```diff
@@ -23,8 +24,23 @@ def add_to_parser(cls, subparsers):
...
         spot_finder_group.add_argument("-o", "--output", required=True)
+        spot_finder_group.add_argument(
+            "--blobs-stack", default=None, required=False,
+            help=(
+                'ImageStack that contains the blobs. Will be max-projected across imaging round '
+                'and channel to produce the blobs_image'
+            )
+        )
+        spot_finder_group.add_argument(
+            "--reference-image-from-max-projection", default=False, action='store_true',
```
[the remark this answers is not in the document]
ttung: In general, we want to decompose the pipeline stages as much as reasonable.  The price we pay for I/O can be alleviated by improvements in our runner.

starfish/pipeline/features/spots/detector/__init__.py
```diff
@@ -43,6 +59,11 @@ def _cli(cls, args, print_help=False):
...
         print('Detecting Spots ...')
         image_stack = ImageStack.from_path_or_url(args.input)
+        if args.blobs_stack is not None:
+            blobs_stack = ImageStack.from_path_or_url(args.blobs_stack)  # type: ignore
+            blobs_image = blobs_stack.max_proj(Indices.ROUND, Indices.CH)
+        else:
+            blobs_image = None
         instance = args.spot_finder_algorithm_class(**vars(args))
-        intensities = instance.find(image_stack)
+        intensities = instance.find(image_stack, blobs_image=blobs_image)
```
ttung: philosophical discussion:

`reference_image_from_max_projection` is passed into the constructor (if I'm reading the code correctly) but `blobs_image` is passed into the find method.  I think this is correct, because `reference_image_from_max_projection` is applied for an entire experiment, but `blobs_image` is different per FOV.

however, i want to make sure everyone is still on board with this distinction.

starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,9 +1,30 @@
+from typing import Union
+
+import xarray as xr
```
ttung: sort!

starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,9 +1,30 @@
...
+import numpy as np
+
 from starfish.image import ImageStack
 from starfish.pipeline.algorithmbase import AlgorithmBase
+from starfish.pipeline.features.spot_attributes import SpotAttributes
+from starfish.intensity_table import IntensityTable
```
ttung: sort!

starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,9 +1,30 @@
...
 
 
 class SpotFinderAlgorithmBase(AlgorithmBase):
-    # TODO ambrosejcarr: should return IntensityTable when refactor is complete
-    def find(self, hybridization_image: ImageStack):
-        """Find spots."""
+    def find(self, hybridization_image: ImageStack, blobs_image: Union[np.ndarray, xr.DataArray]) \
```
ttung: formatting -- would prefer:

```
def find(
    self,
    hybridization_image: ImageStack,
    blobs_image: Union[np.ndarray, xr.DataArray]),
) -> IntensityTable:
```

ttung: Also, I think it would be prudent to consider whether we should accept one of the two (np.ndarray vs xr.DataArray), rather than either.

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+    return spots[['z_min', 'z_max', 'y_min', 'y_max', 'x_min', 'x_max']].astype(int).apply(
+        fn,
+        axis=1
+    )
+
+
+def measure_spot_intensities(
+        data_image: ImageStack,
+        spot_attributes: pd.DataFrame,
+        measurement_function: Callable[[Sequence], Number], ) \
```
ttung: move ) to the next line.  you don't need the \

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+        -> IntensityTable:
+    """given spots found from a reference image, find those spots across a data_image
+
+    Parameters
+    ----------
+    data_image : ImageStack
+        ImageStack containing multiple volumes for which spots' intensities must be calculated
+    spot_attributes : pd.Dataframe
+        Locations and radii of spots
+    measurement_function : Callable[[Sequence], Number]) \
```
ttung: backslash?

starfish/pipeline/features/spots/detector/_base.py
```diff
@@ -1,9 +1,30 @@
...
+            -> IntensityTable:
+        """Finds spots in an ImageStack"""
         raise NotImplementedError()
+
+    def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> SpotAttributes:
+        """Finds spots in a 3d volume"""
+        raise NotImplementedError()
+
+    @staticmethod
+    def _get_measurement_function(measurement_type: str):
```
ttung: why does this not return Callable[[Sequence], Number] like you're doing in `measure_spot_intensities`?

starfish/pipeline/features/spots/detector/detect.py
```diff
@@ -0,0 +1,222 @@
...
+        -> pd.Series:
+    """measure the intensity of each spot in spots in the corresponding image
+
+    Parameters
+    ----------
+    image : Union[np.ndarray, xr.DataArray],
+        3-d volume in which to measure intensities
+    spots : pd.DataFrame
+        SpotAttributes table containing coordinates and radii of spots
+    measurement_function : Callable[[Sequence], Number]) \
```
ttung: no back slash.

also, does it take a sequence?  i thought it takes a numpy array.

starfish/pipeline/features/spots/detector/__init__.py
```diff
@@ -23,8 +24,23 @@ def add_to_parser(cls, subparsers):
         spot_finder_group = subparsers.add_parser("detect_spots")
         spot_finder_group.add_argument("-i", "--input", type=FsExistsType(), required=True)
         spot_finder_group.add_argument("-o", "--output", required=True)
+        spot_finder_group.add_argument(
+            "--blobs-stack", default=None, required=False,
+            help=(
```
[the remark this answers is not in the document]
ambrosejcarr: @dganguli what do you think about changing `blobs-stack` (and in code: `blobs-image`) to `reference-stack` and `reference-image`?

starfish/pipeline/features/spots/detector/__init__.py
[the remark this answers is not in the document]
ambrosejcarr: #373

starfish/pipeline/features/spots/detector/__init__.py
ambrosejcarr: I messed this up in `__init__`. I think reference_image_from_max_projection _probably is_ applied to the whole experiment, but I think it should logically go in the find() method since it shares functionality with `blobs_image`. Technically there is no limitation requiring one to have a `blobs_image` for each FOV, although it would be an odd experimental design that would produce a dataset with only some of the FOVs having this data. 

I've added `reference_image_from_max_projection` to the CLI.

starfish/pipeline/features/spots/detector/_base.py
ambrosejcarr: Yea I think when you land your PR we'll make it all xr.DataArray, but right now I have to hedge. 😛

starfish/pipeline/features/spots/detector/detect.py
ambrosejcarr: Since we're designing this around numpy ufuncs, they are actually more general -- you can feed them lists, pandas series, tuples, or pandas dataframes, and they'll all generate the correct output. We could choose to limit flexibility here as a design decision.

PR #359, opened 2018-08-02
ENH: change filter.filter to filter.run, Update bandpass

ambrosejcarr:
- change `<filter_name>.filter` -> filter.run to improve ease of use of the package (all algorithms will be `<name>.run()`)
- update `bandpass.py` to adhere to various code standards
  - line length 100
  - typed return values
  - typed parameters
  - runtime parameters in run, algorithm parameters in constructor (`verbose` from `__init__` -> `run`)
  - remove default arguments, except for `truncate` which I don't understand
  - assign @dganguli to doc that method (not doc'd in trackpy)
  - add 3d volume support

Notes:
- a bunch of notebooks got brought along for the ride due to the refactor. 
- Further PRs will address the other algorithms in filter.

### DIFF
--- a/starfish/pipeline/filter/bandpass.py
+++ b/starfish/pipeline/filter/bandpass.py
@@ -1,16 +1,19 @@
 from functools import partial
 from typing import Optional
 
-from trackpy import bandpass
 import numpy as np
+from trackpy import bandpass
 
 from starfish.image import ImageStack
+from starfish.types import Number
 from ._base import FilterAlgorithmBase
 
 
 class Bandpass(FilterAlgorithmBase):
 
-    def __init__(self, lshort, llong, threshold, truncate, verbose=False, **kwargs):
+    def __init__(
+            self, lshort: Number, llong: int, threshold: Number, truncate: Number=4,
+            is_volume: bool=False, **kwargs) -> None:
         """
 
         Parameters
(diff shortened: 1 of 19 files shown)


starfish/pipeline/filter/bandpass.py
```diff
@@ -54,21 +62,23 @@ def bandpass(image, lshort, llong, threshold, truncate) -> np.ndarray:
...
         """
-        bandpassed = bandpass(
+        bandpassed: np.ndarray = bandpass(
             image, lshort=lshort, llong=llong, threshold=threshold,
             truncate=truncate
         )
         return bandpassed
 
-    def filter(self, stack: ImageStack, in_place: bool=True) -> Optional[ImageStack]:
+    def run(
```
ttung:
```
    def run(
        self, stack: ImageStack, in_place: bool=True, verbose: bool=False
    ) -> Optional[ImageStack]:
```

starfish/pipeline/filter/bandpass.py
```diff
@@ -85,9 +97,12 @@ def filter(self, stack: ImageStack, in_place: bool=True) -> Optional[ImageStack]
 
         """
         bandpass_ = partial(
```
ttung: if you're going in this direction, i would rather see:

```
        bandpass_ = partial(
            self.bandpass,
            lshort=self.lshort, llong=self.llong, threshold=self.threshold, truncate=self.truncate
        )
```

first argument to `partial` is more different than the rest of the arguments than truncate is to the first arguments.

PR #368, opened 2018-08-03
RFC: update target assignment API

ambrosejcarr:
In attempting to modernize the target assignment API a few questions came up: 

1. Currently, target assignment emits a 2-column `Dataframe` that maps spot ids to cells. We no longer have an explicit representation of spot ids in the IntensityTable, and I wonder if the IntensityTable should just gain a column of "cell id" 

2. As far as I can tell, segmentation in starfish is purely 2d at the moment. This means that our assignment method is also 2d, however we assume but don't test this and I have a feeling that as a result, this method may not work for approaches that don't follow the ISS pipeline. 

Proposal: 
- If we store cell information in the `IntensityTable`, we can simply assign spots to cells using x/y coordinates (ignore z) until such a time as we support 3d segmentation. 
- Storing as an IntensityTable also lets us punt on filtering (since future-starfish is storing a `passes_filter` mask), allowing us to keep "cell" associations for both properly coded and noise spots, which could be nice for downstream QC.

### DIFF
--- a/starfish/pipeline/target_assignment/point_in_poly.py
+++ b/starfish/pipeline/target_assignment/point_in_poly.py
@@ -1,41 +1,73 @@
 import numpy as np
-from skimage.measure import points_in_poly
 import pandas as pd
+import regional
+from skimage.measure import points_in_poly
+from tqdm import tqdm
 
 from starfish.constants import Features
+from starfish.intensity_table import IntensityTable
 from ._base import TargetAssignmentAlgorithm
 
 
 class PointInPoly(TargetAssignmentAlgorithm):
-    def __init__(self, verbose=False, **kwargs):
-        self.verbose = verbose
+
+    def __init__(self, **kwargs):
+        """
+        PointInPoly accepts no parameters, but all pipeline components must accept arbitrary kwargs
+        """
 
     @classmethod
-    def add_arguments(cls, parser):
+    def add_arguments(cls, parser) -> None:
         pass
 
     @staticmethod
-    def _assign(cells_region, spots, use_hull=True, verbose=False):
-        res = pd.DataFrame({'spot_id': range(0, spots.shape[0])})
-        res['cell_id'] = None
+    def _assign(
+            cells_region: regional.many, spots: pd.DataFrame, use_hull=True, verbose=False) \
+            -> pd.DataFrame:
 
-        for cell_id in range(cells_region.count):
+        results = pd.DataFrame({'spot_id': range(0, spots.shape[0])})
+        results['cell_id'] = None
+
+        if verbose:
+            cell_iterator = tqdm(range(cells_region.count))
+        else:
+            cell_iterator = range(cells_region.count)
+
+        for cell_id in cell_iterator:
             if use_hull:
-                verts = cells_region[cell_id].hull
+                vertices = cells_region[cell_id].hull
             else:
-                verts = cells_region[cell_id].coordinates
-            verts = np.array(verts)
-            in_poly = points_in_poly(spots, verts)
-            res.loc[res.spot_id[in_poly], 'cell_id'] = cell_id
-            if verbose:
-                cnt = np.sum(in_poly)
-                print(cell_id, cnt)
+                vertices = cells_region[cell_id].coordinates
+            vertices = np.array(vertices)
+            in_poly = points_in_poly(spots, vertices)
+            results.loc[results.spot_id[in_poly], 'cell_id'] = cell_id
 
-        return res
+        return results
 
-    def assign_targets(self, intensity_table, regions):
+    def run(
+            self, intensity_table: IntensityTable, regions: regional.many, verbose: bool=False) \
+            -> pd.DataFrame:
+        """Assign spots with target assignments to cells
 
+        Parameters
+        ----------
+        intensity_table : IntensityTable
+        regions : regional.many
+            # TODO dganguli can you add an explanation? I'll fix during PR.
+        verbose : bool
+            (default False) report on the progress of gene assignment
+
+
+        Returns
+        -------
+        pd.DataFrame :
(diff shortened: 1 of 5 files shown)


starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
 from starfish.constants import Features
+from starfish.intensity_table import IntensityTable
 from ._base import TargetAssignmentAlgorithm
 
 
 class PointInPoly(TargetAssignmentAlgorithm):
-    def __init__(self, verbose=False, **kwargs):
-        self.verbose = verbose
+
+    def __init__(self, **kwargs):
```
ttung: -> None

starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
-    def add_arguments(cls, parser):
+    def add_arguments(cls, parser) -> None:
         pass
 
     @staticmethod
-    def _assign(cells_region, spots, use_hull=True, verbose=False):
-        res = pd.DataFrame({'spot_id': range(0, spots.shape[0])})
-        res['cell_id'] = None
+    def _assign(
+            cells_region: regional.many, spots: pd.DataFrame, use_hull=True, verbose=False) \
```
ttung: no backslash

ttung: use_hull and verbose should be typed bool

starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
-                print(cell_id, cnt)
+                vertices = cells_region[cell_id].coordinates
+            vertices = np.array(vertices)
+            in_poly = points_in_poly(spots, vertices)
+            results.loc[results.spot_id[in_poly], 'cell_id'] = cell_id
+
+        return results
+
+    def run(
+            self, intensity_table: IntensityTable, regions: regional.many, verbose: bool=False) \
```
ttung: no backslash.

starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
+            -> pd.DataFrame:
+        """Assign spots with target assignments to cells
+
+        Parameters
+        ----------
+        intensity_table : IntensityTable
+        regions : regional.many
+            # TODO dganguli can you add an explanation? I'll fix during PR.
+        verbose : bool
+            (default False) report on the progress of gene assignment
```
ttung: this is a different way of expressing the default.

starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
+import regional
+from skimage.measure import points_in_poly
+from tqdm import tqdm
 
 from starfish.constants import Features
+from starfish.intensity_table import IntensityTable
 from ._base import TargetAssignmentAlgorithm
 
 
 class PointInPoly(TargetAssignmentAlgorithm):
```
dganguli: Rename to PointInPoly2D to call out that this only works in 2D? Or at least, I think it only works in 2D.

dganguli:
I think the API for target_assignement looking like IntensityTable, regional -> IntensityTable with an additional cell_id column makes sense, and is clean. we simply should add something to ensure that cell_id is indeed added to the intensity table, a la ValidatedSpotAttributes, or a post-run hook that raises an exception if cell_id is not found in the table. 

The watershed algorithm wrapped by segmentation.watershed trivially works in 3D (see line 119  (if image and markers are ZxYxX images, you will output a ZxYxX image of cell boundaries). I think removing the max_proj's in lines 35 and 43 over Indices.Z (and handling the requisite 'fake-3d' issues for fundamentally 2D issues will do the trick). 

however, the current target implementation code will break -- as that only works in 2D (see my comment in the code).

dganguli:
yeah @berl definitely want to record spots that are not assigned to a cell. i bet @ambrosejcarr will want to do that with a None or np.NaN or something like that.

ambrosejcarr:
Yep, @dganguli nailed it. Probably `np.nan` 👍

berl:
just for my own edification...  do y'all have smart tools to handle serializing np.nan values for json and csv?

ambrosejcarr:
@berl The limitation in python is that `np.nan` is a float, so that is how our cell assignments must be dtyped for this to work. 

np.nan is serialized to `nan` or `NaN` by most packages (numpy, pandas).

ambrosejcarr:
Target Assignment + IntensityTable: #374 
Watershed in 3D: #375

starfish/pipeline/target_assignment/point_in_poly.py
```diff
@@ -1,41 +1,73 @@
...
-            verts = np.array(verts)
-            in_poly = points_in_poly(spots, verts)
-            res.loc[res.spot_id[in_poly], 'cell_id'] = cell_id
-            if verbose:
-                cnt = np.sum(in_poly)
-                print(cell_id, cnt)
+                vertices = cells_region[cell_id].coordinates
+            vertices = np.array(vertices)
+            in_poly = points_in_poly(spots, vertices)
+            results.loc[results.spot_id[in_poly], 'cell_id'] = cell_id
```
berl: If there are overlapping cells (which can definitely happen if we use a convex hull), this will assign a spot to the last cell it's in.  The current structure (spots gaining a unique `cell_id` column) will always have this problem, but an `assign_targets` dataframe could get around this if it's basically a list of spots that are associated with each cell.

ambrosejcarr: Nice catch. Referencing this for #374 .

PR #370, opened 2018-08-06
[Easy] Remove starfish.viz

dganguli:
This functionality has been re-placed by ImageStack.show_stack. RIP.


starfish/viz.py
```diff
@@ -1,19 +0,0 @@
-import numpy as np
-
-from .stats import im_describe, stack_describe
```
ambrosejcarr: Any chance we can also nuke these as a result?

PR #378, opened 2018-08-08
Added IntensityTable.to_dataframe and a corresponding unit test

dganguli:
IntensityTable stores 'features' as a pandas multi-index in the coordinates portion of the underlying xarray. Users will often want to interact with these features in a simple pandas dataframe. IntensityTable.to_dataframe achieves this result by translating the features multiindex into a simple dataframe.


starfish/test/pipeline/test_intensity_table.py
```diff
@@ -120,3 +126,14 @@ def test_combine_adjacent_features():
         combined_intensities[Features.AXIS][Features.TARGET],
         ['gene_2', 'gene_1']
     )
+
+
+def test_intensity_table_to_dataframe():
+    table, _ = make_empty_intensity_table()
+    table_df = table.to_dataframe()
+    expected_columns = set([Features.X, Features.Y, Features.Z, Features.SPOT_RADIUS])
+    assert(set(table_df[Features.X]) == set(table.coords[Features.X].values))
```
ambrosejcarr: To check that x, y, z, and radius are retaining their linkages, you could try iterating through the multiindex and the dataframe and checking that each 4-tuple are identical.

starfish/test/pipeline/test_intensity_table.py
```diff
@@ -120,3 +126,14 @@ def test_combine_adjacent_features():
...
+    assert(set(table_df[Features.Y]) == set(table.coords[Features.Y].values))
```
ambrosejcarr: Don't need the sets.

starfish/test/pipeline/test_intensity_table.py
dganguli: i think that's overkill for testing that row order is preserved, in that it also checks whether column order is preserved (which we care less about). instead, i have removed the calls to set to ensure row order is indeed preserved by this function.

PR #382, opened 2018-08-09
Update README based on starfish/test/full_pipelines/cli/test_iss.py

ttung:


ttung:
btw, I know I approved it, but I prefer `starfish register` over `starfish registration`.  rationale: would prefer verbs.

PR #386, opened 2018-08-10
remove sudo requirement

ttung:
We don't actually need it as far as I can tell.


ttung:
This seems to pass now. :)

PR #388, opened 2018-08-10
Update show_stack to work in both inline and notebook modes

dganguli:
This PR generalizes #346, which was merged previously.

#346 introduced a more performant version of show_stack, which also allowed one to hover over spots and view their intensities. These features are awesome. However, #346 required the user to run notebooks with %matplotlib notebook. It's possible I'm going cray, but I've noticed that running notebooks in this manner, with #346, lead to several difficulties in analyzing data with multiple calls to show_stack across cells:

1. Plots would often not render upon a notebook re-run
2. Scrolling behavior in the notebook changed where it was hard to scroll past images of show_stack
3. Changing a stack in one cell, changed the visualization of it in an earlier cell, if I could get it to render.

Instead of addressing these issues in this PR, I've submitted a patch. show_stack now infers your matplotlib backend (%inline or %notebook) then runs the corresponding version of show_stack (pre: #346 for %inline and #346 for %notebook). Along the way, I've re-factored show_stack to call a few functions instead of be one monolithic function.

@kevinyamauchi have you encountered any of these problems? If so, we should try to address them in a future PR.


ambrosejcarr:
Ran across this kinda cool ~thing~ tangent in an xarray issue: https://github.com/pydata/xarray/issues/2355#issuecomment-412246705

Might be another way to get performance for us, although the animation aspect wouldn't be relevant here.

starfish/image/_stack.py
```diff
@@ -368,42 +378,68 @@ def show_stack(
             )
             linear_view = np.clip(linear_view, a_min=a_min, a_max=a_max)
 
-        # Create the plot
+        return linear_view, labels, n_tiles
+
+    def _show_matplotlib_notebook(self, linear_view, labels, n_tiles, show_spots, figure_size, color_map):
+        import matplotlib.pyplot as plt
```
ambrosejcarr: Can we move these imports to the top-level?

starfish/image/_stack.py
```diff
@@ -368,42 +378,68 @@ def show_stack(
...
+        def display_slice(plane_index=0):
+            fig, ax = plt.subplots(figsize=figure_size)
+            title_str = " ".join(str(lab).upper() for lab in labels[plane_index])
+            show_plane(ax, linear_view[plane_index], plane_index, title=title_str, cmap=color_map)
+            plt.show()
 
+        return display_slice
+
+    # TODO ambrosejcarr: update to use IntensityTable instead of SpotAttributes
+    # TODO ambrosejcarr: also, show_spots is probably definitely broken
```
ambrosejcarr: 🙃

kevinyamauchi:
Hey, sorry for the late reply. Apparently I never clicked submit on my comment...

Sorry about this, @dganguli! I have experienced issues (2) and (3). 
 
With regards to issue (2) , it appears that matplotlib is catching the scroll event. I can scroll as long as my cursor is outside of the figure. However, I agree this is annoying.  Apparently,[ others agree this is an issue](https://github.com/matplotlib/matplotlib/issues/8911)

For issue (3), I believe this is because we aren't making a copy of the images and subsequent image operations are performed in place. We can try either making a copy of the image to make the figure. Alternatively, we can try drawing the whole stack upon initialization and [toggling layer visibility via the alpha parameter](https://matplotlib.org/gallery/images_contours_and_fields/image_transparency_blend.html).

kevinyamauchi:
A bit of follow up: I tried pre-plotting all of the frames in show_stack() and then toggling visibility with the interact slider (see cff110d653295ffc56001c1d1169a8a99cae242d) and it seems to take care of issue (3). I did not observe issue (1) after both rerunning a cell with show_stack() and also performing a kernal restart/run all on the notebook. 

@dganguli do you think it makes sense to implement the preplotting in this PR? Maybe it makes sense to just merge this one in just to get the fix in place and then I can open another one with the pre-plotting afterwards.

dganguli:
thanks @kevinyamauchi! i'll go ahead and merge this now, and then if you'd like, implement a PR with the pre-plotting?

starfish/image/_stack.py
```diff
@@ -317,25 +316,36 @@ def show_stack(
 
         """
 
-        from ipywidgets import interact, fixed
-        import matplotlib.pyplot as plt
+        # infer if %matplotlib inline or notebook
+        import matplotlib
```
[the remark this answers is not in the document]
dganguli: good point. i've fixed this.

PR #389, opened 2018-08-10
[1/5] stack to experiment

ambrosejcarr:
These 5 PRs refactor some code to make the API match the proposal in #292. The easiest way to review these PRs is probably to load up PR [5/5], reinstall starfish, import it, and explore the API. 

In this PR: 
- Rename stack to experiment, propagate changes across files and notebooks.


ambrosejcarr:
> should be in this diff:
>
> {stack,experiment}.org should be renamed.

I'm confused by what change you want here. The purpose of this PR is to rename those. 

Is the request changes a request to add a follow-up PR with these changes? 

> should be done, but not in this diff:
>
> - extract a hyb, set of aux images, and codebook at the top of each notebook, rather than continuing to reference the experiment throughout the code.
> - tests like the dataset fixtures and starfish/test/pipeline/test_filter.py should not reference experiment, as they're being done on an ImageStack.

ttung:
> I'm confused by what change you want here. The purpose of this PR is to rename those.

The .org field should be renamed.  It is not meaningful.

> Is the request changes a request to add a follow-up PR with these changes?

PR is ++, but issue is also acceptable. :)

ambrosejcarr:
> The .org field should be renamed. It is not meaningful.

Got it. Sorry for confusion.

ambrosejcarr:
> tests like the dataset fixtures and starfish/test/pipeline/test_filter.py should not reference experiment, as they're being done on an ImageStack.

Issue #401

PR #391, opened 2018-08-10
[2/5] image import

ambrosejcarr:
Move:
- filter to Image
- registration to image
- segmentation to image
- _stack _out_ of image and to the top-level package. 

Hooks up: 
starfish.ImageStack
starfish.image.Filter
starfish.image.Segmentation
starfish.image.Registration


notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
```diff
@@ -155,7 +155,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from starfish.pipeline.filter import Filter\n",
+    "from starfish.image._filter import Filter\n",
```
ttung: We don't want people to have to import hidden modules like this.

I see that starfish.image is importing Filter, so can we just do `from starfish.image import Filter`?

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
```diff
@@ -190,7 +190,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from starfish.pipeline.registration import Registration\n",
+    "from starfish.image._registration import Registration\n",
```
ttung: ditto.

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.py
```diff
@@ -226,7 +226,7 @@
 
 # EPY: START code
 from starfish.constants import Indices
-from starfish.pipeline.segmentation.watershed import _WatershedSegmenter
+from starfish.image._segmentation import _WatershedSegmenter
```
ttung: This is awkward as well.

I have this vague notion that we discussed this before (although maybe it was with @dganguli), and the right strategy is:

```
from starfish.image import Segmentation

segmenter = Segmentation.WatershedSegmenter(...)
segmenter.run(...)
```

notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -1,7 +1,7 @@
...
+# EPY: stripped_notebook: {"metadata": {"hide_input": false, "kernelspec": {"display_name": "Python 3", "language": "python", "name": "python3"}, "language_info": {"codemirror_mode": {"name": "ipython", "version": 3}, "file_extension": ".py", "mimetype": "text/x-python", "name": "python", "nbconvert_exporter": "python", "pygments_lexer": "ipython3", "version": "3.6.5"}, "toc": {"nav_menu": {}, "number_sections": true, "sideBar": true, "skip_h1_title": false, "toc_cell": false, "toc_position": {}, "toc_section_display": "block", "toc_window_display": false}}, "nbformat": 4, "nbformat_minor": 2}
```
ttung: LOL at the war between you and @dganguli about which version of python 3.6. :)

starfish/__init__.py
```diff
@@ -1 +1,9 @@
+# top-level objects
 from .starfish import starfish
+from .intensity_table import IntensityTable
+from .codebook import Codebook
+from .experiment import Experiment
+from ._stack import ImageStack
```
ttung: food for thought: why is _stack the only hidden module?

starfish/image/_filter/__init__.py
```diff
@@ -1,6 +1,6 @@
 import argparse
 
-from starfish.image import ImageStack
+from starfish._stack import ImageStack
```
ttung: Can you just do: from starfish import ImageStack?

notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
ambrosejcarr: We sure can.

starfish/image/_filter/__init__.py
ambrosejcarr: Yes. I'm not opinionated about how the paths should look. There are two options: 

1. Use the paths supported by __init__ in both the code and public API
2. Use the paths supported by __init__ only in the public API, and import from the files that modules are found in the codebase. 

I'm open to either. (1) produces uglier code that's slightly easier to follow than (2) because you don't have to open the __init__ module to figure out where the code lives.

ttung: I don't have a strong feeling about (1) vs (2), but if you want to do (2), then I think we should just make _stack public.

starfish/__init__.py
ambrosejcarr: This is a good question. I'm open to making more or less all of these private. What's your opinion?

ttung: It makes it hard to import other stuff in those packages, potentially. I don't know how common that is.

PR #393, opened 2018-08-10
[5/5] spots import

ambrosejcarr:
- Move target assignment to spots
- Move spots to top-level
- Update import pattern to support

starfish.spots
starfish.spots.Detector
starfish.spots.Decoder
starfish.spots.TargetAssignment


ttung:
Can we minimize the number of imports of hidden packages across modules?

PR #395, opened 2018-08-11
BUG: fix skimage anisotropic kernel

ambrosejcarr:
There is an inconsistency in skimage's gaussian filter documentation which suggests that scalar kernel sizes are broadcast to the image depth. This is not true, and this PR fixes rectifies the inconsistency by broadcasting to full-size kernels in the starfish codebase. Since this code is used by multiple functions, it is factored out into `util.py`.

### DIFF
--- a/starfish/image/_filter/util.py
+++ b/starfish/image/_filter/util.py
@@ -1,7 +1,9 @@
-from typing import Tuple
+from typing import Tuple, Union
 
 import numpy as np
 from skimage.morphology import disk, binary_opening
+
+from starfish.types import Number
 
 
 def bin_thresh(img: np.ndarray, thresh: int) -> np.ndarray:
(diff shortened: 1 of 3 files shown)


starfish/image/_filter/util.py
```diff
@@ -70,3 +72,41 @@ def gaussian_kernel(shape: Tuple[int, int]=(3, 3), sigma: float=0.5):
...
+    if isinstance(sigma, tuple):
+        message = ("if passing an anisotropic kernel, the dimensionality must match the data "
+                   "shape ({shape}), not {passed_shape}")
+        if is_volume and len(sigma) != 3:
+            raise ValueError(message.format(shape=3, passed_shape=len(sigma)))
+        if not is_volume and len(sigma) != 2:
+            raise ValueError(message.format(shape=2, passed_shape=len(sigma)))
+        valid_sigma = sigma
+    else:
+        # TODO skimage is does not broadcast the tuple as declared in their docs. See:
```
ttung: is there an open issue we should be subscribing to?

ambrosejcarr: No, it turns out we're sort-of dumb, and they have bad default arguments. I should remove these TODOs.

PR #396, opened 2018-08-11
increase pandas version to ensure multiindex support

ambrosejcarr:
Old versions of pandas do not support `pd.MultiIndex.to_frame()`

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -1,7 +1,7 @@
 click
 matplotlib
 numpy != 1.13.0
-pandas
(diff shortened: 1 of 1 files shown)


ttung:
Do we need to hard-pin our requirements?

ambrosejcarr:
I have only seen one instance where forward changes have broken something for us (a numpy deprecation). If we start hard-pinning, we're more or less requiring virtual environment based usage, which is not ideal because it's relatively rare in python. I would vote for open-ended until we see a strong reason to be more restrictive.

PR #397, opened 2018-08-11
BUG: fix casting issue in MERFISH notebook that was causing a loss of precision

ambrosejcarr:
- Implements a fix that removes a cast to uint prior to scaling to avoid a loss of precision
- Updates the notebook to work with the new API organization
- Note also that the combination of anisotropic kernel and this change do fix the copy number issue.


notebooks/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -117,6 +115,10 @@
 # EPY: END code
 
 # EPY: START code
+# it's important to convert the data to float here to retain the correct precision for the scaling. Later, starfish
```
ttung: NECRO COMMENT: 

ref: https://github.com/spacetx/starfish/issues/316

PR #406, opened 2018-08-14
Use cloudfront instead of S3 directly.

ttung:
All the errors in travis seem to be related to tests that hit S3 directly.  Hopefully this will improve test reliability.


ttung:
> Oh interesting, that's good to know. You deduced it by A/B-ing the notebooks?

no, I visually checked the failed travis runs and noticed a pattern. :)

ambrosejcarr:
@ttung Thinking about the future, could we add something to our linter that catches use of aws links?

ttung:
Probably, but likely not a high priority?  It fails pretty quickly so catching it is not super hard.

PR #405, opened 2018-08-14
[9/5] ENH: replace hidden imports with public ones

ambrosejcarr:
This PR: 

- Adds `SpotAttributes` as a public object in `starfish.spots`
- Fixes all private imports in the non-test code.

PR #428, opened 2018-08-15
Unbreak ISS pipeline

ttung:
Test plan: ran `make run__notebooks/py/ISS_Pipeline_-_Breast_-_1_FOV.py` successfully.

### DIFF
--- a/notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
+++ b/notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
@@ -216,7 +216,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
#404 and #405 fix the ISS notebook. This is redundant to those changes, I believe. I've responded to the most recent round of comments on #404.

ttung:
> #404 and #405 fix the ISS notebook. This is redundant to and possibly in conflict with those changes, I believe. I've responded to the most recent round of comments on #404.

ok.
i just didn't want to leave master in a broken state. 

> What was the reason for the change from detector to SpotFinder? Just more clear?

@ambrosejcarr correct me if I'm wrong, but I think this is the reorg to make the packages less nested and easier for notebook users to deal with.

PR #429, opened 2018-08-15
TST: add stronger assertions to the MERFISH and ISS_Pipeline notebooks

ambrosejcarr:
- Adds a series of strong assertions to the two notebooks that are currently run on master: MERFISH and ISS. 
- These will help diagnose the locations of any breakages until we can convert sub-selections of the image data to small unit tests.

### DIFF
--- a/notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
+++ b/notebooks/ISS_Pipeline_-_Breast_-_1_FOV.ipynb
@@ -211,6 +211,28 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# assert on a small piece of the image to verify that the filters and registration have not changed\n",
+    "expected = np.array(\n",
+    "    [[ 67,  95, 117, 112, 171, 248, 229, 254, 314, 405],\n",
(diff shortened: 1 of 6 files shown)


ttung:
👍

ambrosejcarr:
@dganguli I tried glomming the text blocks, but that requires us to change everything to inplace=False.  My attempts didn't even get past registration in the ISS notebook. 

Registration is currently in-place only, although 50/50 it's a problem with something above. If you want this change in this PR, it will take me a bit to get to it.

dganguli:
It's important to glom test blocks at the end. I think any of the following two approaches are fine:

1. The Tortuous Path -- do inplace computation = False. I think that's no more than 20 min. of work for everything but registration. For registration, this is a good opportunity to re-factor the Registration class to use xarray and expose inplace optionality. Could leverage the ideas from Filters.ZeroByChannelMagnitude.

2. The  Path of Least Resistance -- don't test images, only leave assertions about decoded results (e.g., spots, genes, correlations, etc.) at the end. 

3. Enlightenment -- you've basically written the unit tests. should we just skip this notebook nonsense and write the tests? 

4. Follow the Path of Least Resistance towards Enlightenment. Leave the Tortuous Path for the open issues re: registration by berl.

ambrosejcarr:
Closing this in favor of unit-tests for each pipeline.

PR #433, opened 2018-08-16
Move constants, spot attributes, and validated tables into new module

ttung:
We're starting to see some circular dependency issues, and that generally means we need to organize our code better.  Imports need to be disentangled, and one way to do that is to structure the code so that the imports go in one direction.  I've created a new module, called types, which encompasses constants, spot attributes, and validated tables.  These modules have no external dependencies that are within the starfish code base, which is what we need!


ttung:
> Can you talk through why you didn't move the actual code for SpotAttributes and the stuff in constants to types? I'd rather avoid confusing import paths when possible.

https://github.com/spacetx/starfish/pull/433/files#diff-7278f3a27dd865af85b9d27b0653c3c1
https://github.com/spacetx/starfish/pull/433/files#diff-c0fc7af232801b915e24ce2f37d2103c
https://github.com/spacetx/starfish/pull/433/files#diff-bacc072a6b834252505f4c195886a384
https://github.com/spacetx/starfish/pull/433/files#diff-7e8244f184c4e52d169b485e731e024d

PR #435, opened 2018-08-16
Create experiment package.

ttung:
This is the package for interacting with and creating experiment objects.

1. Rename starfish/experiment.py to starfish/experiment/__init__.py
2. Move examples/support/__init__.py into starfish/experiment/
3. Rename examples/build_sample_experiment.py to starfish/experiment/builder/cli.py

### DIFF
--- a/starfish/experiment/builder/imagedata.py
+++ b/starfish/experiment/builder/imagedata.py
@@ -1,0 +1,68 @@
+from io import BytesIO
+from typing import IO, Tuple
+
+import numpy as np
+from skimage.io import imsave
+from slicedimage import (
+    ImageFormat,
+)
+
+
+class FetchedImage:
+    """
+    This is the contract for providing the data for constructing a :class:`slicedimage.Tile`.
+    """
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        raise NotImplementedError()
+
+    @property
+    def format(self) -> ImageFormat:
+        raise NotImplementedError()
+
+    @property
+    def image_data_handle(self) -> IO:
+        raise NotImplementedError()
+
+
+class ImageFetcher:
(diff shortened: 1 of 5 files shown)


starfish/experiment/builder/image.py
```diff
@@ -0,0 +1,67 @@
+from io import BytesIO
```
ambrosejcarr: This produces `starfish.image` and `starfish.experiment.builder.image`. I wonder if there is a more intuitive construction. I think you have the stuff in the right place but perhaps `image` is not the right name for this module. 

I feel like `io` is a better descriptor, perhaps something along this line that doesn't conflict with the stdlib?

ttung: Renamed to imagedata.

starfish/experiment/builder/imagedata.py
```diff
@@ -0,0 +1,68 @@
...
+        return 1536, 1024
+
+    @property
+    def format(self) -> ImageFormat:
+        return ImageFormat.TIFF
+
+    def image_data_handle(self) -> IO:
+        arr = np.random.randint(0, 256, size=self.shape, dtype=np.uint8)
+        output = BytesIO()
+        imsave(output, arr, plugin="tifffile")
```
dganguli: is this supposed to be imload?

ttung: Nope.  The contract for `FetchedImage` is that it returns a file handle to the file data.  Since we are generating this data in-memory, we create a `BytesIO` object (which is a block of memory addressed as a file), write the random noise image data to it, rewind it back to the beginning, and return that.  When the consumer of the `FetchedImage` gets the file handle, it reads the image we generated.

Issue #418, opened 2018-08-14
Use FOV builder to generate synthetic datasets for ingestion


ambrosejcarr:
Once the tool to generate FOVs is done, use it to unblock our collaborators

3. Generate a few datasets of varying (small) scales
2. Put said data into a public bucket or repository so that HCA folks can begin trying to ingest them.

ambrosejcarr:
Closed by #464, #478

PR #436, opened 2018-08-16
Create entry point for starfish/experiment/builder/cli.py

ttung:
1. Create a "main" method for the script.
2. Add an entry point for the pip package.

Test plan:
```
% pip install -e .
% build_sample_experiment --help
```

Connects to #418

### DIFF
--- a/starfish/experiment/builder/cli.py
+++ b/starfish/experiment/builder/cli.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+from typing import MutableMapping
 
 from starfish.types import Indices
 from starfish.util.argparse import FsExistsType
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,7 +26,10 @@
     packages=setuptools.find_packages(),
     install_requires=install_requires,
     entry_points={
-        'console_scripts': "starfish=starfish.starfish:starfish"
+        'console_scripts': [
+            "starfish=starfish.starfish:starfish",
+            "build_sample_experiment=starfish.experiment.builder.cli:main",
```
joshmoore: I think b_s_e would surprise me if it showed up on `$PATH` after `pip install starfish`

ambrosejcarr: Yep, I had similar thoughts. I think it might make more sense to include it as a sub-module of starfish. For now, tagging it with a name that associates with our package is enough for me because I want to get this out to collaborators. 

Suggestion: 
`build_sample_starfish_experiment` + new issue to include it in the starfish binary.

ttung:
Made this a subcommand of the starfish cli.  Example invocation: `starfish build --fov-count 2 --hybridization-dimensions '{"c": 2, "h": 2, "z": 3}' /tmp/foo`

ambrosejcarr:
Lol, love that as a side effect of this change, it also displays the starfish ascii art.

ttung:
> Lol, love that as a side effect of this change, it also displays the starfish ascii art.

The most important change.  That the tool is wired up is the side effect. :)

PR #437, opened 2018-08-16
Progress bar

shanaxel42:
Adding a the tqdm progress bar lib to stack.read()


ttung:
Does this require https://github.com/spacetx/slicedimage/pull/30 to land first?

shanaxel42:
No it'll still work, but the actual bar won't show up unless it does.

dganguli:
awesome!

PR #445, opened 2018-08-18
BUG: fix verbosity handling and set multichannel false

ambrosejcarr:
* Gaussian filters were not handling verbosity properly.
* Multichannel should be false for our images, as the last axis does not
contains multi-channel data.


dganguli:
wait, does the MERFISH notebook still run after the changes to the verbose flag?

ambrosejcarr:
All pipeline components accept `**kwargs`, so you can feel them anything, for better or worse. I don't think it should be affected. 

e.g. `filt = Filter.GaussianLowPass(sigma=(3, 3, 3), man_starfish_is_great=True)` should be fine.

PR #447, opened 2018-08-18
ENH: change single_synthetic_spot to a factory function

ambrosejcarr:
First of many, many PRs that change fixtures into factories. In this case, I also renamed it to have a more informative name.


ambrosejcarr:
👋 @ttung

PR #448, opened 2018-08-18
ENH: enable out-of-place-registration

ambrosejcarr:
Enables out-of-place registration.

### DIFF
--- a/starfish/image/_registration/fourier_shift.py
+++ b/starfish/image/_registration/fourier_shift.py
@@ -1,4 +1,5 @@
-from typing import Tuple, Union
+from copy import deepcopy
+from typing import Optional, Tuple, Union
 
 import numpy as np
 from scipy.ndimage import fourier_shift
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
Rebased and confirmed this works on master with Tony's cache fix.

starfish/image/_registration/fourier_shift.py
```diff
@@ -34,7 +48,25 @@ def add_arguments(cls, group_parser) -> None:
...
+        Parameters
+        ----------
+        image : ImageStack
+            The stack to be registered
+        in_place : bool
+            If false, return a new registered stack. Else, register in-place (default False)
+
+        Returns
+        -------
+
```
ttung: stuff missing.

starfish/image/_registration/fourier_shift.py
```diff
@@ -54,7 +86,9 @@ def run(self, image: ImageStack) -> ImageStack:
                     result = shift_im(data, shift)
                     image.set_slice(indices=indices, data=result)
 
-        return image
+        if not in_place:
+            return image
+        return None
```
ttung: technically not required.

ambrosejcarr: Mypy got upset without it there at one point. I will check if it gets upset here, and if not will remove it.

ttung:
> Generally speaking, I think we should have image->image pipeline components return ImageStack, even if it's in-place. You can string up a bunch of operations together that way.

curious what you think of ^^, in case you missed it. :)

ambrosejcarr:
> Generally speaking, I think we should have image->image pipeline components return ImageStack, even if it's in-place. You can string up a bunch of operations together that way.

Yep, I see the advantage and am fine with that.

PR #450, opened 2018-08-18
[Easy] fix imagestack.write so that it correct appends a prefix

ambrosejcarr:
- Makes the outcome (`hybridization_images.json` has `.json`) as one would expect from the request for path+prefix.

### DIFF
--- a/starfish/stack.py
+++ b/starfish/stack.py
@@ -761,12 +761,12 @@ class ImageStack:
         return self._tile_shape
 
     def write(self, filepath: str, tile_opener=None) -> None:
-        """write the image tensor to disk
+        """write the image tensor to disk in spaceTx format
 
         Parameters
         ----------
         filepath : str
-            path + prefix for writing the image tensor
+            Path + prefix for the images and hybridization_images.json written by this function
         tile_opener : TODO ttung: doc me.
 
         """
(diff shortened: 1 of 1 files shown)


starfish/stack.py
```diff
@@ -825,7 +825,7 @@ def tile_opener(tileset_path, tile, ext):
 
         Writer.write_to_path(
             self._image_partition,
-            filepath,
+            filepath + '.json',
```
dganguli: Hmmm. Maybe check if .json is already in 'filepath' before adding? Or throw an exception if someone is passing in a file instead of a path? I often do .write('blah.json'). So this would make my stuff write to 'blah.json.json' right?

starfish/stack.py
```diff
@@ -761,12 +761,12 @@ def tile_shape(self):
...
 
     def write(self, filepath: str, tile_opener=None) -> None:
-        """write the image tensor to disk
+        """write the image tensor to disk in spaceTx format
 
         Parameters
         ----------
         filepath : str
-            path + prefix for writing the image tensor
+            Path + prefix for the images and hybridization_images.json written by this function
```
ttung: post-commit comment: I don't understand this.

starfish/stack.py
```diff
@@ -823,6 +823,8 @@ def tile_opener(tileset_path, tile, ext):
                     ),
                     "wb")
 
+        if not filepath.endswith('.json'):
```
ttung: Why is this desirable?  One does not call `imsave("foo",... )` and expect to get foo.tiff

PR #452, opened 2018-08-19
ENH: Exchange multiindex for normal xarray coordinates

ambrosejcarr:
This PR removes the IntensityTable multiindex in favor of multiple independent coordinates. This does not appear to have any negative impact on indexing and passes all tests. I've updated all notebooks to continue to work. 

@dganguli I think this is more maintainable because it hews closer to the core data structures of xarray, however I'd like for you to play with the result and see if it feels better from a UX perspective. 

* rename to_dataframe -> features_to_dataframe (to_dataframe is a method
of xarray)
* change multiindex code for normal xarray coordinates, which are
propagated by groupby
* remove dataframe munging
* minor updates to notebooks to work with new system.


starfish/intensity_table.py
```diff
@@ -147,18 +160,23 @@ def from_spot_data(
...
+            spot_attributes,
+            np.arange(intensities.shape[1]),
+            np.arange(intensities.shape[2]))
 
-        coords = (
-            (Features.AXIS, spot_attributes),
-            (Indices.CH.value, np.arange(intensities.shape[1])),
-            (Indices.ROUND.value, np.arange(intensities.shape[2]))
-        )
+        # coords = (
```
dganguli: did you mean to leave these comments in?

starfish/intensity_table.py
```diff
@@ -236,9 +254,12 @@ def synthetic_intensities(
         x = np.random.randint(0, width, size=n_spots)
         r = np.empty(n_spots)
         r.fill(np.nan)  # radius is a function of the point-spread gaussian size
+        spot_attributes = pd.DataFrame(
+            {Features.Z: z, Features.Y: y, Features.X: x, Features.SPOT_RADIUS: r}
+        )
 
-        names = [Features.Z, Features.Y, Features.X, Features.SPOT_RADIUS]
-        spot_attributes = pd.MultiIndex.from_arrays([z, y, x, r], names=names)
+        # names = [Features.Z, Features.Y, Features.X, Features.SPOT_RADIUS]
```
dganguli: remove?

starfish/intensity_table.py
```diff
@@ -317,10 +338,11 @@ def from_image_stack(
         pixel_coordinates[Features.SPOT_RADIUS] = np.full(
             pixel_coordinates.shape[0], fill_value=np.nan)
 
-        spot_attributes = dataframe_to_multiindex(pixel_coordinates)
+        # spot_attributes = dataframe_to_multiindex(pixel_coordinates)
```
dganguli: remove

starfish/intensity_table.py
```diff
@@ -338,7 +360,7 @@ def _intensities_from_regions(self, props, reduce_op='max') -> "IntensityTable":
         """turn regions back into intensities by reducing over the labeled area"""
         raise NotImplementedError
 
-    def to_dataframe(self) -> pd.DataFrame:
+    def features_to_dataframe(self) -> pd.DataFrame:
```
dganguli: I like typing something.to, and having it tab-complete, to figure out how I can convert some wacky object that i don't understand into a less wacky object that i may potentially understand better. is there a name here that starts w/ to? e.g., to_dataframe_features, to_features_dataframe, etc.?

ambrosejcarr: hah, this is a great point. `to_features_dataframe` it is.

starfish/intensity_table.py
ambrosejcarr: No, sorry. 😞

starfish/intensity_table.py
```diff
@@ -57,9 +56,22 @@ class IntensityTable(xr.DataArray):
     # constant that stores the attr key for the shape of ImageStack
     IMAGE_SHAPE = 'image_shape'
 
+    @staticmethod
+    def _build_coordinates(
```
dganguli: i got confused here, coordinates here doesn't mean Coordinates.X, Y, Z, etc. But rather, it means Xarray Coordinates right? If so, consider re-naming to _build_xarray_coordinates?

starfish/munge.py
```diff
@@ -1,16 +1,9 @@
 from typing import Any, Dict
 
 import numpy as np
-import pandas as pd
 import regional
 
 
-def dataframe_to_multiindex(dataframe: pd.DataFrame) -> pd.MultiIndex:
```
dganguli: insert party jeremy here.

starfish/spots/_detector/detect.py
```diff
@@ -144,7 +142,7 @@ def concatenate_spot_attributes_to_intensities(
     n_round: int = max(inds[Indices.ROUND] for _, inds in spot_attributes) + 1
 
     all_spots = pd.concat([sa.data for sa, inds in spot_attributes])
-    spot_attribute_index = dataframe_to_multiindex(all_spots.drop(['spot_id', 'intensity'], axis=1))
+    features_coordinates = all_spots.drop(['spot_id', 'intensity'], axis=1)
```
dganguli: i had to stare at this awhile. maybe add a comment, that this only keeps columns: x, y, z, radius, quality? or something like that.

starfish/intensity_table.py
ambrosejcarr: That's correct. Will fix. 👍

ambrosejcarr:
Yea, so the `features` dimension needs some kind of index. Thinking about what this would mean, if you made it `index` it would imply that `IntensityTable.index` would operate on the features dimension. 

Technically, `c, r` are also first-class dimensions, so `IntensityTable.c` would be the comparable operation on the channel dimension. 

Are you ok with that interaction? We could also call it `features_index` to be a bit more specific, but it's unnecessarily clunky if you like the above.

starfish/intensity_table.py
dganguli: actually, idiomatic xarray calls coordinates, coords. so maybe _build_xarray_coords

dganguli:
Now that you explain it, and I understand it, I think features makes a lot of sense. The asterisk deliniates it as an index, that applies to all variables with (features) appended to them. So I vote leave as is!

PR #453, opened 2018-08-20
ENH: Remove IntensityTable's image_shape

ambrosejcarr:
* not necessary if we are not supporting reversion from IntensityTable to ImageStack
* adds flexibility to IntensityTable, as now we don't need to pass along an attr.

### DIFF
--- a/starfish/intensity_table.py
+++ b/starfish/intensity_table.py
@@ -72,7 +72,6 @@ class IntensityTable(xr.DataArray):
     @classmethod
     def empty_intensity_table(
             cls, spot_attributes: pd.DataFrame, n_ch: int, n_round: int,
-            image_shape: Tuple[int, int, int]
     ) -> "IntensityTable":
         """Create an empty intensity table with pre-set axis whose values are zero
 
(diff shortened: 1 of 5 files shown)


dganguli:
Right now we have 3 things that are xarrays:

ImageStack
IntensityTable
Codebook

Did we land on merging IntensityTable w/ ImageStack, but keeping Codebook separate? 

Although this seems logically intuitive, I do think there is a useful separation of concerns in ImageStack and IntensityTable. The former is strictly for image processing, visualization, and manipulation. The latter is for analysis of spot attributes, putative barcodes, and general QC for decoded results. The code in the two files reflects this well. As such, I vote we keep these things separate.

On the other hand, this separation is not as clean as I would like (viz a vis image_shape). I also vote for thinking harder about this separation instead of smashing these two classes together prematurely. This requires more thought.

dganguli:
FWIW my comments from above come predominantly from the perspective my using both objects to analyze >5 datasets. Less so from an eng perspective.

ambrosejcarr:
I recall that it made a lot of sense to merge if only considering the `features-as-pixels` case, but less so for the `spots-as-features` case. 

We also liked that having an IntensityTable produces some logical separation between the interfaces of say `Detect`, which is an `ImageStack -> IntensityTable` operation, and `Filter`, which is `ImageStack -> ImageStack`. In the "merged" case, one could feasibly filter an "image" composed of spots in the form of an IntensityTable/ImageStack, since they'd be the same object. Things get a bit jumbled.

ttung:
I still think there is a compelling reason to merge. :)

However, since we're not doing that, I think this is reasonable.

ttung:
Can you rebase?

PR #454, opened 2018-08-20
Upgrade to nbencdec 0.0.5

ttung:
Picks up this sort-of [format change](https://github.com/ttung/nbencdec/commit/9e3548c1aef920f3bf4cbec4f9894504777a23d2).

Reason it's not completely breaking is that it still decodes correctly, just that each markdown cell will have a space in front of each line.

Rationale for this change: markdown cells get prefixed with "# ", and if it's just an empty line, and an editor comes along and strips trailing whitespace, then it becomes "#", which doesn't decode the same as "# ".

### DIFF
--- a/REQUIREMENTS-DEV.txt
+++ b/REQUIREMENTS-DEV.txt
@@ -4,5 +4,5 @@ jsonpath_rw
 pytest-cov>=2.5.1
 pytest-xdist
 mypy
-nbencdec
+nbencdec >= 0.0.5
 -r REQUIREMENTS.txt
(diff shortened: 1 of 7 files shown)


ambrosejcarr:
Is this path to fixing outstanding PRs to run `make regenerate_py` after rebase?

ttung:
I had to regenerate all the encoded `.py` files for this.

ttung:
> Is the path to fixing outstanding PRs to run make regenerate_py after rebase?

Yeah, you would need to rebase, install nbencdec 0.0.5, and `make -j regenerate_py`

PR #460, opened 2018-08-22
Point to data that has correct checksums.

ttung:
This is a one-time deal (hopefully!) so I didn't bother building a proper tool to do this.


ttung:
> If the other notebooks had correct checksums, and that's why they're missing, LGTM.

This is true as far as I can tell.

PR #464, opened 2018-08-22
Add example of raw->spacetx format conversion for ISS Breast Data

dganguli:
This PR:
1. Introduces get_iss_breast_data.py, which converts raw data from s3://czi.starfish.data.public/browse/raw/20180820/iss_breast/ to spaceTx formatted data. This will be our first real multi-fov dataset. 

2. re-names things. image -> tile, image_fetcher -> tile_fetcher

I tested this PR in a notebook locally. I downloaded the 1 FOV from he breast data formatter and ran the ISS breast notebook against it. It worked.

Remaining work for @ambrosejcarr :
1. Run this tool.
2. Upload formatted data to s3 browsable.
3. Copy codebook.json from 1 FOV dataset over to latter bucket.

Remaining work for @ttung :
1. Get cracking on experiment API so that...

Remaining work for Deep:
1. Use experiment API to analyze this entire dataset (blocked by @ttung )
2. Generate copy number plot for call 2 (high priority @ttung )
3. Start spec'ing out what multi-fov experiment processing will require (blocked by @ttung )

### DIFF
--- a/examples/get_iss_breast_data.py
+++ b/examples/get_iss_breast_data.py
@@ -44,19 +44,20 @@ class ISSCroppedBreastPrimaryTileFetcher(TileFetcher):
 
     @property
     def ch_dict(self):
-        ch_str = ['Cy3 5', 'Cy3', 'Cy5', 'FITC']
-        ch = [2, 1, 3, 0]
-        ch_dict = dict(zip(ch, ch_str))
+        ch_dict = {0: 'FITC', 1: 'Cy3', 2: 'Cy3 5', 3: 'Cy5'}
         return ch_dict
 
     @property
     def hyb_dict(self):
         hyb_str = ['1st', '2nd', '3rd', '4th']
-        return dict(zip(range(4), hyb_str))
+        hyb_dict = dict(enumerate(hyb_str))
+        return hyb_dict
 
     def get_tile(self, fov: int, hyb: int, ch: int, z: int) -> FetchedTile:
-        filename = 'slideA_' + str(fov + 1) + '_' + \
-                   self.hyb_dict[hyb] + '_' + self.ch_dict[ch] + '.TIF'
+        filename = 'slideA_{}_{}_{}.TIF'.format(str(fov + 1),
+                                                self.hyb_dict[hyb],
+                                                self.ch_dict[ch]
+                                                )
         file_path = os.path.join(self.input_dir, filename)
         return IssCroppedBreastTile(file_path)
 
@@ -68,9 +69,9 @@ class ISSCroppedBreastAuxTileFetcher(TileFetcher):
 
     def get_tile(self, fov: int, hyb: int, ch: int, z: int) -> FetchedTile:
         if self.aux_type == 'nuclei':
-            filename = 'slideA_' + str(fov + 1) + '_DO_' + 'DAPI.TIF'
+            filename = 'slideA_{}_DO_DAPI.TIF'.format(str(fov + 1))
         elif self.aux_type == 'dots':
-            filename = 'slideA_' + str(fov + 1) + '_DO_' + 'Cy3.TIF'
+            filename = 'slideA_{}_DO_Cy3.TIF'.format(str(fov + 1))
         else:
             msg = 'invalid aux type: {}'.format(self.aux_type)
             msg += ' expected either nuclei or dots'
@@ -82,12 +83,6 @@ class ISSCroppedBreastAuxTileFetcher(TileFetcher):
 
 
 def format_data(input_dir, output_dir):
-    if not input_dir.endswith("/"):
-        input_dir += "/"
-
-    if not output_dir.endswith("/"):
-        output_dir += "/"
-
     def add_codebook(experiment_json_doc):
         experiment_json_doc['codebook'] = "codebook.json"
         return experiment_json_doc
--- a/examples/get_iss_breast_data.py
+++ b/examples/get_iss_breast_data.py
@@ -83,6 +83,23 @@ class AuxImageFetcherDots(ImageFetcher):
 
 
 def format_data(input_dir, output_dir):
+
+    if not input_dir.endswith("/"):
+        input_dir += "/"
+
+    if not output_dir.endswith("/"):
+        output_dir += "/"
+
+    def add_codebook(experiment_json_doc):
+        experiment_json_doc['codebook'] = "codebook.json"
+
+        # TODO: (ttung) remove the following unholy hacks.  this is because we want to point at a
+        # tileset rather than a collection.
+        experiment_json_doc['hybridization_images'] = "hybridization-fov_000.json"
+        experiment_json_doc['auxiliary_images']['nuclei'] = "nuclei-fov_000.json"
+        experiment_json_doc['auxiliary_images']['dots'] = "dots-fov_000.json"
+        return experiment_json_doc
+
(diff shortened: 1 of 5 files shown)


examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+        ch_dict = dict(zip(ch, ch_str))
+        return ch_dict
+
+    @property
+    def hyb_dict(self):
+        hyb_str = ['1st', '2nd', '3rd', '4th']
+        return dict(zip(range(4), hyb_str))
+
+    def get_tile(self, fov: int, hyb: int, ch: int, z: int) -> FetchedTile:
+        filename = 'slideA_' + str(fov + 1) + '_' + \
```
ttung: string concatenation is a huge mess and very hard to read.  consider switching to [f-style python strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings).

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+        return fh
+
+
+class ISSCroppedBreastPrimaryTileFetcher(TileFetcher):
+    def __init__(self, input_dir):
+        self.input_dir = input_dir
+
+    @property
+    def ch_dict(self):
+        ch_str = ['Cy3 5', 'Cy3', 'Cy5', 'FITC']
```
ttung: not sure why this is better than just a literal dict.

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+        ch = [2, 1, 3, 0]
+        ch_dict = dict(zip(ch, ch_str))
+        return ch_dict
+
+    @property
+    def hyb_dict(self):
+        hyb_str = ['1st', '2nd', '3rd', '4th']
```
ttung: ummmm, `dict(enumerate(hyb_str))`?

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+        self.input_dir = input_dir
+        self.aux_type = aux_type
+
+    def get_tile(self, fov: int, hyb: int, ch: int, z: int) -> FetchedTile:
+        if self.aux_type == 'nuclei':
+            filename = 'slideA_' + str(fov + 1) + '_DO_' + 'DAPI.TIF'
+        elif self.aux_type == 'dots':
+            filename = 'slideA_' + str(fov + 1) + '_DO_' + 'Cy3.TIF'
+        else:
+            msg = 'invalid aux type: {}'.format(self.aux_type)
```
ttung: ahhh, string concatenation!

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+            msg += ' expected either nuclei or dots'
+            raise ValueError(msg)
+
+        file_path = os.path.join(self.input_dir, filename)
+
+        return IssCroppedBreastTile(file_path)
+
+
+def format_data(input_dir, output_dir):
+    if not input_dir.endswith("/"):
```
ttung: these should not be necessary.

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+        input_dir += "/"
+
+    if not output_dir.endswith("/"):
+        output_dir += "/"
+
+    def add_codebook(experiment_json_doc):
```
ttung: hmmmm, i think we need a better way to build codebooks programmatically....

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,139 @@
...
+                          )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("input_dir", type=FsExistsType())
+    parser.add_argument("output_dir", type=FsExistsType())
+
+    args = parser.parse_args()
+    s3_bucket = "s3://czi.starfish.data.public/browse/raw/20180820/iss_breast/"
```
ttung: i don't get this.  who's going to do this?

ttung:
Is it possible to generate a full experiment manifest from this?

ambrosejcarr:
@ttung I think this is a WIP that I'm supposed to take over. I was looking through the code yesterday and my main question was "how does one pass a directory with multiple fields of view and expect them to be parsed separately? Do I write `os.path.walk()` into the TileFetcher? 

An example of that would be very helpful. 

When I'm finished using this tool to upload the s3 data, I will add a README to it.

examples/get_iss_breast_data.py
dganguli: who cares.

examples/get_iss_breast_data.py
dganguli: lol i copied this from your code in get_iss.py. i can nuke it.

examples/get_iss_breast_data.py
dganguli: ambrose is going to do this. josh is going to read this code. this is also an example to help collaborators understand how to convert their data.

dganguli:
@ttung  what do you mean is it possible to generate a full experiment manifest from this. That is the entire point. Unless you have a different definition of experiment manifest than I do? The problem, is that we don't have a way of loading this full experiment manifest. That's why we need the multi-fov API ASAP!

@ambrosejcarr parsing images that live in separate directories is easy; however, that's super unimportant to worry about now. regardless, the way you would handle it, is in TileFetcher.get_tile(...). You would implement the logic needed os.path.join self.input_dir to the right thing based on fov_num, hyb, ch, etc. a more interesting question, is what if you'd like build_experiment_json to _write_ hybridization_fov_xxx.jsons to different directories.

dganguli:
@ttung i addressed all ur comments in the PR. thx!

ambrosejcarr:
@dganguli Ok, so if I wanted to use this tool as is, I should put all the existing images that need to be formatted into the same directory, and it will write a series of hybridization.json files into a single output directory? 

I was for some reason stuck on the prior that each FOV would live in its own directory and couldn't reconcile how this would work in that case. I think I'm straightened out.

dganguli:
@ttung can u approve and merge? i'd like to get the corresponding merfish work done today as well.

examples/get_iss_breast_data.py
```diff
@@ -0,0 +1,134 @@
...
+    @property
+    def format(self) -> ImageFormat:
+        return ImageFormat.TIFF
+
+    @staticmethod
+    def crop(img):
+        crp = img[40:1084, 20:1410]
+        return crp
+
+    def tile_data_handle(self) -> IO:
```
ttung: i think this is supposed to be a `@property`. it worked before due to a fluke, but the new slicedimage api will require it.

you can make the change now or i can make it when i land the slicedimage change.

examples/get_iss_breast_data.py
ttung: I think this message is confusing. i would just have the `--help` output tell people where to get the data.

Rationale: the only time this message prints is if the command is running, and to do that, you kind of need the data already.

examples/get_iss_breast_data.py
dganguli: will fix

examples/get_iss_breast_data.py
dganguli: 👍

PR #479, opened 2018-08-24
Use slicedimage's backends to load codebooks.

ttung:
This allows us to use file:/// schema URLs.


ttung:
LOL this depends on slicedimage 0.0.3.  Well, soonish!

PR #481, opened 2018-08-24
RFC: experiment API

ttung:
I haven't converted the existing code or notebooks to use this API, but I'll paste in a few sample runs.

### DIFF
--- a/starfish/experiment/__init__.py
+++ b/starfish/experiment/__init__.py
@@ -1,81 +1,134 @@
 import json
-from typing import Mapping, Optional
+from typing import Callable, MutableMapping, MutableSequence, Optional, Sequence, Set, Union
 
-import numpy as np
 from semantic_version import Version
-from slicedimage.io import resolve_path_or_url
+from slicedimage import Collection, TileSet
+from slicedimage.io import Reader, resolve_path_or_url, resolve_url
+from slicedimage.urlpath import pathjoin
 
+from starfish.codebook import Codebook
 from starfish.stack import ImageStack
 
 
+class FieldOfView:
+    """
+    This encapsulates a field of view.  It contains the primary image and auxiliary images that are
+    associated with the field of view.
+
+    Auxiliary images can be accessed using a key, i.e., FOV[aux_image_type].
+
+    Properties
+    -------
+    name                   The name of the FOV.  In an experiment with only a single FOV, this may
+                           be None.
+    primary_image          The primary image for this field of view.
+    auxiliary_image_types  A set of all the auxiliary image types.
+    """
+    def __init__(
+            self,
+            name: Optional[str],
+            primary_image: Optional[ImageStack]=None,
+            auxiliary_images: Optional[MutableMapping[str, ImageStack]]=None,
+            primary_image_tileset: Optional[TileSet]=None,
+            auxiliary_image_tilesets: Optional[MutableMapping[str, TileSet]]=None,
+    ) -> None:
+        """
+        Fields of views can obtain their primary image from either an ImageStack or a TileSet (but
+        only one).  It can obtain their auxiliary image dictionary from either a dictionary of
+        auxiliary image name to ImageStack or a dictionary of auxiliary image name to TileSet (but
+        only one).
+
+        Note that if the source image is from a TileSet, the decoding of TileSet to ImageStack does
+        not happen until the image is accessed.  Be prepared to handle errors when images are
+        accessed.
+        """
+        self._name = name
+        self._primary_image: Union[ImageStack, TileSet]
+        self._auxiliary_images: MutableMapping[str, Union[ImageStack, TileSet]]
+        if primary_image is not None:
+            self._primary_image = primary_image
+            if primary_image_tileset is not None:
+                raise ValueError(
+                    "Only one of (primary_image, primary_image_tileset) should be set.")
+        elif primary_image_tileset is not None:
+            self._primary_image = primary_image_tileset
+        else:
+            raise ValueError("Field of view must have a primary image")
+        if auxiliary_images is not None:
+            self._auxiliary_images = auxiliary_images
+            if auxiliary_image_tilesets is not None:
+                raise ValueError(
+                    "Only one of (auxiliary_images, auxiliary_image_tilesets) should be set.")
+        elif auxiliary_image_tilesets is not None:
+            self._auxiliary_images = auxiliary_image_tilesets
+        else:
+            self._auxiliary_images = dict()
+
+    @property
+    def name(self) -> Optional[str]:
+        return self._name
+
+    @property
+    def primary_image(self) -> ImageStack:
+        if isinstance(self._primary_image, TileSet):
+            self._primary_image = ImageStack(self._primary_image)
+        return self._primary_image
+
+    @property
+    def auxiliary_image_types(self) -> Set[str]:
+        return set(self._auxiliary_images.keys())
+
+    def __getitem__(self, item) -> ImageStack:
+        if isinstance(self._auxiliary_images[item], TileSet):
+            self._auxiliary_images[item] = ImageStack(self._auxiliary_images[item])
+        return self._auxiliary_images[item]
+
+
 class Experiment:
+    """
+    This encapsulates an experiment, with one or more fields of view and a codebook.  An individual
+    FOV can be retrieved using a key, i.e., experiment[fov_name].
+
+    Constructors
+    -------
+    from_json     Given a URL or a path to an experiment.json document, return an Experiment object
+                  corresponding to the document.
+
+    Methods
+    -------
+    fov           Given a callable that accepts a FOV, return the first FOVs that the callable
+                  returns True when passed the FOV.  Because there is no guaranteed sorting for the
+                  FOVs, use this cautiously.
+    fovs          Given a callable that accepts a FOV, return all the FOVs that the callable returns
+                  True when passed the FOV.
+    fovs_by_name  Given one or more FOV names, return the FOVs that match those names.
+
+    Properties
+    -------
+    codebook      Returns the codebook associated with this experiment.
+    extras        Returns the extras dictionary associated with this experiment.
+    """
     CURRENT_VERSION = Version("1.0.0")
     MIN_SUPPORTED_VERSION = Version("1.0.0")
     MAX_SUPPORTED_VERSION = Version("1.0.0")
 
-    def __init__(self):
-        # data organization
-        self.format_metadata = None
-        self.image = None
-
-        # auxiliary images
-        self.auxiliary_images = dict()
-
-        # readers and writers
-        self.write_fn = np.save  # asserted for now
-
-        # backend & baseurl
-        self.backend = None
-        self.baseurl = None
-
-    @classmethod
-    def from_data(
-            cls,
-            image_stack: ImageStack,
-            aux_dict: Optional[Mapping[str, ImageStack]]=None
-    ) -> "Experiment":
-        """create a Stack from an already-loaded ImageStack
-
-        Parameters
-        ----------
-        image_stack : ImageStack
-            in-memory ImageStack
-        aux_dict : Optional[Mapping[str, ImageStack]]
-            a dictionary of ImageStacks, default None
-
-        Returns
-        -------
-        Stack :
-            a Stack object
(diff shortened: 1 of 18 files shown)


ttung:
Loading the DARTFISH data:

```
In [1]: from starfish.experiment import Experiment

In [2]: a= Experiment.from_json('https://dmf0bdeheu4zf.cloudfront.net/20180821/DARTFISH/fov_001/experiment.json')

In [3]: a.fovs()[0]
Out[3]: <starfish.experiment.FieldOfView at 0x121f634e0>

In [4]: a.fovs()[0].primary_image
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 18/18 [00:00<00:00, 100.58it/s]
Out[4]: <starfish.stack.ImageStack at 0x12813e6d8>

In [5]: a.fovs()[0].auxiliary_image_types
Out[5]: dict_keys(['nuclei'])

In [6]: a.fovs()[0]['nuclei']
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 44.15it/s]
Out[8]: <starfish.stack.ImageStack at 0x10f4c1fd0>

In [6]: 
```

ttung:
Notice that all the actual image data is lazy-loaded, such that if you only operate on one FOV, none of the other tiles get loaded!

ambrosejcarr:
I've started by going through the examples you've given (very helpful!) without looking at the code to try to get an unbiased perspective on how I as a user would interact with the API. I've listed a few suggestions based on this below. 

### Make experiment dict-like
```
In [4]: a.fov_by_name('fov_001')
Out[4]: [<starfish.experiment.FieldOfView at 0x1153734a8>]
```
I think we should require unique names, and this should return a FieldOfView, not a list of them. Am I missing something? 

If that works, Is there a reason that we don't overload `__getitem__` to directly do this? Then you could just go experiment.keys() to get the field of view names. This would feel pretty natural to python programmers, I think. You could leave the `fov_by_name` accessor there and delegate the code to `__getitem__`. `__iter__` and `.items()` could provide the iterators needed to sequentially process a field of view. 

### Properties for image types on `FieldOfView`
```
In [7]: a.fov_by_name('fov_001')[0]['nuclei']
100%|█████████| 1/1 [00:00<00:00, 31.20it/s]
Out[7]: <starfish.stack.ImageStack at 0x12750f240>
```
It looks like you've got `FieldOfView` using `__getitem__ for `auxiliary_images`. Given that we have controlled naming for the auxilary images from the spec, I think I would instead expose a property for `nuclei`, `dots`, and `primary_image` so they're all equal. 

### Miscellaneous questions: 
- Can we expose extras as a dict? 
- How does serialization of pipeline-derived changes to an ImageStack work (1) in-place and (2) out-of-place? 

### General Comments
✔️ love the lazy loading
✔️ love the codebook accessor

ttung:
> I think we should require unique names, and this should return a FieldOfView, not a list of them. Am I missing something?

> If that works, Is there a reason that we don't overload __getitem__ to directly do this? Then you could just go experiment.keys() to get the field of view names. This would feel pretty natural to python programmers, I think. You could leave the fov_by_name accessor there and delegate the code to __getitem__. __iter__ and .items() could provide the iterators needed to sequentially process a field of view.

So after I wrote those comments, I added `__getitem__` to Experiment, which gets a single FOV.  `fov_by_name` was renamed to `fovs_by_name`, to be explicit about what it does.  `fovs_by_name` is an attempt to address a common use case (extracting one or more FOVs).

> It looks like you've got FieldOfView using __getitem__ for auxiliary_images. Given that we have controlled naming for the auxilary images from the spec, I think I would instead expose a property for nuclei, dots, and primary_image so they're all equal.

Only nuclei looks like it's required according to the spec (looks like @dganguli added that Jul 12).
BTW, if we are going to make any of them required, they do _not_ belong in auxiliary images, and we should really promote them to the top level of `experiment.json`.

ttung:
> Can we expose extras as a dict?

Does anyone use extras?  I'm happy to expose it as `Experiment.extras`, but am reluctant to expose it at a higher level until we know what it's going to use it for.

> How does serialization of pipeline-derived changes to an ImageStack work (1) in-place and (2) out-of-place?

Don't understand this question.  :(

ambrosejcarr:
> > How does serialization of pipeline-derived changes to an ImageStack work (1) in-place and (2) out-of-place?

> Don't understand this question. :(

Say I do some things to a field of view in an experiment.

How do I (1) save a new experiment that contains my changes and (2) overwrite the experiment I'm working with?

ambrosejcarr:
> Only nuclei looks like it's required according to the spec (looks like @dganguli added that Jul 12).
BTW, if we are going to make any of them required, they do not belong in auxiliary images, and we should really promote them to the top level of experiment.json.

Sorry, I wasn't suggesting they be required, just that there be a property for them. It could return `None` if the experiment doesn't use it.

starfish/experiment/__init__.py
```diff
@@ -1,93 +1,172 @@
...
+            auxiliary_images[aux_image_type] = Reader.parse_doc(aux_image_url, baseurl)
+
+        if isinstance(hybridization_image, TileSet):
+            # make sure that all the aux images are also TileSets
+            for aux_image_tileset in auxiliary_images.values():
+                assert isinstance(aux_image_tileset, TileSet)
+
+            fov = FieldOfView(
+                None,
+                primary_image_tileset=hybridization_image,
```
dganguli: codebase is starting to get weird about primary_image v.s hybridization_images. since the PR in spaceTx fmt to rename hybridization_images->primary_images, i think all new code should start using this language?

starfish/experiment/__init__.py
```diff
@@ -1,93 +1,172 @@
...
         if not (Experiment.MIN_SUPPORTED_VERSION <= version <= Experiment.MAX_SUPPORTED_VERSION):
             raise ValueError(
                 f"version {version} not supported.  This version of the starfish library only "
                 f"supports formats from {Experiment.MIN_SUPPORTED_VERSION} to "
                 f"{Experiment.MAX_SUPPORTED_VERSION}")
 
-    @classmethod
-    def from_json(cls, json_url: str) -> "Experiment":
-        """Construct an `Experiment` from an experiment.json file format specifier
+    def fovs(self, filter_fn=lambda _: True) -> Sequence[FieldOfView]:
```
dganguli: may be useful to sort list by fov_name?

dganguli:
Experiment.extras will be useful for many reasons. Right now, we need it for MERFISH to store global scale factors that apply to each round/color channel for the entire experiment. We should keep this.

dganguli:
"Say I do some things to a field of view in an experiment.

How do I (1) save a new experiment that contains my changes and (2) overwrite the experiment I'm working with?"

Good question! What do we want this behavior to be?

dganguli:
"Only nuclei looks like it's required according to the spec (looks like @dganguli added that Jul 12).
BTW, if we are going to make any of them required, they do not belong in auxiliary images, and we should really promote them to the top level of experiment.json."

This is a nice thing to do -- but I don't feel strongly about it. I guess if we have changes to the format, we should do 'em sooner than later? I vote whoever feels strongest pushes a PR to spaceTx format and changes the Starfish code after acceptance!

ttung:
> Say I do some things to a field of view in an experiment.

> How do I (1) save a new experiment that contains my changes and (2) overwrite the experiment I'm working with?

One can already save ImageStacks, but is there a compelling reason to construct a new top-level experiment object?

ttung:
> Sorry, I wasn't suggesting they be required, just that there be a property for them. It could return None if the experiment doesn't use it.

I would avoid doing so just because that adds constraints on the names of the FOVs.  i.e., `fovs` would be a valid fov name, but `experiment.fovs` would _not_ return you the fov that is called `fovs`.

starfish/experiment/__init__.py
ttung: Sure, I will make this change.

ambrosejcarr:
I'm thinking about a use case where you want to save out intermediate filtered images after running through a subset of a pipeline. 

One might want to restart from registered+deconvolved images across fields of view, for example. A new experiment seems like the right way to save out + restart, to me.

ttung:
> I'm thinking about a use case where you want to save out intermediate filtered images after running through a subset of a pipeline.

> One might want to restart from registered+deconvolved images across fields of view, for example. A new experiment seems like the right way to save out + restart, to me.

We can certainly add code to do this, but it's almost certainly the wrong model of execution and we should strongly discourage this.

People should be attempting to compose the right recipe for their pipeline using a small number of fields of views, and then applying that recipe to the entire experiment.  And they can save intermediate results for a small number of field of views with ease.  They should **not** be doing partial processing of a pipeline of the entire experiment's data.

ambrosejcarr:
I get where you're coming from with this, but to me science has almost never been this clean. I haven't done that much imaging, but my experience with large-scale science has always been: 

1. pilot on something small
2. do everything all at once
3. realize something weird is going on
4. go back to 2 and figure out where the problem is, iterate on 2-4. 

To do 3 and 4, I often need to save out intermediates while I'm iterating to save time and money. Willing to defer to others with more experience on this -- would like to hear what @dganguli and @berl think here.

ambrosejcarr:
I think it depends on how much time/money the initial steps take, and how linear the process is (i.e. can you trust that a step won't have an interaction with something downstream, and can really be "finalized" before the whole pipeline is done). If you can't finalize intermediate steps, then my approach is without merit. 

For me most of the time I find more than one problem. So, I do as you say -- benchmark on the problem example, fix it, expand back out to dataset. If new problems, find the next problem, benchmark on the example, and expand back out. 

For me, things took long enough to process that I never wanted to wait for the whole processing tree to run. When we start introducing stitching and other components that reduce over multiple fields of view I could see my approach having more utility. If everything is as fast as it is now (minutes to process) then maybe it's not important. 

In either case, I'm not interested in forcing this issue -- I wanna hear the scientists weigh in. :-)

ambrosejcarr:
Um. But, none of this should stop you from fleshing out the MVP as you've made it.

ttung:
> Um. But, none of this should stop you from fleshing out the MVP as you've made it, which is awesome.

Yup, agreed. Working on that already.

berl:
first of all, @ttung  this is great.  

comments.
1. I don't understand the `primary_image` vs. `primary_image_tileset` thing- they're both `ImageStack`s and you can only have one, so why are there 2 different things?
2. I agree with @ambrosejcarr that you should move forward with this before addressing the workflow issue.
3. the workflow issue:  I would love to have an experiment class that would allow *very* flexible processing like the use-case that @ambrosejcarr mentioned, but default to a more sane workflow like @ttung asserted.  I've made an issue/manifesto #487 on the topic.

ttung:
> I don't understand the primary_image vs. primary_image_tileset thing- they're both ImageStacks and you can only have one, so why are there 2 different things?

primary_image_tileset is a TileSet. TileSets are more flexible and are demand-loaded. Rationale: one can instantiate all the TileSets very quickly without computational or network costs. ImageStacks are loaded data.

ambrosejcarr:
> primary_image_tileset is a TileSet. TileSets are more flexible and are demand-loaded. Rationale: one can instantiate all the TileSets very quickly without computational or network costs. ImageStacks are loaded data.

I think I echo @berl's concerns here -- I feel like all the Tileset / ImageStack stuff should really be one class, where the ImageStack just doesn't load immediately. Is there a reason this couldn't happen? 

Also, I love the idea of FieldOfView -- can you comment on what methods you see moving there from ImageStack, if anything? 

Finally, what happens to experiment when you've called spots? Does a FieldOfView now point at an IntensityTable in addition to a Primary and Auxiliary ImageStack? (I think it probably should?)

ttung:
> I think I echo @berl's concerns here -- I feel like all the Tileset / ImageStack stuff should really be one class, where the ImageStack just doesn't load immediately. Is there a reason this couldn't happen?

We can do that; it does require a potentially significant redesign of ImageStack though.

ambrosejcarr:
How long do you think it would take? I'm tentatively open to a sequential approach, but I don't see the value of the extra classes long term, and I think a user could find them confusing -- at the very least the complexity should be hidden from a user. 

Interested in what @dganguli thinks about this, and about the need to get something in immediately.

PR #483, opened 2018-08-25
ENH: add multiprocessing

ambrosejcarr:
Turns out this works as a drop-in replacement. I was just missing the `list()` call on. [Line 622](https://github.com/spacetx/starfish/compare/ajc-multiprocessing?expand=1#diff-58c5e4869c4fa20d3233417448d9acbbR662)

Do we want to add multiprocessing? This makes deconvolving way faster (17s for the whole thing, instead of 15s per iteration * 16 iterations), and cuts 25s off our test suite.

### DIFF
--- a/starfish/stack.py
+++ b/starfish/stack.py
@@ -4,6 +4,7 @@ import warnings
 from copy import deepcopy
 from functools import partial
 from itertools import product
+from multiprocessing import Pool
 from typing import (
     Any, Callable, Iterable, Iterator, List, Mapping, MutableSequence, Optional, Sequence, Tuple,
     Union
@@ -608,7 +609,8 @@ class ImageStack:
             yield array
 
     def apply(
-            self, func, is_volume=False, in_place=True, verbose: bool=False, **kwargs
+            self, func, is_volume=False, in_place=True, verbose: bool=False, multiprocess=True,
+            **kwargs
     ) -> "ImageStack":
         """Apply func over all tiles or volumes in self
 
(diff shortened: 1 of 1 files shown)


berl:
For people running this on a workstation it might be nice to have a keyword to set the number of workers in the pool. Without specifying it in the Pool() constructor, it defaults to  cpu_count() which is great if you don't want to do anything else while it's running.

ambrosejcarr:
Yep, that's a good point. I think I'd also probably set the default to 1, which would turn it off. Mostly interested how this interacts with tony's dreams for scale. It could be that this is disruptive. :-)

Thanks for the reminder about the pool size!

ambrosejcarr:
@joshmoore That's a good idea for a super-user, but I don't think our users will have the required sophistication.

starfish/stack.py
```diff
@@ -608,7 +609,8 @@ def _iter_tiles(
             yield array
 
     def apply(
-            self, func, is_volume=False, in_place=True, verbose: bool=False, **kwargs
+            self, func, is_volume=False, in_place=True, verbose: bool=False, multiprocess=True,
```
ttung: so rather than say multiprocess=True (or False), perhaps you should do: max_workers=None, where None=fork off as many as you have cores.

starfish/stack.py
```diff
@@ -641,23 +645,32 @@ def apply(
                 func, is_volume=is_volume, in_place=True, verbose=verbose, **kwargs
             )
 
-        mapfunc: Callable = map  # TODO: ambrosejcarr posix-compliant multiprocessing
+        if multiprocess:
```
ttung:
```
with ThreadPoolExecutor(max_workers=max_workers) as executor:
    ...
```

ambrosejcarr:
I'm going to merge this into the existing apply PR, will use @ttung's suggestions.

ambrosejcarr:
Closed, implementing in #420.

PR #491, opened 2018-08-29
Add basic matplotlib visualization suite. 

ambrosejcarr:
- Add histogram visualizations
- Add spots visualization
- Add copy number comparison visualization
- Add some basic plotting constants and utility functions
- Remove broken show spots
- Add notebook comparing ISS, DARTFISH, MERFISH.


REQUIREMENTS.txt
```diff
@@ -9,6 +9,7 @@ semantic_version
 scikit-image>=0.14.0
 scikit-learn
 scipy
+seaborn
```
dganguli: add to requirements-notebook?

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+
+# EPY: START markdown
+### Load IntensityTables
+# EPY: END markdown
+
+# EPY: START code
+# IntensityTable can't download from directories without list privileges
+
+data_root = "https://dmf0bdeheu4zf.cloudfront.net/assay_comparison/"
+iss_link = data_root + "iss.nc"
```
dganguli: @ttung and marcus always make me do os.path.join(dir, stuff, stuff2, ... stuffN). i noticed that that's way better than using + as it handles slashes nicely

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+### Load Background Images
+# EPY: END markdown
+
+# EPY: START code
+# construct background images for each assay
+stack = Experiment.from_json('https://dmf0bdeheu4zf.cloudfront.net/20180821/DARTFISH/fov_001/experiment.json')
+
+dartfish_nuclei = stack.auxiliary_images['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
+dartfish_link = data_root + "dartfish_dots_image.npy"
+dartfish_npy = tmp + "dartfish.npy"
```
dganguli: how was this dos_image created? is it the 'decoded_image' from the dartfish notebook?

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+iss_nuclei = stack.auxiliary_images['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
+iss_dots = stack.auxiliary_images['dots'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
+# EPY: END code
+
+# EPY: START code
+stack = Experiment.from_json('https://dmf0bdeheu4zf.cloudfront.net/20180802/MERFISH/fov_001/experiment.json')
+merfish_nuclei = stack.auxiliary_images['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
+
+# merfish doesn't have a dots image, and some of the channels are stronger than others. We can use the scale factors to get the right levels
+merfish_background = stack.image.max_proj(Indices.CH, Indices.ROUND)
```
dganguli: it really would be nice for max_proj to return an ImageStack right! That would obviate all this code, which I also frequently find myself writing.

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+### Show different background types for ISS
+# EPY: END markdown
+
+# EPY: START code
+f, ((ax1, ax2), (ax3, ax4), (ax5, ax6)) = plt.subplots(ncols=2, nrows=3, figsize=(30, 40))
+spots(iss, background_image=np.zeros_like(iss_dots), spots_kwargs=dict(alpha=1.), ax=ax1)
+spots(iss, background_image=iss_dots, spots_kwargs=dict(alpha=1.), ax=ax3)
+spots(iss, background_image=iss_nuclei, spots_kwargs=dict(alpha=1.), ax=ax5)
+
+# ISS doesn't have a decoded image, so leave those plots blank to make this clear
```
dganguli: ISS can have a decoded image though -- if we run a version of the pipeline that uses pixel detection instead of spot detection. we should do that, it will be interesting to comare the threeway of 1. benchmark 2. iss spot detection 3. iss pixel detection. it will also complete the missing column of this plot, thus showing off the flexibility of the code. although this could be it's own separate plot as well w/ 2 or 3 copy number comparison plots: bench vs. spots, bench vs. pixels, spots vs. pixels

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+    ax.set_axis_off()
+f.tight_layout()
+# EPY: END code
+
+# EPY: START markdown
+### Download available copy number information from original authors
+# EPY: END markdown
+
+# EPY: START code
+dartfish_copy_number = pd.read_csv('https://dmf0bdeheu4zf.cloudfront.net/20180821/DARTFISH/fov_001/counts.csv', index_col=0, squeeze=True)
```
dganguli: wow. wher has squeeze=True been my whole life!

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+iss_copy_number = pd.read_csv(data_root + "iss_copy_number_benchmark.csv", index_col=1)['cnt']
+# EPY: END code
+
+# EPY: START code
+f, axes = plt.subplots(ncols=3, nrows=4, figsize=(12, 14))
+
+for dataset, benchmark, axis in zip([iss, merfish, dartfish], [iss_copy_number, merfish_copy_number, dartfish_copy_number], axes[0, :]):
+    compare_copy_number(dataset, benchmark, ax=axis, color='tab:blue')
+
+for dataset, axis in zip(datasets, axes[1, :]):
```
dganguli: these three plots (magnitude, area, distance) should have a parameter for 'threshold' which indicates the threshold that was picked as input to the SpotDetectors. The plot should show a prominent vertical line at that threshold on top of the histogram.

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+    barcode_magnitude_histogram(dataset, bins=20, ax=axis)
```
dganguli: for ISS, the y-axis should read "Number of Spots". Alternatively, each plot could have y-axis "Number of Features" as long as it's clear somewhere that features can either be barcode values for a single pixel, or some reduced set of barcode values (e.g., mean, max, etc.) for all pixels in a spot.

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+
+for dataset, axis in zip(datasets, axes[2, :]):
+    spot_area_histogram(dataset, bins=20, ax=axis)
+
+for dataset, axis in zip([merfish, dartfish], axes[3, 1:]):
```
dganguli: we should be able to plot the spot distance histogram for iss. all the information to compute this metric is in the intensity table. as such, we should plot it. we will also learn something about our API in constructuing this figure. 

additionally, in the olden days, i had a quality metric for iss spot detection but i think it was lost. it was a good metric for the per_round_max_channel_decoder. was this code nuked? if so, why?

dganguli: also, it's probably worth generalizing spot_distance_histogram to spot_quality_histogram. for some assays (e.g., pixel read out ones) the quality will be distance. for others, it will be some other quality score thats particular to the decoder and/or chemistry (e.g., the old iss quality metric)

starfish/plot/intensity_table.py
```diff
@@ -0,0 +1,182 @@
...
+import seaborn as sns
+from scipy.stats import pearsonr
+
+from starfish.stats import feature_trace_magnitudes
+from starfish.types import Features
+from .util import annotate_axis
+
+
+def spot_area_histogram(
+        intensities,
```
dganguli: add typing? Union(IntensityTable, np.Array) or something like that?

starfish/plot/intensity_table.py
```diff
@@ -0,0 +1,182 @@
+import warnings
```
dganguli: i find it confusing to have two intensity_table.pys in the project, even if one is in starfish.plot and the other is in starfish. is there a better name for this file? can we move the function into plots.spots.py?

starfish/plot/intensity_table.py
```diff
@@ -0,0 +1,182 @@
...
+    intensities
+
+    Returns
+    -------
+    matplotlib.axes.Axes :
+        The axis containing the plot
+
+    """
+    ax = ax if ax is not None else plt.gca()
+    ax.hist(intensities[Features.DISTANCE].values, bins=bins, **kwargs)
```
dganguli: Would be interesting if this were parameterizable so that we could also pass in Features.QUALITY in addition to Features.DISTANCE?

starfish/plot/intensity_table.py
dganguli: same comment applies to all other functions in this file

starfish/plot/spots.py
```diff
@@ -0,0 +1,138 @@
...
+
+from starfish.types import Features
+from .style import BACKGROUND_COLORMAP, TARGETS_COLORMAP
+from .util import annotate_axis, dpi_correction, remove_axis_frame, set_default_arguments
+
+
+# TODO use the name of the IntensityTable if it exists for a title
+# TODO ambrosejcarr: currently the colormaps for decoded_image and intensities do not align.
+# however, it should be possible to fix that.
+def spots(
```
dganguli: rename to plot_spots, show_spots, plot_spots_on_background? something more expressive than spots would be nice -- but i'll leave it to you to name it.

starfish/plot/spots.py
```diff
@@ -0,0 +1,138 @@
...
+            decoded_image_kwargs,
+            alpha=0.5,
+            cmap=TARGETS_COLORMAP
+        )
+
+        masked_image = np.ma.masked_equal(decoded_image, 0)
+        ax.imshow(masked_image, interpolation='none', **decoded_image_kwargs)
+
+    remove_axis_frame(ax)
+    annotate_axis(ax, title='Decoded Image')
```
dganguli: make title an optional argument to the function?

starfish/plot/util.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Returns
+    -------
+    float :
+        multiplier for matplotlib.plot.scatter markersize kwarg to ensure correct size in pixels
+
+    """
+    if figure is None:
+        figure = plt.gcf()
+    point_multiplier = 72. / figure.dpi
```
dganguli: where does the 72 come from?

starfish/stack.py
```diff
@@ -489,57 +468,6 @@ def display_slice(plane_index=0):
 
         return display_slice
 
-    # TODO ambrosejcarr: update to use IntensityTable instead of SpotAttributes
```
dganguli: this must have felt so good to nuke!

notebooks/py/assay_comparison.py
ambrosejcarr: Yep. Sure would. :) On my list for "things to do after floats"

notebooks/py/assay_comparison.py
ambrosejcarr: I wonder if we just want to generalize "histogram" to have some basic style parameters and an optional threshold.

starfish/plot/util.py
ambrosejcarr: This is apparently the default scaling parameter between pixels and DPI in matplotlib. I will add a comment. Thanks for the catch.

starfish/stack.py
ambrosejcarr: Oh did it ever.

starfish/stats.py
```diff
@@ -18,3 +21,10 @@ def region_for(label_mat_coo, label):
     regions = [region_for(label_mat_coo, label) for label in unique_labels]
 
     return regional.many(regions)
+
+
+def feature_trace_magnitudes(intensities, norm_order=2) -> np.ndarray:
```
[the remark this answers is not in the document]
ambrosejcarr: Yep, I can put this elsewhere.

starfish/plot/intensity_table.py
```diff
@@ -0,0 +1,182 @@
...
+    targets, counts = np.unique(intensities[Features.TARGET].values, return_counts=True)
+    this = pd.Series(counts, index=targets)
+    tmp = pd.concat([this, other], join='outer', axis=1, sort=True)
+    tmp.columns = ['result', 'comparison']
+
+    # this seaborn function does not accept an axis, but we can make sure it draws to the right
+    # place by setting the current axis
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', FutureWarning)
+        sns.regplot('comparison', 'result', data=tmp, ax=ax, **kwargs)
```
kevinyamauchi: This is a nitpick, but is seaborn necessary here? I think this is the only plot using seaborn, so it may be unnecessary to add an additional dependency for just this plot. I think it shortcuts a bunch of work, so it's probably worth it.

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,237 @@
...
+    ax.set_ylabel(name, fontsize=16)
+
+
+# fix up the figure
+f.tight_layout()
+for ax in np.ravel(axes):
+    sns.despine(ax=ax)
+
+# turn off the empty plots, matplotlib is super bad about this. 
+for ax in (axes[3, 0], axes[0, 0]):
```
kevinyamauchi: Why is there a plot, but no axes for the Copy number comparison - ISS plot? Maybe the seaborn plot on the ax isn't a patch and thus cannot be toggled off with ax.patch.set_visible()?

REQUIREMENTS.txt
ambrosejcarr: Oh I thumbs-upped too soon. I think it needs to be in dev because starfish will die if this doesn't get installed, since it gets imported in the vis suite.

notebooks/py/assay_comparison.py
ambrosejcarr: Max proj over hyb + channels.

notebooks/py/assay_comparison.py
ambrosejcarr: > we should be able to plot the spot distance histogram for iss. all the information to compute this metric is in the intensity table. as such, we should plot it. we will also learn something about our API in constructuing this figure.

Actually, master does _not_ have the required code. My filtering branch rectifies this by reporting the fraction of the signal for each round that resides in the selected channel, however currently the distance is 1 if the spot fails to decode, and zero otherwise. 

We should circle back and update this when that fix is made.

notebooks/py/assay_comparison.py
ambrosejcarr: Oops. I originally didn't have a copy number plot, and I forgot to remove this. Nice catch!

starfish/plot/intensity_table.py
ambrosejcarr: >  I think it shortcuts a bunch of work, so it's probably worth it.

Indeed it does. I think it's worth a dependency. If we're adding more plots, eventually we'll use things from it.

ambrosejcarr:
@kevinyamauchi @dganguli I'm going to merge this. Please let me know if you have any further changes you'd like, I'm happy to make follow-up PRs.

REQUIREMENTS.txt
ttung: What's the rule for something that belongs in requirements-notebook?

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,354 @@
...
+# EPY: END code
+
+# EPY: START markdown
+### Load IntensityTables
+# EPY: END markdown
+
+# EPY: START code
+# IntensityTable can't download from directories without list privileges
+
+data_root = "https://dmf0bdeheu4zf.cloudfront.net/assay_comparison/"
```
ttung: not versioned!!

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,354 @@
...
+iss_link = os.path.join(data_root, "iss.nc")
+merfish_link = os.path.join(data_root, "merfish.nc")
+dartfish_link = os.path.join(data_root, "dartfish.nc")
+
+tmp = os.environ['TMPDIR']
```
ttung: `tempfile.gettempdir()`?

notebooks/py/assay_comparison.py
```diff
@@ -0,0 +1,354 @@
...
+### Load Background Images
+# EPY: END markdown
+
+# EPY: START code
+# construct background images for each assay
+stack = Experiment.from_json(
+    'https://dmf0bdeheu4zf.cloudfront.net/20180821/DARTFISH/fov_001/experiment.json'
+)
+
+dartfish_nuclei = stack.auxiliary_images['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
```
ttung: rebase!

this will likely break horribly.

PR #472, opened 2018-08-23
Migrate docker to conda

joshmoore:
Since the other .travis.yml builds test the REQUIREMENTS*
files, this now tests an environment.yml in docker. There's
an outstanding issue with slicedimage and its enum34 dep
which will need to be fixed before this is mergeable.

see https://github.com/spacetx/starfish/issues/44

Issue #44, opened 2018-02-05
Verify conda compatibility


ttung:

joshmoore:
I've pushed my branch to https://github.com/joshmoore/bioconda-recipes/tree/starfish. Happy to keep it up-to-date but would either need a heads up in Slack or tag for each of https://github.com/spacetx/starfish/commits/master/setup.py so that I can follow the RSS feed.

joshmoore:
https://github.com/spacetx/starfish/releases/tag/0.0.10 cheers. Now following https://github.com/spacetx/starfish/releases.atom

joshmoore:
Hit the first blocker. All prereqs must also be available in bioconda, so I started packaging showit, slicedimage and regional. For slicedimage:

```
Error: bad character '-' in package/version: 0.0.24-dev
```

Thoughts on having a non `-dev` version?

ambrosejcarr:
You know, I've seen lots of `0.0.24rc2` and similar. Perhaps this is why? Should we try `0.0.24dev` first?

ttung:
I think we should somehow get travis testing w/ conda.  Maybe set up an environment, install starfish, and test that we can at least invoke the cli?

joshmoore:
> You know, I've seen lots of 0.0.24rc2 and similar. Perhaps this is why? Should we try 0.0.24dev first?

https://www.python.org/dev/peps/pep-0440/ suggests:

```
[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

> I think we should somehow get travis testing w/ conda. Maybe set up an environment, install starfish, and test that we can at least invoke the cli?

I'd suggest step one would be an environment.yml file, which can also download from pip etc. Then either travis could consume that or it could be consumed in the Dockerfile or another Dockerfile.

My goal with the bioconda recipces would then be to reduce the need for the environment file, i.e. `conda install -c bioconda starfish`.

joshmoore:
```
$ cat starfish.yml
name: starfish
channels:
- defaults
- bioconda
- conda-forge
dependencies:
- click
- matplotlib
- numpy!=1.13.0
- pandas>=0.23.4
- pytest>=3.6.3
- requests
- semantic_version
- scikit-image>=0.14.0
- scikit-learn
- scipy
- tqdm
- trackpy
- validators
- xarray
- pip:
  - regional
  - slicedimage==0.0.1
  - showit==1.1.4
$ conda env create -f starfish.yml
...
[...]

PR #495, opened 2018-08-31
Conda docker

### LINKED ISSUE #511: docker / conda files from sptx-format need to be integrated
attn @joshmoore; I think you added these files.  https://github.com/spacetx/starfish/tree/dd9990970e75218ee5afe02f894cc39eda99edca/sptx-format

joshmoore:
Since the other .travis.yml builds test the REQUIREMENTS*
files, this now tests an environment.yml in docker. There's
an outstanding issue with slicedimage and its enum34 dep
which will need to be fixed before this is mergeable.

see #44
close #511
(Was previously opened as #472)


joshmoore:
A new PR & a force push is _still_ not triggering...

Issue #380, opened 2018-08-08
Reconcile blob_log and trackpy.locate radii


ambrosejcarr:
- Currently these two approaches generate different outputs -- trackpy locate generates radius of gyration, which is a function of the intensity, while blob_log emits the standard deviation. 
- We want to translate these into a common concept so that we can standardize the intensitytable
- As a bonus, numpy has exclusive indexing, so simply adding and subtracting the bounding box as we do in measure is wrong. 
- Finally, we should not even use a box, but rather should create a circle/sphere to cut out parts that should not be bright. 

All of this contributes to "mean" being an extremely fragile way of measuring a spot.

ambrosejcarr:
The question of Pixel radius was also raised here, dealing with whether it is more important to get the bounding box or the area correct. Essentially, pixels are square, and so there is an inconsistency. 

https://github.com/spacetx/starfish/pull/496/files#r215738317

PR #496, opened 2018-08-31
Refactor Tests for Codebook and IntensityTable

ambrosejcarr:
Summary of changes:

`Codebook` 
1. Updated to allow features that previously were filtered out to be stored in the IntensityTable. IntensityTable gains an additional data variable `passes_filters` that records whether or not each feature passes filters. 
2. Updated `per_round_max_channel_decode` to store a proper distance instead of a binary one where decoded=1, failed to decode=0. Now stores `mean_over_rounds(1 - (max_channel / sum(channels)))`
3. Side-effect: `starfish.types.Features` gains `PASSES_FILTERS='passes_filters'`

`IntensityTable`
1. Take SpotAttributes object as input, not a pd.DataFrame _called_ spot_attributes (enforces parameter checking). 
2. Side-effect: updated `SpotAttributes` to _not_ require `intensity` and `radius` as those are not always available. 
3. Side-effect: gaussian_spot_detection, local_max_peak_finder, and pixel_spot_detector all undergo minor changes to support `SpotAttributes` instead of `pd.DataFrame`

`local_max_peak_finder`
1. Squelched a numpy FutureWarning. 

`combine_adjacent_features`
1. Refactored this into a class `CombineAdjacentFeatures` and broke code into sub-methods to make more testable (output is identical). 

```diff
+ 26 files worth of tests
- original intensity_table and codebook tests. 
```

### Of note: 
All API tests pass, including tests of each data type that mimic the notebooks (ISS, DARTFISH, MERFISH). The only changes to the notebook subset tests that were necessary was to subset the intensity_tables using the new `passes_filter` fields to restrict the tests to counting features that pass filters.


starfish/codebook.py
```diff
@@ -442,40 +456,45 @@ def metric_decode(
...
 
-        # mask low intensity features
-        intensity_mask = np.where(norms < min_intensity)[0]
-
         metric_outputs, targets = self._approximate_nearest_code(
             norm_codes, norm_intensities, metric=metric)
 
-        exceeds_distance = np.where(metric_outputs > max_distance)[0]
-        target_index = pd.Index(targets)
+        passes_filters = np.ones_like(targets, dtype=np.bool)
```
ttung: does it make sense to fill it in one pass?

like `passes_filters = logical_or(norms > min_intensity, metric_outputs < max_distance)`

personally, i feel this makes more sense.  i don't know if it actually works, because i suck at numpy, but I feel it should.

starfish/codebook.py
```diff
@@ -442,40 +456,45 @@ def metric_decode(
...
-        intensities[Features.TARGET] = (Features.AXIS, target_index)
-        intensities[Features.DISTANCE] = (Features.AXIS, metric_outputs)
+        # set targets, distances, and filtering results
+        norm_intensities[Features.TARGET] = (Features.AXIS, targets)
+        norm_intensities[Features.DISTANCE] = (Features.AXIS, metric_outputs)
+        norm_intensities[Features.PASSES_FILTERS] = (Features.AXIS, passes_filters)
 
-        return intensities
+        # norm_intensities is a DataArray, make it back into an IntensityTable
+        return IntensityTable(norm_intensities)
```
ttung: any perf implications for this?

notebooks/py/DARTFISH_Pipeline_-_Human_Occipital_Cortex_-_1_FOV.py
```diff
@@ -225,6 +225,18 @@ def compute_magnitudes(stack, norm_order=2):
 spot_intensities, results = psd.run(zero_norm_stack)
 # EPY: END code
 
+# EPY: START code
+np.unique(spot_intensities['passes_filters'], return_counts=True)
```
dganguli: i'm getting confused between 'filter' -- for image processing and 'filter' -- for annotating rows in an intensity table that meet certain criteria. is there a better name for filter in the latter case? e.g., passes_criteria, passes_thresholds, passes_*?

starfish/codebook.py
ttung: test if mutating one data set changes the other.

starfish/codebook.py
```diff
@@ -513,24 +532,33 @@ def _view_row_as_element(array: np.ndarray) -> np.ndarray:
...
-        for i in np.arange(a.shape[0]):
+        # decode the intensities
+        for i in np.arange(codes.shape[0]):
             targets[np.where(a[i] == b)[0]] = codes[Features.TARGET][i]
-        target_index = pd.Index(targets.astype('U'))
 
-        intensities[Features.TARGET] = (
-            Features.AXIS, target_index)
+        # a code passes filters if it decodes successfully
+        passes_filters = np.ones_like(targets, dtype=np.bool)
```
ttung: two passes again.

starfish/intensity_table.py
```diff
@@ -88,10 +90,12 @@ def empty_intensity_table(
             empty IntensityTable
 
         """
-        cls._verify_spot_attributes(spot_attributes)
+        if not isinstance(spot_attributes, SpotAttributes):
```
ttung: Unclear why we validate the data type here where we don't really elsewhere.

starfish/codebook.py
```diff
@@ -513,24 +532,33 @@ def _view_row_as_element(array: np.ndarray) -> np.ndarray:
...
+        self._validate_decode_intensity_input_matches_codebook_shape(intensities)
+
         max_channels = intensities.argmax(Indices.CH.value)
         codes = self.argmax(Indices.CH.value)
 
+        # calculate distance scores by evaluating the fraction of signal in each round that is
+        # found in the non-maximal channels.
+        max_intensities = intensities.max(Indices.CH.value)
+        round_intensities = intensities.sum(Indices.CH.value)
+        distance = 1 - (max_intensities / round_intensities).mean(Indices.ROUND.value)
```
dganguli: this quality score is interesting -- but not an established one. perhaps leave a TODO to further explore this? and an issue

starfish/codebook.py
```diff
@@ -513,24 +532,33 @@ def _view_row_as_element(array: np.ndarray) -> np.ndarray:
...
 
-        intensities[Features.TARGET] = (
-            Features.AXIS, target_index)
+        # a code passes filters if it decodes successfully
+        passes_filters = np.ones_like(targets, dtype=np.bool)
+        passes_filters[pd.isnull(targets)] = False
+
+        intensities[Features.TARGET] = (Features.AXIS, targets.astype('U'))
+        intensities[Features.DISTANCE] = (Features.AXIS, distance)
+        intensities[Features.PASSES_FILTERS] = (Features.AXIS, passes_filters)
```
dganguli: right now, we filter (hopefully to rename to threshold, see comment above) decoded features on area and intensity. we should ALSO filter on distance/quality. Perhaps submit an Issue for this?

starfish/test/codebook/test_from_json.py
```diff
@@ -0,0 +1,108 @@
...
+    ]
+    return codebook_data
+
+
+def test_codebook_loads_from_local_file() -> None:
+    """dumps the codebook data to a temporary json file and reads it back into a Codebook."""
+
+    # dump codebook to disk
+    codebook_data: List = codebook_json_data_factory()
+    directory = tempfile.mkdtemp()
```
ttung: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

starfish/test/codebook/test_from_json.py
```diff
@@ -0,0 +1,108 @@
...
+    codebook = Codebook.from_json(codebook_json)
+    assert codebook.sizes[Indices.ROUND] == 2
+    assert codebook.sizes[Indices.CH] == 3
+    assert codebook.sizes[Features.TARGET] == 2
+
+    # clean up
+    shutil.rmtree(directory)
+
+
+@pytest.mark.skip('Tested elsewhere now that we use slicedimage backend. TODO Confirm and delete.')
```
ttung: nuke

nuclear-explosion3

starfish/intensity_table.py
```diff
@@ -307,32 +298,15 @@ def from_image_stack(
         y = np.arange(ymin, ymax)
         x = np.arange(xmin, xmax)
 
-        pixel_coordinates = pd.DataFrame(
+        attribute_data = pd.DataFrame(
```
dganguli: rename to feature_attribute_data?

starfish/intensity_table.py
```diff
@@ -307,32 +298,15 @@ def from_image_stack(
...
             data=np.array(list(product(z, y, x))),
             columns=['z', 'y', 'x']
         )
-        pixel_coordinates[Features.SPOT_RADIUS] = np.full(
-            pixel_coordinates.shape[0], fill_value=np.nan)
-
-        image_size = cropped_data.shape[:3]
-
-        return IntensityTable.from_spot_data(intensity_data, pixel_coordinates, image_size)
+        attribute_data[Features.SPOT_RADIUS] = np.full(attribute_data.shape[0], fill_value=0.5)
```
ambrosejcarr: Leave a comment about these == pixels, therefore diameter should be easily calculated from the diameter. Think about area of 1 pixel and how to calculate this.

ambrosejcarr: There is a really interesting question about how we should calculate radius vs area.

notebooks/py/DARTFISH_Pipeline_-_Human_Occipital_Cortex_-_1_FOV.py
```diff
@@ -225,6 +225,18 @@ def compute_magnitudes(stack, norm_order=2):
 spot_intensities, results = psd.run(zero_norm_stack)
 # EPY: END code
 
+# EPY: START code
```
ambrosejcarr: Think about removing this. Are these needed?

starfish/test/codebook/test_from_json.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Test that codebook can be saved to disk and recovered, and that the recovered codebook is
+    identical to the one that it was serialized from.
+    """
+    # Create a codebook
+    codebook_array = codebook_json_data_factory()
+    codebook = Codebook.from_code_array(codebook_array)
+
+    # Dump it to a temporary file
+    directory = tempfile.mkdtemp()
```
ttung: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

starfish/spots/_detector/combine_adjacent_features.py
```diff
@@ -14,120 +18,358 @@ class ConnectedComponentDecodingResult(NamedTuple):
...
+            self._single_spot_attributes,
+            decoded_image=decoded_image,
+            target_map=target_map,
+            min_area=self._min_area,
+            max_area=self._max_area
+        )
+
+        iterable = tqdm(region_properties)
+        results = mapfunc(applyfunc, iterable)
+        spot_attrs, passes_area_filter = zip(*results)
```
dganguli: let's make sure that when we filter spot attribute tables, we're filtering on the same quantities across the different decoders. right now, the max_channel_per_hyb_round decoder filters on area and intensity. this decoder filters on area only. ALL decoders should filter on area, intensity, and quality/distance. this will allow for better unification of our API, pipeline, etc. Perhaps leave a more craftfully worded version of his as an Issue?

starfish/test/codebook/test_metric_decode.py
```diff
@@ -0,0 +1,110 @@
...
+    intensity_table = IntensityTable.from_spot_data(data, SpotAttributes(spot_attributes_data))
+    return intensity_table
+
+
+def codebook_factory() -> Codebook:
+    """
+    Codebook with two codewords describing an experiment with three channels and two imaging rounds.
+    Both codes have two "on" channels.
+    """
+    codebook_array = [
```
ttung: does it make sense to piggy back on `codebook_array_factory`?

starfish/test/spots/detector/pixel_spot_detector/test_run.py
```diff
@@ -0,0 +1,7 @@
+# fixtures needed: codebook, ImageStack
```
ambrosejcarr: Nuke

notebooks/py/DARTFISH_Pipeline_-_Human_Occipital_Cortex_-_1_FOV.py
ambrosejcarr: `is_valid` ?

starfish/codebook.py
ambrosejcarr: image

View, not copy. I think we're OK. 

I didn't add a test to the suite, since we would have to return a non-`IntensityTable` to do so.

starfish/intensity_table.py
ambrosejcarr: Added check to `from_spot_data.` 👍

starfish/intensity_table.py
ambrosejcarr: Added note to #380

starfish/spots/_detector/combine_adjacent_features.py
ambrosejcarr: Updated #499

starfish/test/codebook/test_metric_decode.py
ambrosejcarr: Yes. As you suggested, I have made a follow-up ticket to aggregate these factories into a single module: #526.

ambrosejcarr:
I addressed all comments and uploaded fixed notebooks since the float PR likely won't be in until after the weekend. Previously I had waited until the floats PR to put the fixes in.

starfish/codebook.py
```diff
@@ -442,40 +456,45 @@ def metric_decode(
...
             norm_codes, norm_intensities, metric=metric)
 
-        exceeds_distance = np.where(metric_outputs > max_distance)[0]
-        target_index = pd.Index(targets)
-
-        # remove targets associated with distant codes
-        target_index.values[exceeds_distance] = 'None'
-        target_index.values[intensity_mask] = 'None'
+        # only targets with low hamming distances and high intensities should be retained
+        passes_filters = np.logical_and(
```
dganguli: mind blown. didn't know about np.logical_and

starfish/codebook.py
```diff
@@ -442,40 +456,45 @@ def metric_decode(
...
         metric_outputs, targets = self._approximate_nearest_code(
             norm_codes, norm_intensities, metric=metric)
 
-        exceeds_distance = np.where(metric_outputs > max_distance)[0]
-        target_index = pd.Index(targets)
-
-        # remove targets associated with distant codes
-        target_index.values[exceeds_distance] = 'None'
-        target_index.values[intensity_mask] = 'None'
+        # only targets with low hamming distances and high intensities should be retained
```
dganguli: nuke word hamming, distance is generalized via the 'metric' argument.

starfish/codebook.py
```diff
@@ -442,40 +456,45 @@ def metric_decode(
...
+        self._validate_decode_intensity_input_matches_codebook_shape(intensities)
+
         # normalize both the intensities and the codebook
         norm_intensities, norms = self._normalize_features(intensities, norm_order=norm_order)
         norm_codes, _ = self._normalize_features(self, norm_order=norm_order)
 
-        # mask low intensity features
-        intensity_mask = np.where(norms < min_intensity)[0]
-
         metric_outputs, targets = self._approximate_nearest_code(
```
dganguli: i don't remember if this was true in master or not. but we are taking a performance hit if we:

slower behavior
1. decode
2. threshold on intensity

versus.

faster behavior
1. threshold on intensity
2. decode

in the latter, we have less rows of data to pass to sklearn.nearestneighbors, which i believe is an n^2 operation.  on the other hand, the slower behavior does let us keep all distances computed for all pixels. 

my intuition  is that there should be an option in this function that lets you toggle between behavior 1 and behavior 2. perhaps this is outside the context for this PR -- but what do you think?

starfish/types/_constants.py
```diff
@@ -40,3 +40,5 @@ class Features:
     CODE_VALUE = 'v'
     SPOT_RADIUS = 'radius'
     DISTANCE = 'distance'
+    PASSES_FILTERS = 'passes_filters'
```
dganguli: I still think this should be called PASSES_THRESHOLDS. filters get's conflated with image filtering, which I find deeply confusing.

starfish/intensity_table.py
ttung: Actually, I think you can forgo all this verification.  That is what type annotations is supposed to do for you.  The only place we don't have type annotation coverage is the notebooks, and it's likely it'll bomb out anyway if it's the wrong type.

starfish/test/codebook/test_metric_decode.py
ttung: Why can't I use :stately-ambrose: here? :(

ambrosejcarr:
> Why can't I use :stately-ambrose: here? :(

`#starfish-github` translated it! Good to know...

starfish/codebook.py
ambrosejcarr: Spoke offline, current implementation is OK due to use of approximate nearest neighbors algorithm, which is much faster than n^2. 

However, I created an issue to address this, as @dganguli correctly points out that we could make this much faster through intelligent pre-filtering  #532.

PR #501, opened 2018-09-03
Update Starfish to use Float data Exclusively

ambrosejcarr:
Summary of changes: 

Filters: 
- Added missing ability to set the number of requested processors
- Added utility method `util.preserve_float_range` to clip below zero and scale to max when values above 1 are detected. 
- Changed clipping behavior to adhere to the [0, 1] window

IntensityTable: 
- Convert values to float in [0, 1] to adhere to starfish. Right now it scales values to the full dynamic range, but that is easily changed. 

ImageStack: 
- cuts out a lot of code used to merge dtypes, uses `skimage.img_as_float` to convert. 
- adds `_validate_data_dtype_and_range` which is called when setting data in ImageStack (and during construction)
- adds warnings when data outside [0, 1] is found by `get_slice`
- changes synthetic spots to be emitted as float data

Tests: 
- Changes the image values of each pipeline test to match
- Dartfish and ISS did not require additional changes
- MERFISH x deconvolution produced a smalled subset of zeroed out channels, producing a bad result. 
- I reduced the number of convolutions to prevent this (9) and the result correlates well across the data. 

Notebooks: 
- All notebooks are updated. The only one that required significant work was MERFISH, which has some internal conversions

@dganguli it would be great if you could take a look at the pipeline tests and other non-testing code to verify this all matches your expectations for how we should deal with float data.


ambrosejcarr:
I don't understand why this isn't passing tests. These work for me on multiple local macosx environments. :(

starfish/image/_filter/bandpass.py
```diff
@@ -25,7 +25,7 @@ def __init__(
         threshold : float
             zero any pixels below this intensity value
         truncate : float
-            # TODO dganguli: this is not documented by trackpy, can you help?
+            truncate the filter at this many standard deviations
```
dganguli: truncate the gaussian kernel, used by the gaussian filter, by this many standard deviations.

starfish/image/_filter/util.py
```diff
@@ -1,6 +1,7 @@
 from typing import Tuple, Union
```
ambrosejcarr: We need to discuss this in depth, specifically how we deal with values greater than 1.

starfish/image/_registration/fourier_shift.py
```diff
@@ -84,6 +84,13 @@ def run(self, image: ImageStack, in_place: bool=True) -> Optional[ImageStack]:
                     data, axes = image.get_slice(indices=indices)
                     assert len(axes) == 0
                     result = shift_im(data, shift)
+
```
ambrosejcarr: I can and should use preserve float range here instead of this scaling.

starfish/intensity_table.py
```diff
@@ -244,6 +245,10 @@ def synthetic_intensities(
         data *= np.random.poisson(mean_photons_per_fluor, size=data.shape)
         data *= np.random.poisson(mean_fluor_per_spot, size=data.shape)
 
+        # convert data to float for consistency with starfish
```
ambrosejcarr: producing full dynamic range data no matter what, think about adopting preserve_float_range or skimage.img_as_float

starfish/stack.py
```diff
@@ -119,23 +119,26 @@ def __init__(self, image_partition: TileSet) -> None:
...
-                    dst_range = np.iinfo(self._data.dtype).max - np.iinfo(self._data.dtype).min + 1
-                    data = data * (dst_range / src_range)
-                warnings.warn(
-                    f"Tile "
-                    f"(R: {tile.indices[Indices.ROUND]} C: {tile.indices[Indices.CH]} "
-                    f"Z: {tile.indices[Indices.Z]}) has "
-                    f"dtype {data.dtype}.  One or more tiles is of a larger dtype "
-                    f"{self._data.dtype}.",
-                    DataFormatWarning)
+            data = img_as_float(data)
```
ambrosejcarr: What happens here if someone submits a uint16 and another uint32 tile? Probably worth a warning.

ambrosejcarr: Tony would put this back.

starfish/stack.py
```diff
@@ -119,23 +119,26 @@ def __init__(self, image_partition: TileSet) -> None:
             c = tile.indices[Indices.CH]
             zlayer = tile.indices.get(Indices.Z, 0)
             data = tile.numpy_array
-            if max_size != data.dtype.itemsize:
```
ttung: I would put this back.  It's a very weird situation and probably should flag it.

starfish/stack.py
ttung: dive into this to understand what we do here.

starfish/stack.py
```diff
@@ -119,23 +119,26 @@ def __init__(self, image_partition: TileSet) -> None:
...
-                    DataFormatWarning)
+            data = img_as_float(data)
             self.set_slice(indices={Indices.ROUND: h, Indices.CH: c, Indices.Z: zlayer}, data=data)
+
         # set_slice will mark the data as needing writeback, so we need to unset that.
         self._data_needs_writeback = False
 
+    @staticmethod
+    def _validate_data_dtype_and_range(data: Union[np.ndarray, xr.DataArray]) -> None:
+        """verify that data is of dtype float and in range [0, 1]"""
```
ttung: shouldn't this be `[0, 1)`

starfish/stack.py
```diff
@@ -1019,8 +1040,15 @@ def select_uint_dtype(array):
...
 
-        # set correct dtype and convert to stack
+        # set the smallest int datatype that supports the data's intensity range
         image = select_uint_dtype(image)
+
+        # convert to float for ImageStack
+        with warnings.catch_warnings():
+            # possible precision loss when casting from uint to float is acceptable
+            warnings.simplefilter('ignore', UserWarning)
+            image = img_as_float(image)
```
ambrosejcarr: Think about img_as_float here.

starfish/test/image/test_slicedimage_dtype.py
```diff
@@ -2,6 +2,7 @@
```
ambrosejcarr: Try putting these back after I restore the warning in stack.

starfish/image/_filter/util.py
ambrosejcarr: Here's the start of an in-depth discussion of how we and other packages deal with values greater than 1: https://docs.google.com/document/d/1TNbzOOml-sOp9PEEM7Y8LQHansxPS02flEOFyetNfv8/edit#heading=h.7kmz03ilv9in

starfish/stack.py
ambrosejcarr: Dove in. https://docs.google.com/document/d/1TNbzOOml-sOp9PEEM7Y8LQHansxPS02flEOFyetNfv8/edit#heading=h.7kmz03ilv9in

ttung:
scandalous attempt to sneak in float64 into the starfish pipeline!

starfish/test/image/test_slicedimage_dtype.py
ambrosejcarr: tony fixed the rest of these.

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -50,47 +50,90 @@ def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
...
+        ----------
+        .. [1] http://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution
 
-        # here be dragons. img_deconv is a float. this should not work, but the result looks nice
-        # modulo boundary values? wtf indeed.
-        img_deconv = img_deconv.astype(np.uint16)
-        return img_deconv
+        Notes
+        -----
+        This code is copied from skimage.restoration. We copied it to implement a bugfix wherein
```
ttung: Can you point this to the PR that you're copying from?

starfish/image/_filter/util.py
```diff
@@ -135,4 +135,4 @@ def preserve_float_range(array: Union[xr.DataArray, np.ndarray]) -> Union[xr.Dat
             array[array < 0] = 0
     if np.any(array > 1):
         array /= array.max()
-    return array
+    return array.astype(np.float32)
```
ttung: why is this the right model?  shouldn't the input to this method already be a float32?

starfish/stack.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, image_partition: TileSet) -> None:
     @staticmethod
     def _validate_data_dtype_and_range(data: Union[np.ndarray, xr.DataArray]) -> None:
         """verify that data is of dtype float and in range [0, 1]"""
-        if data.dtype != float:
+        if data.dtype != np.float32:
             raise TypeError(
                 f"ImageStack data must be of type float, not {data.dtype}. Please convert data "
```
ttung: float32

starfish/stack.py
```diff
@@ -259,7 +259,7 @@ def get_slice(
         slice_list, axes = self._build_slice_list(indices)
         result = self._data.values[slice_list]
 
-        if result.dtype != float:
+        if result.dtype != np.float32:
             warnings.warn(
                 f"Non-float dtype: {result.dtype} detected. Data has likely been set using private "
```
ttung: non float-32

starfish/image/_filter/util.py
ttung: Are there instances where this gets passed a non-float32 type?

starfish/test/full_pipelines/api/test_dartfish.py
```diff
@@ -146,7 +146,7 @@ def test_dartfish_pipeline_cropped_data():
 
     # verify number of spots detected
     spots_passing_filters = spot_intensities[Features.PASSES_THRESHOLDS].sum()
-    assert spots_passing_filters == 54
+    assert spots_passing_filters == 53
```
ttung: say what?

starfish/test/full_pipelines/api/test_dartfish.py
```diff
@@ -179,4 +179,4 @@ def test_dartfish_pipeline_cropped_data():
     corrcoef = np.corrcoef(x, y)
     corrcoef = corrcoef[0, 1]
 
-    assert np.round(corrcoef, 5) == -0.00992
+    assert np.round(corrcoef, 5) == 0.04422
```
ttung: say what?

starfish/stack.py
```diff
@@ -18,7 +18,7 @@
 from scipy.ndimage.filters import gaussian_filter
 from scipy.stats import scoreatpercentile
 from skimage import exposure
-from skimage import img_as_float, img_as_uint
+from skimage import img_as_float32, img_as_uint
```
ttung: I realize you may hate me, but I think a prudent way of handling this is storing img_as_float32 and np.float32 as "constants" in `starfish.types`, and use that throughout the codebase.

starfish/image/_filter/util.py
ambrosejcarr: Yes, when e.g. a filter like Richardson Lucy casts to float 64. This way, we can just reduce precision at the end after we've already guaranteed we're in the right scale. 

I think this is the simplest place to add this guarantee of the right first size

starfish/stack.py
ambrosejcarr: Yea, I could see this being useful. Good suggestion. :+1:

starfish/test/full_pipelines/api/test_dartfish.py
ambrosejcarr: This reverts a previous value change, so I was not concerned about it:

5194045

starfish/test/full_pipelines/api/test_dartfish.py
ambrosejcarr: ... but now that I go back and look, this is not the correct value for the correlation. Thanks for raising. 

I'll dig in and figure out why this changed.

ambrosejcarr: 1-off -- Due to the reversion to the previous "`53` identified targets" instead of `54`, we now fail to observe 1 `RELN` count, and as a result, the correlation shifts. This is a direct consequence of the above change.

starfish/image/_filter/util.py
```diff
@@ -108,3 +109,30 @@ def validate_and_broadcast_kernel_size(
...
+        Array whose values are in the interval [0, 1].
+
+    """
+    array = array.copy()
+    if np.any(array < 0):
+        if isinstance(array, xr.DataArray):
+            array.values[array.values < 0] = 0
+        else:
+            array[array < 0] = 0
+    if np.any(array > 1):
```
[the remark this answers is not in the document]
ambrosejcarr: Not needed; see comment in slack about this being a bit of a red herring. `Numpy` deals with machine precision in integer vs float comparisons:

https://spacetx.slack.com/archives/CAR91U354/p1536763650000100

starfish/stack.py
```diff
@@ -119,23 +120,36 @@ def __init__(self, image_partition: TileSet) -> None:
...
 
+    @staticmethod
+    def _validate_data_dtype_and_range(data: Union[np.ndarray, xr.DataArray]) -> None:
+        """verify that data is of dtype float and in range [0, 1]"""
+        if data.dtype != float:
+            raise TypeError(
+                f"ImageStack data must be of type float, not {data.dtype}. Please convert data "
+                f"using skimage.img_as_float prior to calling set_slice."
+            )
+        if np.min(data) < 0 or np.max(data) > 1:
```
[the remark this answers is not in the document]
ambrosejcarr: see above.

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -176,20 +254,24 @@ def test_merfish_pipeline_cropped_data():
...
+        expected_counts.index,
+        result_counts.index
     )
 
     tmp = pd.concat([result_counts, benchmark_counts], join='inner', axis=1).values
 
     corrcoef = np.corrcoef(tmp[:, 1], tmp[:, 0])[0, 1]
 
-    assert np.round(corrcoef, 4) == 0.9166
+    assert np.round(corrcoef, 4) == 0.9739
```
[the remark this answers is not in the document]
ambrosejcarr: There's something weird going on with MERFISH x local installation environment. Are you on a virtual environment? If no, can you send @ttung a freeze of your environment? This is coming up for me as well.

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -76,21 +76,24 @@ def richardson_lucy_deconv(
             Deconvolved image, same shape as input
 
         """
-
-        # TODO ambrosejcarr: the restoration function is producing the following warning:
-        # /usr/local/lib/python3.6/site-packages/skimage/restoration/deconvolution.py:389:
-        # RuntimeWarning: invalid value encountered in true_divide:
-        # relative_blur = image / convolve_method(im_deconv, psf, 'same')
-        img_deconv: np.ndarray = restoration.richardson_lucy(
-            img, psf, iterations=num_iter, clip=clip
```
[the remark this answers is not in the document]
ambrosejcarr: Correct.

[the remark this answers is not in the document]
ambrosejcarr: #547

starfish/stack.py
ambrosejcarr: #549

PR #505, opened 2018-09-04
remove unused run method

ttung:


ambrosejcarr:
> We could do that, but it would not have a meaningful method signature.

Ah, that's a good point. Which do you think is better? I'm fine with whichever you decide.

PR #508, opened 2018-09-04
Integrate sptx-format into starfish

ttung:
A few items are not yet complete:
1. CI/CD integration
2. Docker / conda files not yet integrated.
3. README not integrated.
4. Did not fix lint warnings (this would create a lot of diff churn that are not meaningful to review).

Test plan: `python setup.py sdist` and pip install that zip in a different virtualenv.  run starfish and `validate-sptx --experiment-json https://dmf0bdeheu4zf.cloudfront.net/20180828/DARTFISH/experiment.json`


ambrosejcarr:
When you say "README not integrated" Do you just mean that the README is sitting in a subdir and is not referenced by the main starfish one? If that's the case, then I think this is good to go. Also, jokes on you -- pytest is smarter than you think 😀 

image

ambrosejcarr:
I am super excited to try to start validating our data and fix all the breakages :-)

ttung:
> When you say "README not integrated" Do you just mean that the README is sitting in a subdir and is not referenced by the main starfish one? If that's the case, then I think this is good to go.

Yes, that is correct.

> Also, jokes on you -- pytest is smarter than you think 😀

I just meant that I didn't look at `.travis.yml` from https://github.com/spacetx/sptx-format/.  But it turns out there's nothing special being done there! :)

PR #515, opened 2018-09-05
Remove the fov id from the field of view manifests

ttung:
They are redundant since the field of view manifests are indexed by their name already.

With this change, https://github.com/spacetx/slicedimage/pull/55 *should* no longer be necessary.


joshmoore:
Cheers. That was (I think) my last blocker to having a passing validation!

joshmoore:
Oh boy. 🍰 from Tony for Josh.

PR #531, opened 2018-09-08
Add sphinx autobuild support

ambrosejcarr:
- Adds a basic doc hierarchy to starfish that we can add to. 
- Adds sphinx-autobuild to eventually support readthedocs
- Integrates some of the READMEs from sptx-format into the new doc hierarchy
- Demonstrates the autodocing for "core modules": `ImageStack, Codebook`, and `IntensityTable` and fixes documentation bugs. 
- Does not implement auto-doc for any other modules (yet!)
- Does not add readthedocs (we don't want to point users at this until it's useful)

cc @joshmoore in case you know about sphinx and have feelings about this kinda stuff.


docs/source/contributing/contributing.md
```diff
@@ -2,15 +2,15 @@
```
ttung: This is a move?  Doesn't README.md link to this?

ambrosejcarr: Ah, I missed that. Sorry -- yes I will need to adjust the README.md link. 

I agree, I was lazy and just included the sphinx makefile. I'll teach myself the syntax and add that. I imagine I can port the logic present in `subdir.mk` for notebooks?

ttung: > I agree, I was lazy and just included the sphinx makefile. I'll teach myself the syntax and add that. I imagine I can port the logic present in subdir.mk for notebooks?

You could do that or you could call out using recursive make.  Given that this is a standardized makefile, I would do the latter.

ambrosejcarr: > You could do that or you could call out using recursive make. Given that this is a standardized makefile, I would do the latter.

https://github.com/spacetx/starfish/issues/536

PR #537, opened 2018-09-10
Follow the same conversion steps as ImageStack would be for the conversion tests

ttung:
It's convert to float, then merge.  Right now, we're merging and then converting to float.

This fixes the tests in `starfish/test/image`


ttung:
I set `ajc-floats-only` as the base branch. Feel free to merge this if you approve.

PR #523, opened 2018-09-06
The Fuzz

joshmoore:
* Adds `starfish validate`  (from https://github.com/spacetx/starfish/issues/511#issuecomment-419117650)
 * Adds `starfish validate --fuzz` (closes #513 )
 * Adds `starfish.util.exec.stages` (TODO from https://github.com/spacetx/starfish/pull/519/files#diff-6ea6d7c599292e7adc3c2e49933ea7b7R33)

Likely to conflict with gh-514 if we can get that in.

PR #538, opened 2018-09-11
Propagate validation

joshmoore:
* [x] handle @ttung comments from gh-523
 * [x] idea for a `strict` flag (and/or `STARFISH_STRICT_LOADING` variable) to `Experiment.from_json` 
   - [x] prevents needing to find every experiment loading and explicitly adding validation
   - [ ] needs to cache the json rather than using `resolve_path_or_url` twice
* [x] (temporarily) demonstrate tying together starfish & slicedimage without a release of slicedimage
* [ ] decide between (or for both of) eager and lazy validation

Test plan:
 * set STARFISH_STRICT_LOADING to true in your environment and then do any loading via Experiment.from_json.
 * Also try either corrupting a fileset or loading an older fileset and see the failure.


joshmoore:
I've rolled back the temporary testing of `validate()` on slicedimage objects. That needs discussion in https://github.com/spacetx/slicedimage/pull/61. This has been updated with the new travis build.

joshmoore:
Oops: https://travis-ci.org/spacetx/starfish/jobs/430949120

```
6.03s$ make run__notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
[ -e notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py.skip ] || python notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
/home/travis/build/spacetx/starfish/validate_sptx/util.py:80: UserWarning: 
 Additional properties are not allowed ('fov' was unexpected)
	Schema:         	https://github.com/spacetx/starfish/sptx-format/schema/field_of_view/field_of_view.json
	Subschema level:	0
	Path to error:  	additionalProperties
	Filename:       	fov_001/hybridization.json
  warnings.warn(message)
/home/travis/build/spacetx/starfish/validate_sptx/util.py:80: UserWarning: 
 Additional properties are not allowed ('fov' was unexpected)
	Schema:         	https://github.com/spacetx/starfish/sptx-format/schema/field_of_view/field_of_view.json
	Subschema level:	0
	Path to error:  	additionalProperties
	Filename:       	fov_001/nuclei.json
  warnings.warn(message)
Traceback (most recent call last):
  File "notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py", line 33, in <module>
    experiment = Experiment.from_json("https://dmf0bdeheu4zf.cloudfront.net/20180919/MERFISH/experiment.json")
  File "/home/travis/build/spacetx/starfish/starfish/experiment/experiment.py", line 186, in from_json
    raise Exception("validation failed")
Exception: validation failed
make: *** [run__notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py] Error 1
```

Getting this green may require disabling notebooks or having other PRs merged.

joshmoore:
Actually, I take that back. With gh-515, we actually need a new data release *or* we need to add "fov" back in and make it optional with some form of jsonschema-`deprecation` flag.

@ambrosejcarr @ttung

starfish/experiment/experiment.py
```diff
@@ -151,21 +153,38 @@ def __repr__(self):
...
 
         Returns
         -------
         Experiment :
             Experiment object serving the requested experiment data
 
+        Environment variables
+        ---------------------
+        STARFISH_STRICT_LOADING :
+             If set, then all JSON loaded by this method will be
```
ttung: I think the documentation here is somewhat unclear.  One could read it as:

`STARFISH_STRICT_LOADING` set, but strict=False ==> no validation.

starfish/util/exec.py
```diff
@@ -36,11 +35,17 @@ def stages(commands: Sequence[Sequence[Union[str, Callable]]],
...
     Environment variables
     ---------------------
-    STARFISH_COVERAGE:
+    STARFISH_COVERAGE :
          If set, then command lists will have `coverage run ...` options
          prepended before execution.
     """
-    tempdir = tempfile.mkdtemp()
+
+    if keep_data:
```
ttung: I could have sworn there was a `delete` option for `tempfile.TemporaryDirectory()`, but clearly I'm wrong.  In that case, I think the original code is ok as `tempfile.TemporaryDirectory()` doesn't really provide any significant improvement.

validate_sptx/validate_sptx.py
```diff
@@ -48,7 +48,10 @@ def validate(experiment_json: str, fuzz: bool=False) -> bool:
 
     # validate experiment.json
     experiment_validator = SpaceTxValidator(_get_absolute_schema_path('experiment.json'))
-    valid &= experiment_validator.validate_object(experiment, name, fuzz=fuzz)
+    if fuzz:
```
ttung: why not a top-level command fuzz()?

starfish/experiment/experiment.py
joshmoore: Pushed  a clarification.

starfish/util/exec.py
joshmoore: Happy to rollback, but the use of TD didn't feel particularly dirtier than of shutil.

validate_sptx/validate_sptx.py
ttung: My instinct is that a separate command makes a lot more sense.  However, your concerns about code duplication is on point.

In the long term, I'd like the validation schema to somehow reference the child schema docs so this traversal can be done programmatically rather than explicitly.  e.g., the schema for the `codebook` field in experiment.json should reference the codebook schema.  As we are validating the `codebook` field, it should be know to load up `codebook/codebook.json` and validate the referenced file against that.

My guess is that jsonschema does not provide facilities to do this, and we would need to build our own internal referencing system.

Let's leave the code as is for now, and return to it when we have cycles.

starfish/util/exec.py
ttung: Ok, I'll leave it up to your judgement.

validate_sptx/validate_sptx.py
joshmoore: Interesting. I actually remembered the parent schema's referring to the child schemas. Perhaps this was something on the HCA side. I've listed this as a follow-up in https://github.com/spacetx/starfish/issues/539.

On the code side, another idea (also for more cycles) would be to convert this to an instance as opposed to just a function. That might allow for both clarity and code-reuse.

joshmoore: e.g. I think `codebook` is missing its `$ref`: https://github.com/spacetx/starfish/blob/master/validate_sptx/schema/experiment.json#L14

ttung:
Please wait for #619 to land before you land this. :)  Or you can approve + land + rebase.

joshmoore:
That's all notebooks green with validation. 🎉

validate_sptx/validate_sptx.py
ambrosejcarr: > Interesting. I actually remembered the parent schema's referring to the child schemas. Perhaps this was something on the HCA side. I've listed this as a follow-up in #539.

My current level of understanding is that explicit `$ref` linking means that the file expects to see an `object` in place of the schema, rather than a file containing the object, meaning that we'd end up with one big json blob. If that's _wrong_, then there is a big opportunity for us to clean up the specification by removing these implicit links...

joshmoore: No, this is the same issue as https://github.com/spacetx/starfish/pull/647#discussion_r220971186, i.e. you can pull other schema files into your main schema, but I don't believe you can currently pull json instance documents into the current json document.

joshmoore: Some glimmer of hope in json-refs, e.g. https://jsonref.readthedocs.io/en/latest/

PR #541, opened 2018-09-11
Remove (the right) prematurely checked in notebook

ambrosejcarr:
Removing the correct notebook this time.


ambrosejcarr:
> don't you need to add back the assay_comparison notebook?

# 540

PR #550, opened 2018-09-12
Freeze requirements

ttung:
Add a makefile rule (ported from the one I wrote for the HCA DSS) to repin requirements.


ambrosejcarr:
Can you explain what this does?

ttung:
general usage:

1. edits go to REQUIREMENTS.txt.in and REQUIREMENTS-DEV.txt.in
1. `make REQUIREMENTS.txt` or `make REQUIREMENTS-DEV.txt` will build each.  `make refresh_all_requirements` will rebuild both for the lazy ppl.

ttung:
Alternatively, we could just pin a few core requirements (numpy, scikit-image).  Thoughts?

ttung:
Ready for review.

ttung:
Added missing files.  Edited autogenerated warning.

README.md
```diff
@@ -8,23 +8,19 @@ See this [document](https://docs.google.com/document/d/1IHIngoMKr-Tnft2xOI3Q-5rL
...
-pip 10.0.1 from /usr/local/lib/python3.6/site-packages/pip (python 3.6)
-```
-
-While not required, you may wish to set up a [virtualenv](https://virtualenv.pypa.io/en/stable/). To do this, execute:
+Starfish names its dependencies and lists explicit versions, due to sensitivity to subtle algorithm changes.  For that reason, it is strongly encouraged that you set up a [virtualenv](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments).  Create a work folder and set up the virtual environment like:
 ```
+% mkdir starfish
+% cd starfish
 % python -m venv .venv
+% .venv/bin/activate
```
ambrosejcarr: needs `source .venv/bin/activate` right?

Makefile
```diff
@@ -21,6 +22,24 @@ mypy:
 docs-%:
 	make -C docs $*
 
+refresh_all_requirements:
```
ambrosejcarr: the individual make commands work fine, this one throws an error for me. 

```bash
$ make refresh_all_requirements 
[ ! -e .requirements-env ] || exit 1
virtualenv .REQUIREMENTS.txt.in-env
/bin/bash: virtualenv: command not found
make[1]: *** [REQUIREMENTS.txt] Error 127
make: *** [refresh_all_requirements] Error 2
(.venv) 
``` 

I used a pre-existing virtualenv, although it was set up according to your commands below. Could that contribute? 

I'm inside the virtualenv when I call this, as well.

README.md
ttung: right.

Makefile
ttung: Let me try an alternative.

ambrosejcarr:
> Alternatively, we could just pin a few core requirements (numpy, scikit-image). Thoughts?

I think this would cause all the problems with pinning with none of the security it should provide. I think it's all or none. :-)

joshmoore:
Down the road, https://github.com/pypa/pipfile may be a more official solution.

PR #551, opened 2018-09-12
ENH: add writer to MERMAID's input data format

ambrosejcarr:
- Adds a simple writer to emit `csv.gz` from IntensityTable in the form of `x, y, gene, gene`. The result is usable in MERMAID as described in the documentation below. 

cc @JEFworks


starfish/intensity_table.py
```diff
@@ -172,6 +172,39 @@ def save(self, filename: str) -> None:
...
+        # verify the IntensityTable has been decoded
+        if Features.TARGET not in self.coords.keys():
+            raise RuntimeError(
+                'IntensityTable must be decoded before it can be converted to MERMAID input.'
+            )
+
+        # construct the MERMAID dataframe
+        df = self.to_features_dataframe()
+        mermaid_data = df[[Indices.X, Indices.Y, Features.TARGET, Features.TARGET]]
+        mermaid_data.columns = ['x', 'y', 'gene1', 'gene2']
```
dganguli: i just confirmed from the mermaid code/example data, that you can add an arbitrary number of genes here, but they will be the same column. the viz generates a drop down menu and a color for any categorical variable. so this is consistent with the demo currently on the website

starfish/intensity_table.py
```diff
@@ -172,6 +172,39 @@ def save(self, filename: str) -> None:
         """
         self.to_netcdf(filename)
 
+    def save_mermaid(self, filename: str) -> pd.DataFrame:
```
ttung: Would suggest making intensity_table a package, and separating out peripheral code like this to a separate module.

ambrosejcarr: I've done the package conversion in #662. Could you be more specific about how you would factor out the code? I looked into [pandas](https://github.com/pandas-dev/pandas/blob/v0.23.4/pandas/core/frame.py#L1515) and [xarray](https://github.com/pydata/xarray/blob/ab96954883200f764a0dd50870e4db240c119265/xarray/core/dataarray.py#L1594) both of which leave their writers attached to their main objects. 

`Numpy`, by contrast, uses the `np.save(obj, file_handle_or_filename)` pattern. Personally I find the `pandas`/`xarray`/`current starfish` version more intuitive.

ambrosejcarr:
Added tests, addressed comment as best I could.

starfish/intensity_table.py
ttung: Now you can make a starfish/intensity_table/conversion.py file, with a `save_mermaid` method that accepts an IntensityTable and writes to a file.

ambrosejcarr: > Now you can make a starfish/intensity_table/conversion.py file, with a save_mermaid method that accepts an IntensityTable and writes to a file.

Ok, but I dislike the UX of this. I think it's easier for users to type `my_intensity_table.save_` and hit tab to get the options, vs look around in the docs for the way to do it and realize it's in `starfish.IntensityTable`, unless I'm missing the way to attach it back to `IntensityTable`

ttung: So there's a few options:

1. Allow our modules to become enormous.  I am not a fan of this, because it makes code hard to work with.
2. Create stub functions in IntensityTable that call out to methods implemented elsewhere.  This means our docs are probably complete, but we have a bunch of stub code laying around.
3. Create stub functions in IntensityTable at constructor time.  We don't have a bunch of stub code laying around, but our autodocs will probably not generate correctly.

I kind of suspect we will have to do (1), as much as I dislike it...

Issue #384, opened 2018-08-09
Make out-of-place computation the default


ambrosejcarr:
- Make in-place=False the default for apply loops and filters in `ImageStack`.

ttung:
Why does it suck?

ambrosejcarr:
It's great for memory utilization when the algorithms are designed to do piecewise in-place computation, but makes it much harder to develop/debug in interactive sessions.

ttung:
Interesting.  I bet that internally, the algorithms are not in-place anyway.

ambrosejcarr:
I believe this can now be closed.

ambrosejcarr:
Reopening because in-place is still default -- originally the ticket was closed because out-of-place is now _possible_ everywhere, but the default still needs to be changed.

PR #554, opened 2018-09-12
Saxelrod merfish scale by percent

shanaxel42:


starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -239,11 +235,11 @@ def test_merfish_pipeline_cropped_data():
         norm_order=2,
         crop_size=(0, 40, 40)
     )
-    spot_intensities, prop_results = psd.run(low_passed)
+    spot_intensities, prop_results = psd.run(scaled_image)
```
ambrosejcarr: Question here: what is the average intensity (float value) across the `scaled_image` before and after this change? I would scale the threshold used in the PixelSpotDetector by this value, and it might (or might not 🙃) bring the number of counts back down towards the number of spots_passing_filters.

notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -25,6 +25,7 @@
 
 from showit import image as show_image
 from starfish.experiment import Experiment
+from starfish.image._filter.scale_by_percentile import ScaleByPercentile
```
ttung: Would prefer that we do:

```

sc_filt = Filter.ScaleByPercentile(..)
```

notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -107,27 +108,9 @@
...
-# at least we can do it with get_slice and set_slice right now.
-from copy import deepcopy
-scaled_image = deepcopy(low_passed)
-
-for indices in primary_image._iter_indices():
-    data = scaled_image.get_slice(indices)[0]
-    scaled = data / scale_factors[indices[Indices.ROUND.value], indices[Indices.CH.value]]
-    scaled_image.set_slice(indices, scaled)
+sc_filt = ScaleByPercentile(p=90)
+scaled_image = sc_filt.run(low_passed, in_place=False)
```
ttung: I thought our default is supposed to be in_place=False now, but it looks like this is not the case for ScaleByPercentile...?

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -179,53 +180,48 @@ def test_merfish_pipeline_cropped_data():
         low_passed.numpy_array[5, 0, 0, 40:50, 45:55]
     )
 
-    # scale the data by the scale factors
```
ttung: yay!

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -179,53 +180,48 @@ def test_merfish_pipeline_cropped_data():
...
+         [6.18005455e-01, 5.73010147e-01, 7.28941679e-01, 9.46671605e-01,
+            9.61733460e-01, 6.58423424e-01, 2.80564368e-01, 7.11256936e-02,
+            1.05127227e-02, 9.25380969e-04],
+         [4.02738214e-01, 5.31649411e-01, 9.14191008e-01, 1.40832996e+00,
+            1.56639075e+00, 1.12429368e+00, 4.89756435e-01, 1.24755450e-01,
+            1.83621347e-02, 1.97034585e-03]],
         dtype=np.float32
     )
+
+    # print(scaled_image.numpy_array[5, 0, 0, 40:50, 45:55])
```
ttung: intentional?

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -274,4 +270,4 @@ def test_merfish_pipeline_cropped_data():
 
     corrcoef = np.corrcoef(tmp[:, 1], tmp[:, 0])[0, 1]
 
-    assert np.round(corrcoef, 4) == 0.9739
+    assert np.round(corrcoef, 4) == 0.894
```
ttung: this seems like a pretty big change.  @dganguli / @ambrosejcarr this ok?

starfish/test/full_pipelines/api/test_merfish.py
shanaxel42: whoops

notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
shanaxel42: Yea scale by percentile still has in_place=True default, but I could change it?

ttung: oh hm, it's still True everywhere.  

@ambrosejcarr : what happened re https://github.com/spacetx/starfish/issues/384?  It was closed but why?

starfish/test/full_pipelines/api/test_merfish.py
dganguli: well, did floats get merged? if so, this should be re-based off of master, in which case the the jump is from 0.93 -> 0.894 instead of 0.9739 -> 0.894. which is less bad?

regardless, i would leave the test_merfish.py as is. But NOT commit the notebook (or it's .py representation). This has the big con that the test diverges from the notebook. But that's OK I think. The point of the test is to test the CLI -- which this does. 

The real solution, is to implement a ScaleByValues filter, which makes a PipelineComponent that achieves L115-L130 of the .py representation of the notebook. Then, the test would also use this filter. Then the test and the notebook would be in sync. 

But since we don't have that solution, I think we just make an Issue, nuke the .ipynb (and it's .py version) and merge. Thoughts?

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -179,53 +180,47 @@ def test_merfish_pipeline_cropped_data():
...
-    # scale the data by the scale factors
-    scale_factors = {
-        (t[Indices.ROUND], t[Indices.CH]): t['scale_factor']
-        for t in experiment.extras['scale_factors']
-    }
-    for indices in low_passed._iter_indices():
-        data = low_passed.get_slice(indices)[0]
-        scaled = data / scale_factors[indices[Indices.ROUND.value], indices[Indices.CH.value]]
-        low_passed.set_slice(indices, scaled)
+    sc_filt = ScaleByPercentile(p=90)
```
ambrosejcarr: I think we need to adjust this filter a little bit to have this make sense. As it stands, the filter is _scaling by the max_value_ which is directly counter to what it says it wants to do. The below diff would fix that by enabling above-max-value intensities to get clipped. In this regime, all values are actually scaled by the stated percentile. 

```diff
$ g diff starfish/image/_filter/scale_by_percentile.py
diff --git a/starfish/image/_filter/scale_by_percentile.py b/starfish/image/_filter/scale_by_percentile.py
index 07e9d72..de57a7e 100644
--- a/starfish/image/_filter/scale_by_percentile.py
+++ b/starfish/image/_filter/scale_by_percentile.py
@@ -5,6 +5,7 @@ import numpy as np
 
 from starfish.stack import ImageStack
 from ._base import FilterAlgorithmBase
+from .util import preserve_float_range
 
 
 class ScaleByPercentile(FilterAlgorithmBase):
@@ -30,16 +31,21 @@ class ScaleByPercentile(FilterAlgorithmBase):
             "--p", default=100, type=int, help="scale images by this percentile")
 
     @staticmethod
-    def scale(image: np.ndarray, p: int) -> np.ndarray:
+    def scale(image: np.ndarray, p: int, clip=True) -> np.ndarray:
         """Clip values of img below and above percentiles p_min and p_max
 
         Parameters
         ----------
         image : np.ndarray
             image to be scaled
-
         p : int
             each image in the stack is scaled by this percentile. must be in [0, 100]
+        clip : bool
+            if True, clip any values that end up out of scale as a result of filtering. This has
+            the effect of eliminating outlier intensities. If False, the top-end value will be
+            used to rescale the dynamic range to [0, 1], which may mitigate or eliminate the effect
+            of scaling in the case that the image was already saturated (contains 1-valued pixels).
+            (Default = True)
 
         Notes
         -----
@@ -54,11 +60,12 @@ class ScaleByPercentile(FilterAlgorithmBase):
         """
         v = np.percentile(image, p)
 
-        # asking for a float percentile clipping value from an integer image will
-        # convert to float, so store the dtype so it can be restored
-        dtype = image.dtype
         image = image / v
-        return image.astype(dtype)
+        if clip:
+            image[image > 1] = 1
+        image = preserve_float_range(image)
+
+        return image
 
     def run(
             self, stack: ImageStack, in_place: bool=True, verbose: bool=False,
```

starfish/test/full_pipelines/api/test_merfish.py
ambrosejcarr: > well, did floats get merged? if so, this should be re-based off of master, in which case the the jump is from 0.93 -> 0.894 instead of 0.9739 -> 0.894. which is less bad?

This is actually the correct correlation -- this test works off a smaller part of one of the images, and because this subset is _not_ affected negatively by deconvolution (which has per-_image_ funkiness), it does not change, and is actually 0.97. Does that make sense @dganguli ?

notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
ambrosejcarr: I misinterpreted the ticket -- I closed it because out-of-place is now _possible_ everywhere. Thanks for finding this, will reopen and remember to leave more specific messages.

starfish/test/full_pipelines/api/test_merfish.py
shanaxel42: This is based off master, so this is actually the api test. I thought we were going to try and unify the api tests and the notebooks by calling the notebook in the api test instead of going though all the steps. If that's the case I can just remove the filter from this api test and notebook and only add it to the cli test in my other branch

PR #516, opened 2018-09-05
Add validation to test_iss.py

joshmoore:
Depends on gh-515

Intended to be equivalent to [test.sh](https://github.com/joshmoore/build-starfish/commit/01446ce99886387e2024d6e0c6575fd2c39983bb)

PR #557, opened 2018-09-13
Adding CLI pipeline tests

shanaxel42:
So this where I added all the cli pipeline tests. And in the process of getting them to work made some minor changes to some of the filter and detector files. Mostly just adding arguments to the argparser. But some other random stuff that I ran into as well like changing ImageStack.write to always rewrite the tile data. In hindsight this probably should have been split up into a few different PRs. So sorry in advance for that. The main changes are:
-base_cli test
-test_dartfish_cli 
-test_merfish_cli
-test_allen_smFISH cli. 

Everything else is in support of that.


ambrosejcarr:
Could you add a little summary of what I'm being requested to review here? :-)

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -45,14 +45,14 @@ def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
         group_parser.add_argument(
             '--sigma', type=float, help='standard deviation of gaussian kernel')
         group_parser.add_argument(
-            '--clip', action='store_true',
-            help='(default False) if True, clip values below -1 and above 1')
+            '--no-clip', action='store_false',
+            help='(default True) if True, clip values below -1 and above 1')
```
ambrosejcarr: Should be `below 0 and above 1` to follow the updated API.

starfish/image/_filter/zero_by_channel_magnitude.py
```diff
@@ -32,7 +32,10 @@ def __init__(self, thresh: int, normalize: bool, **kwargs) -> None:
     @classmethod
     def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
         group_parser.add_argument(
-            '--thresh', type=int, help='minimum magnitude threshold for pixels across channels')
+            '--thresh', type=float, help='minimum magnitude threshold for pixels across channels')
+        group_parser.add_argument(
+            '--normalize', action="store_true", help='Scales all rounds to have '
```
ambrosejcarr: would prefer help on next line and no wrapping.

starfish/spots/_detector/__init__.py
```diff
@@ -71,4 +71,8 @@ def _cli(cls, args, print_help=False):
             blobs_image=blobs_image,
             reference_image_from_max_projection=args.reference_image_from_max_projection
         )
+
+        # When PixelSpotDetector is used run() returns a tuple
```
ambrosejcarr: Any way we could adjust this to pickle and save the secondary results that it emits instead of dropping them? A general solution could be just to pickle and save additional results as "additional_spot_finder_results.pkl"

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -37,7 +40,10 @@ def __init__(
             an ImageStack (default = 0)
 
         """
-        self.codebook = codebook
+        if isinstance(codebook, str):
```
ambrosejcarr: We have churned on this before, and the agreement we came to is that this complexity should be banished to CLI-specific methods (`run`), since users are unlikely to need to load the `codebook` from a string. Is there a reason this isn't possible here?

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -9,7 +12,7 @@
 
 class PixelSpotDetector(SpotFinderAlgorithmBase):
     def __init__(
-            self, codebook: Codebook, metric: str, distance_threshold: float,
+            self, codebook, metric: str, distance_threshold: float,
```
ambrosejcarr: if keeping this (see below) it should be `Union[Codebook, str]`

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -48,7 +54,11 @@ def __init__(
         self.crop_y = crop_y
         self.crop_z = crop_z
 
-    def run(self, stack: ImageStack) -> Tuple[IntensityTable, ConnectedComponentDecodingResult]:
+    def run(
+            self, stack: ImageStack,
+            blobs_image: Optional[Union[np.ndarray, xr.DataArray]] = None,
```
ambrosejcarr: huh, I feel like it is confusing to have these here, since they're not used...

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -96,6 +106,16 @@ def add_arguments(cls, group_parser):
         group_parser.add_argument(
             "--area-threshold", type=float, default=2, help="minimum area of a feature"
```
ambrosejcarr: We also churned on this -- our decision was that we don't have enough scientific consensus to give good defaults, so we wouldn't provide any in those cases. I think the API does that properly here, and the CLI should match that.

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,54 @@
...
+    __test__ = False
+
+    SUBDIRS = Tuple[str]
+    STAGES = Tuple[List]
+
+    def verify_results(self, intensities):
+        pass
+
+    def test_run_pipline(self):
+        tempdir = tempfile.mkdtemp()
```
ambrosejcarr: I think tony likes `tempfile.TemporaryDirectory()`

starfish/test/full_pipelines/cli/test_allen_smFISH_cli.py
```diff
@@ -0,0 +1,81 @@
...
+            "--max-size", "3",
+            "--separation", "5",
+            "--percentile", "10",
+            "--is-volume"
+        ],
+
+    )
+
+    def verify_results(self, intensities):
+        # TODO DEEP SAYS WAIT ON THIS TEST TILL STUFF GETS FIGURED OUT
```
ambrosejcarr: Love it.

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import List, Tuple
+
+from starfish.intensity_table import IntensityTable
+from starfish.util import clock
+
+
+class CLITest(unittest.TestCase):
+    __test__ = False
+
+    SUBDIRS = Tuple[str]
```
ambrosejcarr: Check with @ttung about how to make it clear that these are abstract. 

Also this class needs docblocks explaining how it should be extended to create more tests. :-)

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,54 @@
...
+    STAGES = Tuple[List]
+
+    def verify_results(self, intensities):
```
ambrosejcarr: Instead of pass I'd use `raise NotImplementedError()`

starfish/test/full_pipelines/cli/test_dartfish_cli.py
```diff
@@ -0,0 +1,104 @@
...
+            "--codebook", lambda tempdir, *args, **kwargs: os.path.join(
+                tempdir, "registered", "codebook.json"),
+            "--distance-threshold", "3",
+            "--magnitude-threshold", ".5",
+            "--min-area", "5",
+            "--max-area", "30",
+        ],
+    )
+
+    def verify_results(self, intensities):
```
ambrosejcarr: This makes me think we should refactor the API tests to generate an object that this can check against as well. That way we have one source of truth. Maybe you're already ahead of me? :-)

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,54 @@
...
+import tempfile
+import unittest
+from typing import List, Tuple
+
+from starfish.intensity_table import IntensityTable
+from starfish.util import clock
+
+
+class CLITest(unittest.TestCase):
+    __test__ = False
```
ambrosejcarr: Can you add a comment about what this does (and that you have to make it true for the test to work, based on what I think I'm seeing below)?

starfish/test/full_pipelines/cli/test_iss.py
```diff
@@ -145,25 +144,13 @@ class TestWithIssData(unittest.TestCase):
                 "$['codebook']",
             ),
             "-o", lambda tempdir, *args, **kwargs: os.path.join(
-                tempdir, "results", "decoded-spots.nc"),
+                tempdir, "results", "spots.nc"),
```
ambrosejcarr: Previously we ran into some bugs where overwriting the spots file caused some serious problems. I would keep these having separate names from one another if possible.

starfish/test/full_pipelines/cli/test_iss.py
```diff
@@ -145,25 +144,13 @@ class TestWithIssData(unittest.TestCase):
...
-            gene_counts = pd.Series(counts, genes)
-            assert gene_counts['ACTB_human'] > gene_counts['ACTB_mouse']
-        finally:
-            if os.getenv("TEST_ISS_KEEP_DATA") is None:
-                shutil.rmtree(tempdir)
+    def verify_results(self, intensities):
+        genes, counts = np.unique(
+            intensities.coords[Features.TARGET], return_counts=True)
+        gene_counts = pd.Series(counts, genes)
+        assert gene_counts['ACTB_human'] > gene_counts['ACTB_mouse']
```
ambrosejcarr: Sigh, this assertion is pretty worthless (p ~ 0.5 that it catches an error :wink:). Can we add a TODO to make this a stronger test? 

Also I think I created a TEST dataset to replace these data. Could you slot that in instead of these data? If you have to swap the sign on the assertion you have my blessing.

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_run_pipline(self):
+        tempdir = tempfile.mkdtemp()
+        coverage_enabled = "STARFISH_COVERAGE" in os.environ
+        print(tempdir)
+
+        def callback(interval):
+            print(" ".join(stage[:2]), " ==> {} seconds".format(interval))
+
+        try:
+            for subdir in self.SUBDIRS:
```
joshmoore: Can you make use of `starfish.util.exec` here?

shanaxel42: Didn't see that! yes totally, I'll refactor

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,34 @@
...
+    STAGES = Tuple[List]
+    spots_file = "spots.nc"
+
+    def verify_results(self, intensities):
+        raise NotImplementedError()
+
+    def test_run_pipline(self):
+        tempdir = exec.stages(self.STAGES, self.SUBDIRS, keep_data=True)
+        intensities = IntensityTable.load(os.path.join(tempdir, "results", self.spots_file))
+        self.verify_results(intensities)
```
joshmoore: Nice. re:

 * https://github.com/spacetx/starfish/pull/516#issuecomment-419557387
 * specifically https://github.com/joshmoore/starfish/commit/d2d028499a31c3049bb9f08192e2dbeab0bf63c8#diff-3ab8577a09cb5f1f0d1527d305d08bd4

if you can see a way forward how to pass your `verify_results` _into_ `exec.stages` (perhaps as a post-processing stage that takes _only_ `tempdir` as input) then the `shutil.rmtree` block could be removed. Note: on review, @ttung asked me to change this from rmtree to a use of `TemporaryDirectory`. That's currently in https://github.com/spacetx/starfish/pull/538/commits/6a590d6ab7bde830514bb686466ee55453cf892f

starfish/test/full_pipelines/cli/_base_cli_test.py
```diff
@@ -0,0 +1,34 @@
+import os
+import shutil
+import unittest
+from typing import List, Tuple
+
+from starfish.intensity_table import IntensityTable
+from starfish.util import exec
+
+
+class CLITest(unittest.TestCase):
```
ttung: can this just be decorated with `unittest.skip`?
https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

shanaxel42: It looks like thats for individual test within a test class? If I put the decorator on the test_run_pipieline method in the base class the subclasses also get skipped.

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,31 @@
...
+from starfish.util.argparse import FsExistsType
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("experiment_url")
+    parser.add_argument("output_dir", type=FsExistsType())
+    args = parser.parse_args()
+
+    # save image stacks locally
```
shanaxel42: So I ended up doing this kinda hackily to avoid needing those zip files. I figured since this is just for these cli tests this might be ok but @ttung let me know if you want a more long term solution for saving an experiment object to a direcotry.

ttung: This is almost perfect.

I would just change it to:

```
for fov in exp.fovs():
    os.mkdir(fov.name)
    fov.write(....)
```

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,31 @@
...
+    exp = Experiment.from_json(args.experiment_url + 'experiment.json')
+
+    fov_dir = args.output_dir + '/fov_001/'
+    os.mkdir(fov_dir)
+
+    exp.fov().primary_image.write(fov_dir + 'hybridization.json')
+    for image_type in exp.fov().auxiliary_image_types:
+        exp.fov()[image_type].write(fov_dir + (image_type + '.json'))
+
+    # get codebook from url and save locally to tmp dir
```
ttung: Can you add a TODO to save experiment.json?

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -14,7 +14,7 @@
 class DeconvolvePSF(FilterAlgorithmBase):
 
     def __init__(
-            self, num_iter: int, sigma: Number, clip: bool=False, **kwargs) -> None:
```
ttung: Does this belong in this PR?

starfish/spots/_detector/__init__.py
```diff
@@ -40,6 +41,12 @@ def add_to_parser(cls, subparsers):
                 'are found in this image and then measured across all images in the input stack.'
             )
         )
+        spot_finder_group.add_argument(
+            "--codebook", default=None, required=False,
+            help=(
```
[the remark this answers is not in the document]
ttung: ping on this comment.

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -84,17 +88,28 @@ def run(self, stack: ImageStack) -> Tuple[IntensityTable, ConnectedComponentDeco
 
     @classmethod
     def add_arguments(cls, group_parser):
-        group_parser.add_argument("--codebook-input", help="csv file containing a codebook")
+        group_parser.add_argument("--codebook", help="json file containing a codebook")
```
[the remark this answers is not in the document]
ttung: Do you still need this?

starfish/test/full_pipelines/cli/_base_cli_test.py
ttung: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

`Classes can be skipped just like methods:`

shanaxel42: But the subclasses also get skipped if I add that decorator to the whole class

ttung: Ah right, so do this instead:

1. make CLITest _not_ an instance of unittest.TestCase.
2. make each of the concrete test classes in `starfish/test/full_pipelines/cli` extend (CLITest, unittest.TestCase).

starfish/image/_filter/richardson_lucy_deconvolution.py
shanaxel42: Yea well it was something ambrose wanted me to change when I was working on the merfish pipeline and asking him about it. It is kinda out of the cope of this PR though if you want me to make a separate one for it.

starfish/spots/_detector/pixel_spot_detector.py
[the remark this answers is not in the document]
shanaxel42: Do you mean do you need the codebook at all? It get's used in PixelSpotDetector run()

starfish/image/_filter/richardson_lucy_deconvolution.py
ttung: Yup, that would be ideal.

starfish/spots/_detector/pixel_spot_detector.py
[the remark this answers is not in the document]
ttung: Why does the both SpotFinder and PixelSpotDetector need the codebook arguments?

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("experiment_url")
+    parser.add_argument("output_dir", type=FsExistsType())
+    args = parser.parse_args()
+
+    # save image stacks locally
+    exp = Experiment.from_json(args.experiment_url + 'experiment.json')
```
ttung: Sorry I didn't notice this the first time; you should use `pathlib` to do these manipulations:

`str(pathlib.PurePosixPath(args.experiment_url, "experiment.json"))`

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+    for fov in exp.fovs():
+        fov_dir = args.output_dir + "/" + fov.name + "/"
```
ttung:
```
fov_dir = pathlib.Path(args.output_dir, fov.name)
fov_dir.mkdir()
fov.primary_image.write(str(fov_dir / "hybridization.json"))
for image_type in fov.auxiliary_image_types:
    fov[image_type].write(str(fov_dir / f"{image_type}.json"))
```

examples/get_cli_test_data.py
ttung: ping

starfish/image/_filter/__init__.py
```diff
@@ -51,6 +51,5 @@ def _cli(cls, args, print_help=False):
         print('Filtering images ...')
         stack = ImageStack.from_path_or_url(args.input)
         instance = args.filter_algorithm_class(**vars(args))
-        instance.run(stack, in_place=True)
-
-        stack.write(args.output)
+        output = instance.run(stack, in_place=False)
```
ttung: in_place `defaults` to False now.

starfish/spots/_detector/__init__.py
```diff
@@ -26,19 +28,24 @@ def add_to_parser(cls, subparsers):
...
-            "--reference-image-from-max-projection", default=False, action='store_true',
-            help=(
+            "--reference-image-from-max-projection", default=False, action='store_true', help=(
                 'Construct a reference image by max projecting imaging rounds and channels. Spots '
                 'are found in this image and then measured across all images in the input stack.'
             )
         )
+        spot_finder_group.add_argument(
+            "--codebook", default=None, required=False, help=(
+                'A spaceTx spec-compliant json file that describes a three dimensional tensor '
```
ttung: nit: use consistent style (' vs ") for strings.

starfish/spots/_detector/__init__.py
```diff
@@ -61,15 +68,24 @@ def _cli(cls, args, print_help=False):
...
-        instance = args.spot_finder_algorithm_class(**vars(args))
-        intensities = instance.run(
-            image_stack,
-            blobs_image=blobs_image,
-            reference_image_from_max_projection=args.reference_image_from_max_projection
-        )
-        intensities.save(args.output)
+            intensities = instance.run(image_stack)
+
+        # When PixelSpotDetector is used run() returns a tuple
```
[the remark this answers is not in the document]
ttung: https://github.com/spacetx/starfish/issues/637

starfish/spots/_detector/pixel_spot_detector.py
```diff
@@ -48,7 +50,9 @@ def __init__(
         self.crop_y = crop_y
         self.crop_z = crop_z
 
-    def run(self, stack: ImageStack) -> Tuple[IntensityTable, ConnectedComponentDecodingResult]:
+    def run(
+            self, stack: ImageStack,
```
ttung: nit: i think this should be -4 indent, but nbd.

starfish/test/full_pipelines/cli/_base_cli_test.py
ttung: Oh yeah.  You can make it a `@property` method that raises `NotImplementedError()`.

Subclasses will then return what's expected.

starfish/test/full_pipelines/cli/_base_cli_test.py
ttung: Also, consider adding type hints.

examples/get_cli_test_data.py
shanaxel42: What's the advantage of using pathlib to format the urls? When I do this the resulting url is https:///dmf0bdeheu4zf.cloudfront.net/20180919/DARTFISH-TEST/experiment.json which throws an invalid url error because of the extra /

ttung: ok that's unfortunate.  pathlib has a very intriguing interface, but may not work the best for URLs.  `os.posixpath.join` will definitely work though.

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("experiment_url")
+    parser.add_argument("output_dir", type=FsExistsType())
+    args = parser.parse_args()
+
+    # save image stacks locally
+    exp = Experiment.from_json(os.path.join(args.experiment_url, "experiment.json"))
```
ttung: You want `posixpath` in case this gets run on Windows.

examples/get_cli_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+            fov[image_type].write(str(fov_dir / f"{image_type}.json"))
+
+    # get codebook from url and save locally to tmp dir
+    codebook = requests.get(os.path.join(args.experiment_url, "codebook.json"))
+    data = codebook.json()
+    with open(pathlib.Path(args.output_dir, 'codebook.json'), 'w') as f:
+        json.dump(data, f)
+
+    # get experiment.json from url and save locally to tmp dir
+    experiment = requests.get(os.path.join(args.experiment_url, "experiment.json"))
```
ttung: You also want `posixpath` here.

PR #563, opened 2018-09-14
Build consistent files for iss breast data set.

ttung:
tifffile writes files that include a timestamp.  If that timestamp is not provided, it defaults to the current time.  That means every time we run the script, we get slightly different files.  This is undesirable because:
  1. it makes it hard to upload data sets with file format changes.
  2. it pollutes the disk cache

With this change, we now set the timestamp to unixtime 0.

Test plan: regenerated the ISS breast data set, then ran strings on one of the TIFF files.
```
[tt]:~/microscopy/starfish-formatted-data> strings breast/nuclei-fov_007-Z0-H0-C0.tiff | grep 1969
1969:12:31 16:00:00
```

### DIFF
--- a/examples/get_iss_breast_data.py
+++ b/examples/get_iss_breast_data.py
@@ -1,9 +1,11 @@
 import argparse
 import io
 import os
+from datetime import datetime
 from typing import IO, Tuple
 
-from skimage.io import imread, imsave
+from skimage.external import tifffile
+from skimage.io import imread
 from slicedimage import ImageFormat
 
 from starfish.experiment.builder import FetchedTile, TileFetcher, write_experiment_json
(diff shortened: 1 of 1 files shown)


ttung:
> Oh, nice. Easier fix than I thought!

I tried to slurp the original TIFF's timestamp, but that turned out to be awful because they store the timestamp in ASCII (wtf??) and silly multipage TIFF garbage.

> Wouldn't it also make sense to calculate the hash for the file after it's written? Or am I missing something here? I feel like writing the file and updating the metadata should go together.

We do calculate the hash after it's written.

However, to save a ton of time uploading the data set, I copy the old data set to the new data set (on S3), and then use aws sync to update the files that changed from the previous data set.  Unbeknownst to me, aws sync only considers timestamps and file size when doing sync.  Since the timestamp is a fixed size, the different files have the same size.  Timestamp is a bit subjective (do you run `get_iss_breast_data.py` first or do you run the first aws copy first)?  So then we have new metadata files (with new checksums), and tiff files that did not get updated.

This would be solved with a tool that examines checksums rather than (timestamp, file size) when considering sync.

ambrosejcarr:
Lol, got it.

cgohlke:
> they store the timestamp in ASCII (wtf??)

according to [TIFF specification](https://www.awaresystems.be/imaging/tiff/tifftags/datetime.html)

> silly multipage TIFF garbage.

?

PR #566, opened 2018-09-14
Clean up experiment builder.

ttung:
1. Split imagedata.py into providers.py (contracts) and defaultproviders.py (default implementations).
2. Fixed and improved documentation.
3. Add type annotations.

Test Plan: `python examples/get_iss_breast_data.py /Users/ttung/Downloads/starfish_data/mignardi_breast_1/raw/ ../starfish-formatted-data/breast/`


ttung:
I prefixed all the methods that should be hidden with _, and now there's nothing left that's public and undocumented!

PR #567, opened 2018-09-14
Allow TileFetchers to provide the coordinates for a tile.

ttung:
There are some TODOs introduced by this PR that @dganguli will probably need to fill in.

Test plan: `python examples/get_iss_breast_data.py /Users/ttung/Downloads/starfish_data/mignardi_breast_1/raw/ ../starfish-formatted-data/breast/` + visual inspection

Depends on #566

### DIFF
--- a/examples/get_iss_breast_data.py
+++ b/examples/get_iss_breast_data.py
@@ -2,14 +2,14 @@ import argparse
 import io
 import os
 from datetime import datetime
-from typing import IO, Tuple
+from typing import IO, Mapping, Tuple, Union
 
 from skimage.external import tifffile
 from skimage.io import imread
 from slicedimage import ImageFormat
 
 from starfish.experiment.builder import FetchedTile, TileFetcher, write_experiment_json
-from starfish.types import Indices
+from starfish.types import Coordinates, Indices, Number
 from starfish.util.argparse import FsExistsType
 
 SHAPE = 1044, 1390
@@ -22,6 +22,15 @@ class IssCroppedBreastTile(FetchedTile):
     @property
     def shape(self) -> Tuple[int, ...]:
         return SHAPE
+
+    @property
+    def coordinates(self) -> Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]:
+        # FIXME: (dganguli) please provide proper coordinates here.
+        return {
+            Coordinates.X: (0.0, 0.0001),
+            Coordinates.Y: (0.0, 0.0001),
+            Coordinates.Z: (0.0, 0.0001),
+        }
 
     @property
     def format(self) -> ImageFormat:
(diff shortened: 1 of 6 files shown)


examples/get_iss_breast_data.py
```diff
@@ -23,6 +23,15 @@ def __init__(self, file_path):
     def shape(self) -> Tuple[int, ...]:
         return SHAPE
 
+    @property
+    def coordinates(self) -> Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]:
```
ambrosejcarr: Funny, Matt Cai mentioned today that coordinates are the one thing that's not in the filename when delivering single-plane TIFFs. Do we have a plan on how users should provide these by default? 

... a json file mapping filenames to coordinates, perhaps? It'll be important to set up such a default soon.

starfish/experiment/builder/providers.py
```diff
@@ -25,6 +27,17 @@ def shape(self) -> Tuple[int, ...]:
...
         raise NotImplementedError()
 
+    @property
+    def coordinates(self) -> Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]:
+        """Return the tile's coordinates in the global coordinate space..
+
+        Returns
+        -------
+        Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]
+            Maps from a coordinate type to its value or range.
```
ambrosejcarr: I'd make this more expressive. `Maps from a coordinate type (e.g. 'x', 'y', or 'z') to its value or range.`

examples/get_iss_breast_data.py
ttung: > Do we have a plan on how users should provide these by default?

I don't envision that we'll be strongly opinionated about this process.

starfish/experiment/builder/providers.py
ttung: Fixed.

ttung:
> Is there a follow-up PR exposing coordinates in ImageStack and TileSet? :-)

Sort of.  Remember that coordinates are at the _Tile_ level, not TileSet.  Presumably one would have to index (using round/ch/z) to get at the coordinates.

ambrosejcarr:
Ah, right. Perhaps registration methods could expose Coordinates on ImageStacks... we're going to need a way to shift IntensityTable coordinates when we concatenate them across adjacent tiles.

ambrosejcarr:
I guess a follow-up question -- do we think users will ever give tiles in the same fov different coordinates? (should these _be_ fov level attributes?)

ttung:
> I guess a follow-up question -- do we think users will ever give tiles in the same fov different coordinates? (should these be fov level attributes?)

yes, one of the data sets we have (Simone's, I think) has that characteristic.

PR #576, opened 2018-09-16
[easy] Organize stack and experiment so they're less susceptible to circular dependencies

ttung:
Move ImageStack from stack/__init__.py to stack/imagestack.py
Move Experiment from experiment/__init__.py to experiment/experiment.py

Explicitly import them from starfish.stack.imagestack or starfish.experiment.experiment _except_ for public facing code, which imports from starfish.


ambrosejcarr:
Could `ImageStack` just live in `starfish/imagestack.py`? 

> Explicitly import them from starfish.stack.imagestack or starfish.experiment.experiment except for public facing code, which imports from starfish.

This probably belongs in CONTRIBUTING.md, where we should start to enumerate our decisions about the codebase.

ttung:
> Got it, makes sense. One residual request: can we have the root directory named image_stack or imagestack instead of stack? I find the difference unnecessarily misleading.

Done.

PR #577, opened 2018-09-16
[Easy] update conf to correctly point to root directory of repository

ambrosejcarr:
It was pointing at `/docs/` when it needs to point at `/`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -11,10 +11,11 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../..'))
+dir_, _ = os.path.split(__file__)
+root_dir = os.path.abspath(os.path.join(dir_, '..', '..'))
+sys.path.insert(0, root_dir)
 
 
 # -- Project information -----------------------------------------------------


docs/source/conf.py
```diff
@@ -14,7 +14,7 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath('../..'))
```
ttung: this is not a reliable way of doing this.

1. consider getting the abspath of `__file__`, and walk up two directory tree components.
2. (preferable) use pkg_resources to find the root.

ambrosejcarr: Fair. I think this conf.py file is only called by the Makefile, so due to the limited use patterns it's not as fragile as you think. However, can't hurt to make it more robust! 

I wasn't able to figure out how `pkg_resources` improves things, since it sends us to `starfish/starfish/` which is also wrong (we need `starfish/`). Anchored this to `__file__`.

docs/source/conf.py
```diff
@@ -11,10 +11,9 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#
 import os
 import sys
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath(os.path.join(__file__, '..', '..')))
```
ttung: Interesting approach.  I did not realize you could do .. after a file.

ttung: Well, I guess you can, although this produces the wrong result!

```
In [73]: os.path.abspath(os.path.join("/Users/ttung/microscopy/starfish/docs/source/conf.py", "..", ".."))
Out[73]: '/Users/ttung/microscopy/starfish/docs'
```

ambrosejcarr: 🤦‍♂️ 
Fixed.

PR #578, opened 2018-09-17
Make in-place opt-in.

### LINKED ISSUE #384: Make out-of-place computation the default

ttung:
Default is out-of-place.

Depends on #598 

Fixes #384

### DIFF
--- a/notebooks/DARTFISH_Pipeline_-_Human_Occipital_Cortex_-_1_FOV.ipynb
+++ b/notebooks/DARTFISH_Pipeline_-_Human_Occipital_Cortex_-_1_FOV.ipynb
@@ -127,8 +127,8 @@
     "sc_filt = Filter.ScaleByPercentile(p=100)\n",
     "z_filt = Filter.ZeroByChannelMagnitude(thresh=.05, normalize=False)\n",
     "\n",
-    "norm_stack = sc_filt.run(stack, in_place=False)\n",
-    "zero_norm_stack = z_filt.run(norm_stack, in_place=False)"
+    "norm_stack = sc_filt.run(stack)\n",
+    "zero_norm_stack = z_filt.run(norm_stack)"
    ]
   },
   {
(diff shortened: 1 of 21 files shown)


ttung:
> Next step here is that I think this breaks the ISS and allen_smFISH notebooks (since they still assume in_place=True)

Um, are you sure?  I ran make -j run_notebooks and that passed.

ambrosejcarr:
> Um, are you sure? I ran make -j run_notebooks and that passed.

No, but I go look at the notebooks and they're not assigning the outputs of the filters to anything. So, unless the filters don't matter, I don't think the notebooks should be passing? The ISS notebook, for example, has this block of code: 

```python

# filter raw data
masking_radius = 15
filt = Filter.WhiteTophat(masking_radius, is_volume=False)
for img in images:
    filt.run(img, verbose=True)
```

ttung:
> No, but I go look at the notebooks and they're not assigning the outputs of the filters to anything. So, unless the filters don't matter, I don't think the notebooks should be passing? The ISS notebook, for example, has this block of code:

LOL that is amazing.  So either the in-place code is hilariously broken or we don't need filters.

ttung:
Wait, we have zero asserts in the notebooks now... of course they pass. 😠

ttung:
Ok fixed.  It turns out _not_ filtering didn't trigger the small handful of asserts that are in the notebook. hahahaha.

ambrosejcarr:
My understanding of the current testing plan is: 

1. Update .py files
3. Test the .py files (which have really stringent tests!) 
3. Use passing .py files to update notebooks

Is this right @shanaxel42 ?

shanaxel42:
Yea @ambrosejcarr that flow is correct. Once the api test refactor is merged they will catch stuff like this. The tests didn't tigger an error here because they don't currently test on any of the values from the notebooks.

Issue #529, opened 2018-09-07
Update schema $ID fields to reflect their new home in starfish. 


ambrosejcarr:
After migration, the ID fields now point at non-existent github locations, e.g.   
`"$id": "https://github.com/spacetx/sptx-format/schema/codebook.json",`

These changes will subsume spacetx/sptx-format#13

PR #580, opened 2018-09-17
Point schema $ID fields to their new home in starfish

ttung:
Fixes https://github.com/spacetx/starfish/issues/529


ttung:
Removed extra fields as @ambrosejcarr suggested.

Issue #301, opened 2018-07-05
filter by 2d tile vs as a 3d volume


ttung:
see https://github.com/spacetx/starfish/pull/261/files#r196164551

PR #588, opened 2018-09-18
ENH: Generalize apply to group arbitrary axes

ambrosejcarr:
This PR generalizes apply to work for other subsets of ImageStack than just Tiles and Volumes. 

- Swaps `is_volume` for `apply_over`
- `apply_over` is a `Set` of `Indices` that should be used to form sub-tensors by calling `ImageStack.get_slice`

Examples: 
1. Tiles: `{Indices.X, Indices.Y}`
2. Volumes: `{Indices.X, Indices.Y, Indices.Z}`
3. Channels: `{Indices.CH}`

It does not change the API for filters (still uses is_volume). Do we still want to execute the proposed change in #301, or can someone think of a way to leverage this syntax to make something clearer?


starfish/image/_filter/util.py
```diff
@@ -136,3 +136,10 @@ def preserve_float_range(array: Union[xr.DataArray, np.ndarray]) -> Union[xr.Dat
     if np.any(array > 1):
         array /= array.max()
     return array.astype(np.float32)
+
+def determine_axes_to_apply_over(is_volume: bool):
```
ttung: What does it mean to apply over an axis?  Which axes to group over to form something to filter?  The terminology is mostly what I have an issue with.

Also, return type, pls.

ambrosejcarr: `split_by`? That's probably the closest to the split-apply-combine terminology that's common in the python ecosystem

ttung: > split_by? That's probably the closest to the split-apply-combine terminology that's common in the python ecosystem

ok sure that seems reasonable.

also, i think that's perhaps common to the sci-py ecosystem...

ambrosejcarr:
@shanaxel42 Would you be able to help me figure out why the tests are failing? I think this would be really useful to get in, but I'm a bit confused. The changes are pretty small!

PR #592, opened 2018-09-18
travis caching

shanaxel42:
Caching the cache dir between builds.


ttung:
Should we enable the [pip cache](https://docs.travis-ci.com/user/caching/#pip-cache) while we're at it?  May have to explicitly name the path since I would imagine that multiple cache directives would conflict...?

PR #594, opened 2018-09-19
Fix the notebook + pypi pkg test

ttung:
python -m venv on travis produces broken virtualenvs.  This circumvents the issue by invoking virtualenv instead.  This means it doesn't always work perfectly in local dev environments as virtualenv is not a requirement and python3.6 is, but working on travis is more important for this rule.

Test plan: see https://travis-ci.org/spacetx/starfish/builds/430296829

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,12 @@ REQUIREMENTS-DEV.txt : REQUIREMENTS.txt.in
 
 include notebooks/subdir.mk
 
-slow: fast run_notebooks docker
-	python -m venv .notebooks-exec-env
+slow: fast run_notebooks docker pip-notebook-test
+
+# Note that this rule is very specifically tailored to travis's environment, as `python -m venv` on
+# travis does not produce a functional virtualenv.
+pip-notebook-test:
+	virtualenv -p $$(which python) .notebooks-exec-env
 	.notebooks-exec-env/bin/pip install -r REQUIREMENTS-NOTEBOOK.txt
 	.notebooks-exec-env/bin/pip install starfish
 	make PYTHON=.notebooks-exec-env/bin/python run_notebooks


joshmoore:
Doh, you were faster. I was going to propose: https://github.com/spacetx/starfish/compare/master...joshmoore:fix-master?expand=1

I can test this fix in the travis environment, but my question is: can we not get rid of this virtualenv? Would be good to understand the initial requirement. From my perspective, `pip install starfish` in the notebook venv seems to be installing the wrong version.

ttung:
> I can test this fix in the travis environment, but my question is: can we not get rid of this virtualenv? Would be good to understand the initial requirement. From my perspective, pip install starfish in the notebook venv seems to be installing the wrong version.

This is to test that the library on pypi is compatible with the code in the notebooks.  Rationale: we were often getting questions about API changes breaking the notebooks.

joshmoore:
Understood. Can I propose then that:

 * I test this in the `master|merge` setting and assuming happy, merge.
 * In a follow-up, I perform the testing you're looking for in either a _cron_ travis job or in a _tag_ travis job so that we don't block master based on unrelated breakages. ("released")
 * We do the _same_ testing on `master` but with `pip install -e .` ("latest")
 * and once it's ready, we do it on open PRs in the `merge` branch. ("merge")

ttung:
> I test this in the master|merge setting and assuming happy, merge.

What does this entail?

> In a follow-up, I perform the testing you're looking for in either a cron travis job or in a tag travis job so that we don't block master based on unrelated breakages. ("released")

I think this is a good idea.

> We do the same testing on master but with pip install -e . ("latest")

We already do this.  We run the notebooks _twice_ on master.

joshmoore:
> What does this entail?

For the moment, push whatever you'd like to have travis-master-style tested to `merge` that works due to `master|merge) make slow`.

> We already do this. We run the notebooks twice on master.

Gotcha, but as long as it's happening _somewhere_ then we're ok, no? Otherwise, I fear going from green to green states is going to be tricky.

PR #595, opened 2018-09-19
[controversial] Change the coordinates' axes so it ends with a c.

ttung:
i.e., xc, yc, zc.

Rationale: the z coordinate value collides with the z index data.

Test plan: ran travis


ttung:
I am the worst.  This change turns to be theoretically unnecessary, as my sketch for how coordinates would work was pretty wrong (should be a 6-tuple for each round/c/z, and not a 2x2x2).

Options:
1. Keep this in.  This is my preferred option since I think that two enums with equal values is a risky proposition.
2. Back this out.

Thoughts?  @ambrosejcarr @dganguli

Issue #260, opened 2018-06-08
notebook executions when landing on master should parallelize / --keep-going


ttung:

joshmoore:
Agreed that this seems to be covered by gh-596. Looks like the notebooks may even run per PR (see gh-609) which might be bug or feature. Closing.

PR #596, opened 2018-09-19
Travis updates: stages, notifications, etc.

joshmoore:
TBD


joshmoore:
Looks like the next step will be to split the notebook runs into multiple jobs.

joshmoore:
🍏 
screen shot 2018-09-19 at 23 29 55

joshmoore:
Probably worth getting feedback on how disgusting this verbose strategy is before moving forward. The only halfway house I can think of between parallelization and nice, simple travis descriptions is to define sets of notebooks:

 * Notebooks: A-F
 * Notebooks: G-L (etc)

cc: @ttung @ambrosejcarr @shanaxel42

Makefile
```diff
@@ -43,18 +46,19 @@ REQUIREMENTS-DEV.txt : REQUIREMENTS.txt.in
...
 	docker run -ti --rm spacetx/starfish build --fov-count 1 --hybridization-dimensions '{"z": 1}' /tmp/
 
-.PHONY: all fast lint lint-non-init lint-init test mypy refresh_all_requirements slow docker
+install-src:
+	pip install --force-reinstall --upgrade -r REQUIREMENTS-DEV.txt -r REQUIREMENTS-NOTEBOOK.txt
+	pip install -e .
+	pip freeze
+
+install-pypi:
+	pip install -r REQUIREMENTS-NOTEBOOK.txt
```
ttung: the two lines can be combined into

`pip install -r REQUIREMENTS-NOTEBOOK.txt starfish`

Makefile
```diff
@@ -43,18 +46,19 @@ REQUIREMENTS-DEV.txt : REQUIREMENTS.txt.in
...
-	make PYTHON=.notebooks-exec-env/bin/python run_notebooks
+slow: fast run_notebooks docker
 
 docker:
 	docker build -t spacetx/starfish .
 	docker run -ti --rm spacetx/starfish build --fov-count 1 --hybridization-dimensions '{"z": 1}' /tmp/
 
-.PHONY: all fast lint lint-non-init lint-init test mypy refresh_all_requirements slow docker
+install-src:
+	pip install --force-reinstall --upgrade -r REQUIREMENTS-DEV.txt -r REQUIREMENTS-NOTEBOOK.txt
```
ttung: btw, not really sure that we need to install the notebook requirements for "install-src" but it's a matter of semantics since nobody else calls this rule...

ttung:
This looks really great, except, as you said, the very verbose .travis.yml.  I think that's actually ok, but we need some mechanism to flag that a notebook is _not_ in .travis.yml, which is almost certain to happen.

ttung:
I think this also addresses https://github.com/spacetx/starfish/issues/260

ambrosejcarr:
Just dipping in here to say: Awesome job. This is a great improvement.

Makefile
joshmoore: see gh-609

Makefile
joshmoore: see gh-609

PR #600, opened 2018-09-19
Change tile_data_handle to tile_data

ttung:
Instead of returning a file handle to the source, we now just get the tile data and use the new file_format argument to `generate_partition_document` introduced in https://github.com/spacetx/slicedimage/pull/64.  This means we don't have to generate the TIFF file and save to an in-memory file in the experiment/image builders.

This was previously opened as #584 but github helpfully closed that for me.

Test plan: test_iss cli test should cover this case.  separately, re-ran the entire test suite.

### DIFF
--- a/REQUIREMENTS-DEV.txt
+++ b/REQUIREMENTS-DEV.txt
@@ -80,7 +80,7 @@ semantic-version==2.6.0
 showit==1.1.4
 simplegeneric==0.8.1
 six==1.11.0
-slicedimage==0.0.6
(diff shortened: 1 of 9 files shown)


ttung:
> Looks fine. We don't need the complexity of returning a file handle anymore?

That was needed to hack around the fact that we didn't support writing ImageStacks to TIFFs.  We grabbed the file handle from the experiment builder and dumped that out to another file, in essence producing a TIFF.

PR #609, opened 2018-09-20
Travis cleanup (see 596)

### LINKED ISSUE #605: A mechanism to flag when a notebook is in notebooks/py but not in .travis.yml
Should be a makefile rule that just fails.

Followup from https://github.com/spacetx/starfish/pull/596

joshmoore:
* add test (`make check_notebooks`) for untested notebooks
 * add help to makefile
 * add check_requirements
 * re-simplify .travis.yml
 * implement other suggestions of gh-596

Testing plan:
 * check `make help` is pleasing to the eye
 * add a `new notebooks/py/*.py` and check that `make check_notebooks` fails
 * review the fact that PRs *don't* run notebooks (a subset?)
 * comment on the checks that should happen in cron (like `check_requirements`)
   - include docs?

closes #605


joshmoore:
Temporary test commit fails with:

```
0.02s$ make check_notebooks
grep notebooks/py/allen_smFISH.py .travis.yml
make: *** [check__notebooks/py/allen_smFISH.py] Error 1
```

if a file matching `notebooks/py/*.py` isn't greppable in `.travis.yml`

joshmoore:
help output:

```
$ make help
Main starfish make targets:

    help                        print this text
    all                         alias for fast
    docs-TASK                   alias for 'make TASK' in the docs subdirectory
    refresh_all_requirements    regenerate REQUIREMENTS files
    slow                        alias for 'fast run_notebooks docker'
    run_notebooks               run all files matching 'notebooks/py/*.py'
    docker                      build docker and run a simple container
    install-src                 pip install the current directory
    install-pypi                pip install starfish from pypi
    install-travis              chooses between src and pypi based on TRAVIS_EVENT_TYPE
```

Hopefully the `install-travis` target isn't too invasive.

joshmoore:
Also building the `(master|merge)` variant of .travis.yml under https://travis-ci.org/spacetx/starfish/builds/430901778

joshmoore:
430901778 didn't run the notebooks. Last commit fixes that and is tested by https://travis-ci.org/spacetx/starfish/builds/430904316

joshmoore:
@ttung : I think  https://github.com/spacetx/starfish/pull/609/commits/ba503208673859d20b468a39da0f4b10b27265d1 leads to:
```
[ -e notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py.skip ] || python notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
100%|███████████████████████████████████████████| 16/16 [00:00<00:00, 18.61it/s]
Traceback (most recent call last):
  File "notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py", line 46, in <module>
    primary_image.show_stack({Indices.CH: 0})
  File "/home/travis/build/spacetx/starfish/starfish/imagestack/imagestack.py", line 408, in show_stack
    color_map
  File "/home/travis/build/spacetx/starfish/starfish/imagestack/imagestack.py", line 487, in _show_matplotlib_inline
    from ipywidgets import interact
ModuleNotFoundError: No module named 'ipywidgets'
make: *** [run__notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py] Error 1
```
(in 43090416) Please advise.

ttung:
> @ttung : I think ba50320 leads to:

Ah yeah, that's right.  Ok just leave it.  I just thought the name of the rule (install-src) didn't jive with the packages it was installing (hence my comment about it being a matter of semantics...).

joshmoore:
Ok. I'll rollback. The other name I had thought of was `install-dev`. Open to alternatives.

joshmoore:
`merge` went green with the revert in https://travis-ci.org/spacetx/starfish/builds/430914354

joshmoore:
Not pushing anything here re: gh-586. Might be better to get this in separately.

joshmoore:
https://travis-ci.org/spacetx/starfish/builds/431434815 green again after fixing for requirements the problem that @ttung fixed for notebooks in gh-594.

Happy to take comments.

joshmoore:
@ttung : with validation now turned on, it likely makes sense to expedite at least the .travis.yml changes here to speed up PR testing (_sans_ notebooks). If there's anything here that's worrisome, I'll split it into multiple PRs.

Makefile
```diff
@@ -52,13 +100,50 @@ docker:
...
-	pip install -r REQUIREMENTS-NOTEBOOK.txt
-	pip install starfish
-
-.PHONY: all fast lint lint-non-init lint-init test mypy refresh_all_requirements slow docker install-dev install-pypi
+	pip install -r REQUIREMENTS-NOTEBOOK.txt starfish
+
+install-travis: install-travis-$(TRAVIS_EVENT_TYPE)
+
+install-travis-cron: install-pypi
+install-travis-: install-src
```
ttung: What is this for?

.travis.yml
```diff
@@ -9,67 +9,44 @@ stages:
...
     - name: Released Notebooks
-      if: (branch =~ ^master OR branch =~ ^merge) AND tag = cron
-      script:
-        - make install-pypi
-        - make run_notebooks
+      if: type = cron and branch =~ /^(master|merge)/
+      script: make check_notebooks install-travis run_notebooks
+    - name: Check Requirements
+      if: type = cron AND branch = /^master$/
+      script: make refresh_all_requirements check_requirements
```
ttung: This particular test will break all the time.  What's the expected frequency of this cron?

.travis.yml
```diff
@@ -9,67 +9,44 @@ stages:
...
+      if: type = push and branch =~ /^(master|merge)/
+      script: make docker
     - name: Allen Notebook
-      if: (branch =~ ^master OR branch =~ ^merge) AND tag != cron
-      env:
-      script:
-        - make install-src
-        - make run__notebooks/py/allen_smFISH.py
+      if: type = push and branch =~ /^(master|merge)/
+      script: make install-travis run__notebooks/py/allen_smFISH.py
```
ttung: why not just do `make install-src <notebook>`?

Makefile
joshmoore: For when `TRAVIS_EVENT_TYPE` is not defined.

.travis.yml
joshmoore: At the moment, it's set to weekly.

.travis.yml
joshmoore: If the TRAVIS_EVENT_TYPE logic is also in the job definition, then things get duplicated.

Based on "_I would prefer that the make commands in .travis.yml invoke the explicit rule that it needs (install-src vs install-pypi)_", happy to expand that out, but the above becomes:

```
      if: type = push and branch =~ /^(master|merge)/ and type = cron
      script: make install-pypi run__notebooks/py/allen_smFISH.py
      if: type = push and branch =~ /^(master|merge)/ and type != cron
      script: make install-src run__notebooks/py/allen_smFISH.py
```

which felt error prone.

joshmoore: Sorry, I take that back. In this case since `type = push` is already listed, yes, I can change from travis to the specific version. The other cases look more like what I pasted here. I used `install-travis` everywhere for consistency. To be clear:

 * You want specific targets where possible.
 * You'd like `install-travis` to be removed completely?

ttung: Doesn't every stage have a if: type = xxx?  Or am I completely oblivious?

I think that `install-travis` makes things less clear, so I would like to get rid of that.  If there's a case for multiple targets to be invoked from one stage, pls let me know.

joshmoore: Ditched it. Merge build is running in https://travis-ci.org/spacetx/starfish/builds/433298975

notebooks/subdir.mk
```diff
@@ -1,20 +1,25 @@
...
+check_notebooks: $(py_check_targets)
 validate_notebooks: $(ipynb_validate_targets)
 regenerate_ipynb: $(ipynb_regenerate_targets)
 regenerate_py: $(py_regenerate_targets)
 
 $(py_run_targets): run__%.py :
 	[ -e $*.py.skip ] || $(PYTHON) $*.py
 
+$(py_check_targets): check__%.py :
+	grep $*.py .travis.yml
```
ttung: Consider `grep -q` since otherwise we'll be spewing a bunch of stuff to the console.

.travis.yml
ttung: hm... let's see how it goes and we can tweak how often we want to run it. :)

notebooks/subdir.mk
joshmoore: Pushed.

.travis.yml
joshmoore: :) Agreed. Two options would be to make it a soft-fail so by visiting travis you can see it's red and needed, but the overall build would be green. Alternatively, the build could open the PR with the bump instead of just complaining. Or just remove this. Whatever turns out to be useful and least annoying.

joshmoore:
Travis has gone green with the minor grep fix. Landing this so we can iterate on top of it.

PR #612, opened 2018-09-21
Save the coordinates from the tiles to an xarray during ImageStack construction

### LINKED ISSUE #560: Starfish should expose all metadata in the spaceTx-format
Since I, the user, have constructed this data format metadata, I expect to be able to retrieve it from the appropriate starfish object. 

Ideally, the UX should be such that the name of the field in the format should be the same as the name of the property in the object?

ttung:
Test plan: Added a test to verify that the coordinates are saved and retrieved correctly.

Depends on #610, #586, https://github.com/spacetx/slicedimage/pull/66

Fixes #560

### DIFF
--- a/REQUIREMENTS-DEV.txt
+++ b/REQUIREMENTS-DEV.txt
@@ -81,7 +81,7 @@ semantic-version==2.6.0
 showit==1.1.4
 simplegeneric==0.8.1
 six==1.11.0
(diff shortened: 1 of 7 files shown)


ttung:
This won't pass CI until a new pip pkg is released for slicedimage.

PR #619, opened 2018-09-24
Point notebooks at data sets without the 'fov' fields in TileSets.

ttung:
The fov field was removed in https://github.com/spacetx/sptx-format/pull/24, but some of the data sets still have it.

Test plan: make -j run_notebooks


ttung:
Updated the assay_comparsion notebook...

Issue #590, opened 2018-09-18
MeanHighPassFilter needs to be modernized post-floats


dganguli:
The following code in high_pass is no longer needed:

```
if image.dtype != np.uint16:
    DataFormatWarning( "Mean filters currently only support uint16 images. 
      Image data will be converted.")
image = img_as_uint(image)
```

FYI @kevinyamauchi

kevinyamauchi:
Awesome! Should I take a crack at it to make sure I understand how types/conversions work now?

ambrosejcarr:
@kevinyamauchi Did you get the information you needed from @dganguli post-sync yesterday? If not lets chat -- This would be great for you to contribute. 👍

kevinyamauchi:
I think @dganguli and I are meeting today. I'll check in with him about it then.

kevinyamauchi:
Fixed by #620

PR #620, opened 2018-09-24
Modernize MeanHighPass

kevinyamauchi:
As requested in #590 , this PR modernizes the mean high pass filter to reflect the change to floats for all image data in the starfish pipeline.
* removes type check
* uses new kernal shape check function
* updated docstrings

### DIFF
--- a/starfish/image/_filter/mean_high_pass.py
+++ b/starfish/image/_filter/mean_high_pass.py
@@ -3,7 +3,6 @@ from functools import partial
 from typing import Callable, Optional, Tuple, Union
 
 import numpy as np
-import xarray as xr
 from scipy.ndimage.filters import uniform_filter
 
 from starfish.imagestack.imagestack import ImageStack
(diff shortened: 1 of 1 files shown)


starfish/image/_filter/mean_high_pass.py
```diff
@@ -58,27 +49,25 @@ def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
             help="indicates that the image stack should be filtered in 3d")
 
     @staticmethod
-    def high_pass(image: np.ndarray, size: Number) -> np.ndarray:
+    def high_pass(
+            image: Union[xr.DataArray, np.ndarray], size: Union[Number, Tuple[Number]]
+    ) -> Union[xr.DataArray, np.ndarray]:
```
ttung: The output type for this method appears to only be `nd.ndarray`.  I can't see any paths that would yield a `xr.DataArray`.

kevinyamauchi: Good point. Fixed.

kevinyamauchi:
@ttung, can you take another look when you get a chance?

ttung:
For future reference, you can use this dropdown, click my name, and it'll put it back in my queue.

screen shot 2018-09-25 at 10 37 02 pm

starfish/image/_filter/mean_high_pass.py
```diff
@@ -60,28 +50,24 @@ def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
     @staticmethod
     def high_pass(image: np.ndarray, size: Number, rescale: bool=False) -> np.ndarray:
```
ttung: you could legitimately accept `xr.DataArray`, I *think*, but I'm not 100% sure.  it will just return `np.ndarray`

ttung:
also, ideally we should write tests for this...

kevinyamauchi:
> also, ideally we should write tests for this...

Good point. I'll add an issue and try and work on it later this week.

PR #621, opened 2018-09-24
Fix image type Filter docstrings

kevinyamauchi:
Some of the Filter classes have unclear or potentially wrong types in the docstring for the input image in the filtering method (e.g., high_pass() in GaussianHighPass). I changed the GHP and GLP filters to np.float32 based on my current understanding of the pipeline and the preserve_float_range() calls.

In going through the filters, I noticed that some (e.g., Clip) still have code to support int. Should filters still support int even though the pipeline internal standard is float? I can make a separate issue if this requires discussion.

### DIFF
--- a/starfish/image/_filter/gaussian_high_pass.py
+++ b/starfish/image/_filter/gaussian_high_pass.py
@@ -48,7 +48,7 @@ class GaussianHighPass(FilterAlgorithmBase):
 
         Parameters
         ----------
-        image : numpy.ndarray[np.uint16]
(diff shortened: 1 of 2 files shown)


ttung:
> In going through the filters, I noticed that some (e.g., Clip) still have code to support int. Should filters still support int even though the pipeline internal standard is float? I can make a separate issue if this requires discussion.

Yes, we should fix that.

ttung:
Looks like only the allen_smFISH test exercises that code path, but it doesn't validate the output.

PR #626, opened 2018-09-24
Adding data module

### LINKED ISSUE #430: Update `get_iss_data` and add similar fetching functions for other supported assays
We should have a way of rapidly pulling in small amounts of testing data for each of the assays to support the creation of unit tests. 

ACs: 

- [x] - subset each assay's testing data so that it can be loaded and fully processed in < 30s while retaining a biological outcome: 
- [x] - image contains at least two cells
- [x] - location selected contains a biological feature that is visible
- [x] - put data on s3 in a cloudfront `test` directory
- [ ] - create a sub-package `data` that contains the following methods: 
  - `osmFISH()`
  - `ISS_breast()`
  - `ISS_culture()`
  - `MERFISH()`
  - `allen_smFISH()`
  - `DARTFISH()`
- [ ] - each method should generate an `Experiment` object that can be used for testing, backed by the s3 cloudfront link.

shanaxel42:
New data module that makes it easy to generate an experiment object for each dataset. 

Usage:

from starfish import data
iss_exp = data.ISS()

fixes #430


joshmoore:
Nice! I started on something like this to wrap the `from_json` calls with validation before moving to the `STARFISH_STRICT_LOADING` variable. I could see have a helper method here:

```
def load(name, url="https://dmf0bdeheu4zf.cloudfront.net/", version="20180924", strict=False):
    return Experiment.from_json(
        f'{url}{version}/{name}/experiment.json', strict=strict)
```

so that `version` becomes one of the very few places that needs updating when new data is issued.

Is the idea to also use these methods from within notebooks or just unit tests?

starfish/data/__init__.py
```diff
@@ -0,0 +1,22 @@
+from starfish import Experiment
+
+
+def MERFISH():
```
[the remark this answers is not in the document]
shanaxel42: GOING FOR IT

shanaxel42:
> Nice! I started on something like this to wrap the `from_json` calls with validation before moving to the `STARFISH_STRICT_LOADING` variable. I could see have a helper method here:
> 
> ```
> def load(name, url="https://dmf0bdeheu4zf.cloudfront.net/", version="20180924", strict=False):
>     return Experiment.from_json(
>         f'{url}{version}/{name}/experiment.json', strict=strict)
> ```
> so that `version` becomes one of the very few places that needs updating when new data is issued.
> 
> Is the idea to also use these methods from within notebooks or just unit tests?

I think @ambrosejcarr  was intending  for them to get used by people who just wan to quickly get and play around with the different pipelines. But I updated the notebooks to use them as well!

starfish/data/__init__.py
```diff
@@ -0,0 +1,35 @@
+from starfish import Experiment
+
+
+def MERFISH(test_data: bool=False):
```
ambrosejcarr: +1 to @ttung comment that we should make this `use_test_data`. 

Also made #648

ttung: https://github.com/spacetx/starfish/pull/651

PR #629, opened 2018-09-25
ISS CLI script is broken.

ttung:
This fixes it.

Test plan: `. docs/source/usage/iss/iss_cli.sh` in starfish folder.


joshmoore:
Does this need regular testing?

PR #631, opened 2018-09-25
Changing clip default to true in deconvolution

shanaxel42:


starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -45,14 +45,14 @@ def add_arguments(cls, group_parser: argparse.ArgumentParser) -> None:
...
-            '--clip', action='store_true',
-            help='(default False) if True, clip values below -1 and above 1')
+            '--no-clip', action='store_false',
+            help='(default True) if True, clip values below 0 and above 1')
 
     # Here be dragons. This algorithm had a bug, but the results looked nice. Now we've "fixed" it
     # and the results look bad. #548 addresses this problem.
     @staticmethod
     def richardson_lucy_deconv(
-            image: np.ndarray, iterations: int, psf: np.ndarray, clip: bool=False) -> np.ndarray:
```
ttung: Since this method is basically internal to this class, I think we should just make the clip parameter required (i.e., no default value).

ttung: Also, docblock below needs to be updated.

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -65,8 +65,8 @@ def richardson_lucy_deconv(
         iterations : int
            Number of iterations. This parameter plays the role of
            regularisation.
-        clip : boolean, optional
-           True by default. If true, pixel value of the result above 1 or
+        clip : boolean
+            If true, pixel value of the result above 1 or
            under -1 are thresholded for skimage pipeline compatibility.
```
ambrosejcarr: It's [0, 1] not [-1, 1] now.

PR #583, opened 2018-09-18
Refactoring api tests to use notebook data

shanaxel42:
The api test now load and run their corresponding notebooks with the test data then check the intermediate variables.

PR #633, opened 2018-09-25
Regenerated jupyter notebooks from .py files

ttung:
#596 changed it such that travis only runs `make fast` instead of `make all`, which meant we were not validating that the jupyter notebooks stayed in sync.  #583 changed the .py file without updating the .ipynb file.

Test plan: `make validate_notebooks` passes.


joshmoore:
Hmmm... apologies for losing that, but I'm confused. Doesn't `all` just depend on `fast` which in turn depends on `lint mypy test`?

ttung:
ummm, yes, you are right.

however, `validate_notebooks` is hooked onto all, and not `fast`.  I guess we could switch it to `fast` since it is pretty fast.

joshmoore:
Ah, ok. I see via the `all:` in subdirs.mk. Up to you if you want fast to have validate_notebooks or call all from `Syntax and Tests` _or_ Add another job which is just for notebook validation. I don't have a feeling for relative times.

ttung:
Note #634.  That solves the issue. :)

joshmoore:
New I should have put a `#deprecated` on all. :) Ok. Landing that and dealing with the conflict. Open to just killing `fast` then.

PR #635, opened 2018-09-25
Fix the list of dimensions written to the TileSet metadata

ttung:
Fix the list of dimensions written to the TileSet metadata

It should be the set of all physical and non-physical dimensions.

Test plan: make -j test


ttung:
There is a small chance that this is controversial, so I'm expanding the reviewer set. :)

PR #641, opened 2018-09-26
[CODEBOOK 3/3] Versioned codebook.

### LINKED ISSUE #530: Codebook needs a version field
Finish https://github.com/spacetx/sptx-format/pull/11

ttung:
1. Codebook (on disk) is now a dictionary.  It has a "version" field and a "mappings" field.
2. Updated schema to match this change.
3. Removed test_from_json.py as it is a pretty low value test.
4. Updated notebooks to pull new data sets.

Test plan: make -j run_notebooks fast

Fixes #530
Depends on #639, #640, #622


shanaxel42:
Is there a benefit in having experiment and codebook each keep their own separate version? It might be easier to keep them aligned and just have one "spacetx" version.

shanaxel42:
Makes sense!

starfish/codebook/codebook.py
```diff
@@ -307,8 +323,17 @@ def from_json(
         """
         backend, name, _ = resolve_path_or_url(json_codebook)
         with backend.read_contextmanager(name) as fh:
-            code_array = json.load(fh)
-        return cls.from_code_array(code_array, n_round, n_ch)
+            codebook_doc = json.load(fh)
+
+        if isinstance(codebook_doc, list):
+            raise ValueError(
+                f"codebook is a list and not an array.  It is highly likely that you are using a "
```
ambrosejcarr: should this be "codebook is a list and not a dictionary"?

starfish/codebook/codebook.py
```diff
@@ -126,6 +133,15 @@ def _empty_codebook(cls, code_names: Sequence[str], n_ch: int, n_round: int):
...
+    def _verify_version(cls, semantic_version_str: str) -> None:
+        version = Version(semantic_version_str)
+        if not (MIN_SUPPORTED_VERSION <= version <= MAX_SUPPORTED_VERSION):
+            raise ValueError(
+                f"version {version} not supported.  This version of the starfish library only "
+                f"supports codebook formats from {MIN_SUPPORTED_VERSION} to "
+                f"{MAX_SUPPORTED_VERSION}")
+
     @classmethod
     def from_code_array(
```
ambrosejcarr: I intended this constructor to slurp in an already-in-memory codebook. 

I feel like it could be cleaner to update it to demand a Dict that includes a correct version parameter and do the verification, as otherwise this method is unaware of any changes to the codeword fields of the codebook.

ttung: I disagree.  Managing the format should be the responsibility of this class.  If you want something that deals with a parsed document, it should 1. be internal to this class (i.e., something prefixed with _), and 2. not be named `from_code_array` (maybe `_from_codebook_document`).

starfish/codebook/codebook.py
ttung: Fixing.

starfish/codebook/codebook.py
ambrosejcarr: Ok, please update the example to be consistent with the new spec (I think just "codebook -> mappings"?)

ttung: > Ok, please update the example to be consistent with the new spec (I think just "codebook -> mappings"?)

um where?

ttung: I landed.  I will fix whatever comments you're referring to in a subsequent PR.

PR #645, opened 2018-09-27
Allow setting of allow_caching

joshmoore:
For testing the performance of the S3 storage, I'd like to be able to disable caching.  (Note: just setting `~/.starfish-cache` to `000` doesn't suffice. DiskCache doesn't fail gracefully) If this is a generally agreed upon strategy, I'll propagate it to all the other `from_json` methods. If not, then I'll look into otherwise configuring the backend, e.g. by passing the maximum size to 0.

----

## starfish.util.config.Config object ##

The loading mechanism for configs is influenced by Ansible's `@filename` syntax:

```
ansible-playbook release.yml --extra-vars "@some_file.json"
```

but omits any YAML-ness. The `Config` constructor will accept one of:

 * a json-like object
 * a json-like string
 * a @filename containing json data

from either:
 * the value passed to the constructor, or
 * an environment variable ("STARFISH_CONFIG" by default)

The `Config` object itself is a `UserDict` and so can be used like any dictionary. Additionally a null-safe `lookup()`, name and signature open to discussion, can be used for reaching deep in to the higher.

One suggestion for the data layout _within_ the config json is to follow roughly Django's style (which itself uses DiskCache heavily). See https://docs.djangoproject.com/en/2.1/topics/cache/. A potential example would be:

```
{
    "cache":
    {
        "backend": ...,
        "location": "~/.starfish-cache",
        "timeout": 0
    }
}
```

(There are a _number_ of packages like this that could be used instead of spinning our own.)

----
Open questions:
 * [ ] propagate a variable to all `starfish` CLI commands
 * [ ] propagate the `config` variable through all the `from_json` methods?
 * [ ] ... through other starfish library methods?
 * [ ] ... into slicedimage? (barring use of the "lookup" method)


joshmoore:
No, thanks for the comment. Definitely want to hear preferred design upfront.

joshmoore:
@ttung : pushed another concept piece. Feedback welcome.

joshmoore:
(NB I may be offline by the time anyone gets a chance to look at this. Feel free to close/steal the commits or PR and run with it)

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+from typing import Any, Dict, Sequence, Union
+
+
+class Config(collections.UserDict):
+
+    def __init__(self, value: Union[str, Dict]=None,
+                 key: str="STARFISH_CONFIG")->None:
+        """
+        Parse user-arguments, environment variables, and
```
ttung: s/user-arguments/user arguments

ttung: also, the project guidelines wraps at 100col.  you can probably unwrap a bit.

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
...
+        value: Union[str, Dict]
+            Either a json-object-like structure which will be passed
+            unmodified to the Config constructor, or a string which
+            will be used to build such a structure. If the string
+            starts with an "@", then it will be interpreted as the
+            filename of a json file which should be read. Otherwise,
+            it will be parsed as a json string.
+        key: str
+            The name of an environment variable which will be used
+            if the value is None.
```
ttung: should mention `STARFISH_CONFIG`

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
...
+                with open(value[1:], "r") as o:
+                    data = json.loads(o.read())
+            else:
+                data = json.loads(value)
+        else:
+            data = value
+
+        collections.UserDict.__init__(self, data)
+
+    def lookup(self, keys: Sequence[str], value: Any=None)->Any:
```
ttung: This needs documentation.

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
...
+        data: Any = self.data
+        for key in keys:
+            try:
+                data = data.get(key)
+            except Exception:
```
ttung: Should be:

`except KeyError`

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
+import collections
+import json
+import os
+from typing import Any, Dict, Sequence, Union
+
+
+class Config(collections.UserDict):
```
ttung: I'm on the fence whether being a UserDict is good or not...

starfish/util/config.py
```diff
@@ -0,0 +1,60 @@
...
+
+        collections.UserDict.__init__(self, data)
+
+    def lookup(self, keys: Sequence[str], value: Any=None)->Any:
+        data: Any = self.data
+        for key in keys:
+            try:
+                data = data.get(key)
+            except Exception:
+                return value
```
ttung: I think that if a default value is not specified, it should raise `KeyError`.  This can be done by having the default value set to a token object.  If value == token object, then a default was not provided.

starfish/util/config.py
joshmoore: data could also be None though.

starfish/util/config.py
joshmoore: :+1:

starfish/util/config.py
joshmoore: Other options you're considering?

ttung: I dunno, I would be okay with just invoking `Config.lookup` for everything.

starfish/util/config.py
ttung: Ahh. Right.  So maybe `except (AttributeError, KeyError)` then. :)

starfish/util/config.py
joshmoore: Other changes pushed. Time to decide on this.

joshmoore: Pushed a commit to look at no `UserDict`

starfish/util/config.py
```diff
@@ -0,0 +1,77 @@
...
+                with open(value[1:], "r") as o:
+                    data = json.loads(o.read())
+            else:
+                data = json.loads(value)
+        else:
+            data = value
+
+        self.data = data
+
+    def lookup(self, keys: Sequence[str], value: Any=__NO_VALUE_PASSED)->Any:
```
ambrosejcarr: spaces around `->`

joshmoore: Doh. Pushed. `$ git grep -E "\S[-][>]\S"` now happily shows only PDFs. Perhaps we add that to `make lint` ... :smile:

joshmoore:
Green again. Merging. Happy to iterate as necessary once folks start using it in anger.

Issue #614, opened 2018-09-21
Experiment.from_json() on a "single-fov experiment" yields uninformative error


ambrosejcarr:
Calling from_json() on an `experiment.json` where the `primary_images.json` is directly linked to the `experiment.json` instead of through a `manifest.json` intermediary yields the following uninformative error message: 

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-fe0a85d3077c> in <module>()
      3 
      4 experiment_json = os.path.expanduser('~/Downloads/BaristaSeq/files_and_metadata/experiment.json')
----> 5 Experiment.from_json(experiment_json)

~/projects/spacetx/starfish/starfish/experiment/experiment.py in from_json(cls, json_url)
    185 
    186         fovs: MutableSequence[FieldOfView] = list()
--> 187         for fov_name, primary_tileset in primary_image.all_tilesets():
    188             aux_image_tilesets_for_fov: MutableMapping[str, TileSet] = dict()
    189             for aux_image_type, aux_image_collection in auxiliary_images.items():

AttributeError: 'TileSet' object has no attribute 'all_tilesets'
```

We should: 
- [ ] validate this aspect of experiment 
- [ ] throw an error if this validation fails. 

Related to #524

joshmoore:
I think the strict setting of gh-538 would have slightly improved this UX, since before this exception was thrown a warning would have been printed with the specific problem. To improve on that one would need to:

 * add a lazy-loading option for strict validation such that data isn't pre-loade (too slow)
 * perhaps make lazy-loading strict-validation the default (no user action needed)
 * modify validate_sptx to not (always) use `warnings` but instead throw friendly exceptions

Issue #646, opened 2018-09-27
TileSet code path still exists


joshmoore:
spacetx-fov-writer wasn't updated to use a manifest file:

```
{
    "version": "4.0.0",
    "auxiliary_images": { },
    "extras": { },
    "primary_images": "hybridization-fov_000.json",
```

Loading its output via:

```
from starfish.experiment.experiment import Experiment
from starfish.types import Indices
experiment = Experiment.from_json(
    "http://nginx:9000/public/fov_000/experiment.json",
    strict=True)
image = experiment.fov().primary_image
image.show_stack({Indices.CH: 0})
```

led to:

```
Traceback (most recent call last):
  File "/tmp/example.py", line 9, in <module>
    strict=True)  # soon: allow_caching=False
  File "/src/starfish/experiment/experiment.py", line 207, in from_json
    for fov_name, primary_tileset in primary_image.all_tilesets():
AttributeError: 'TileSet' object has no attribute 'all_tilesets'
```

joshmoore:
Dupe of gh-614.

PR #647, opened 2018-09-27
Restrict 'primary_images' to manifests

joshmoore:
`primary_images` was a string matching "*.json" which continued to work whether collection or tileset. By updating ~the schema only a manifest will be permitted~ updating the validator, all linked files will be assumed to be manifests. If they do not pass validation, then fov-validating will be short-circuited.

closes gh-614
seealso gh-646


validate_sptx/schema/experiment.json
```diff
@@ -14,10 +14,8 @@
       "$ref": "schema/version.json"
     },
     "primary_images": {
-      "type": "string",
-      "description": "Images that measure transcript or protein targets in one or more fields of view.",
-      "example": "primary_images.json",
-      "pattern": "^.*?.json$"
+      "$ref": "schema/fov_manifest.json",
```
ambrosejcarr: This still does require a match of `"pattern": "^.*?.json$"`, does it not? It's just that it needs to be more specific? 

And, shouldn't the description string refer to a manifest? 

`"description": "Manifest specifying the location(s) of one or more Field of View sub-manifests"`

Also I don't see code here to make things more strict, yet.

ambrosejcarr:
... I wrote a comment and it was overwritten by the time I hit enter.

validate_sptx/examples/experiment/field_of_view.json
```diff
@@ -0,0 +1,59 @@
...
+      "file": "fdaj3902.tiff",
+      "tile_format": "TIFF",
+      "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+      "extras": {
+        "are_we_missing_something":"probably"
+      }
+    },
+    {
+      "coordinates": {
+        "xc": [0.0, 0.1],
```
ambrosejcarr: I think our codebase should stop perpetuating the idea that coordinates don't matter :-) Increment either `x` or `y` to start at `0.1`?

validate_sptx/schema/experiment.json
joshmoore: I have as yet not found a way to do what I think you and @ttung actually want which is to recursively validate _linked files_. For the moment, I've just dropped the `possible_manifests` logic out of the validator. The next step would be to have validator discover (or parse out of the schema) what to expect on a file by file basis, but I don't think anything is built into jsonschema for that. The alternative would be to go to monolithic schemas & instance documents.

validate_sptx/examples/experiment/field_of_view.json
joshmoore: This is copied from `validate_sptx/examples/field_of_view` but happy to do whatever's appropriate. (Possibly a next PR?)

joshmoore: "This is copied". What I _probably_ should do here is to make this a symlink and then I won't be propagating but only maintaining the status quo. (If symlinks are determined copacetic in the repo.)

ambrosejcarr: cc @ttung w.r.t. symlinks. 

> This is copied from validate_sptx/examples/field_of_view but happy to do whatever's appropriate. (Possibly a next PR?)

Makes sense, this is a bad habit we (I?) introduced back when we were exclusively dealing with single fields of view. Now that we are more aware of multiple-fov experiments our examples should match expectations. 

Down the road, I could see Starfish throwing a validation error when given two "fields of view" that are in the exact same coordinate space. That seems like it violates our expectations about what a field of view is. 

... slippery slope, since we allow overlapping fovs, so two with 10% offsets should probably be OK. However, I think the example as posed is definitely wrong.

joshmoore: I don't think I will be able to get back to this PR within the next. I'd propose landing it (and if I'm offline, please feel free to do that) and I can look into cleaning up the validation tests as a follow-on.

validate_sptx/examples/experiment/codebook.json
```diff
@@ -0,0 +1,24 @@
+[
```
ttung: I don't understand the difference between this and validate_sptx/examples/codebook/codebook.json.

validate_sptx/examples/experiment/field_of_view.json
```diff
@@ -0,0 +1,59 @@
+{
```
ttung: i don't understand the difference between this and validate_sptx/examples/field_of_view/field_of_view.json

validate_sptx/examples/experiment/no_manifest.json
```diff
@@ -0,0 +1,10 @@
+{
```
ttung: would suggest naming this no_fov_manifest, as manifest is one of those highly overloaded terms.

validate_sptx/schema/fov_manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://github.com/spacetx/starfish/sptx-format/schema/codebook.json",
+  "$id": "https://github.com/spacetx/starfish/sptx-format/schema/fov-manifest.json",
```
ttung: 👍

validate_sptx/test_experiment.py
```diff
@@ -30,3 +33,17 @@ def test_dartfish_example_experiment():
...
         "validate_sptx", "examples/experiment/dartfish_experiment.json")
     assert validator.validate_file(dartfish_example)
+
+
+# see #614
+def test_no_manifest_example_experiment():
+    no_manifest_example = resource_filename(
+        "validate_sptx", "examples/experiment/no_manifest.json")
+
+    # SpaceTxValidator doesn't handle multiple files so this passes
```
ttung: How much work do you think it would be to make SpaceTxValidator validate like the tree walker?

validate_sptx/validate_sptx.py
```diff
@@ -54,44 +54,38 @@ def validate(experiment_json: str, fuzz: bool=False) -> bool:
...
-                    fovs.append((json.load(fh), fov))
-
-        else:  # manifest is a field of view
-            fovs.append((manifest, None))
-
-    # validate fovs
-    assert len(fovs) != 0
-    fov_validator = SpaceTxValidator(_get_absolute_schema_path('field_of_view/field_of_view.json'))
-    for fov, filename in fovs:
+    for manifest, filename in manifests:
```
ttung: shouldn't this validate against fov_manifest.json?

validate_sptx/examples/experiment/codebook.json
joshmoore: None, but I assume tests should refer only to files in a given directory (to prevent unintended breakages). Happy with any of:

 * re-use from other directories
 * symlink to other directories
 * copy but use more explicit naming

validate_sptx/examples/experiment/field_of_view.json
joshmoore: As above.

validate_sptx/examples/experiment/no_manifest.json
joshmoore: Very fair. Will push shortly.

validate_sptx/validate_sptx.py
joshmoore: Hmmm.... I might be missing something, but that's what I thought it was doing:

```
    manifest_validator = SpaceTxValidator(_get_absolute_schema_path('fov_manifest.json'))
...
    for manifest, filename in manifests:
...
            if not manifest_validator.validate_object(manifest, filename):
                valid = False
```

joshmoore:
Ah, I just saw @ambrosejcarr's https://github.com/spacetx/starfish/pull/693/files#diff-33eb96e0055240545cba4a7547f71b85L173 for how to store environment variables. I'll push a quick commit.

validate_sptx/examples/experiment/field_of_view.json
joshmoore: 🥂 Marking this as resolved, happy to add appropriate validation if desired (e.g. "if all tiles are at the same coordinates, complain loudly")

validate_sptx/test_experiment.py
joshmoore: Which I could see doing here as well. In the schema, I'd like to have a marker, though, that a string field points to a file and of which schema.

joshmoore:
Ok. Comments implemented and travis is green. Landing to see if the currently green master stays that way.

Issue #636, opened 2018-09-26
Mean high pass filter needs a test


kevinyamauchi:
As @ttung pointed out in #620, the mean high pass filter needs a test. Unless there are any objections, I will use the same style of test as the Gaussian high pass (compare the image sum before and after filtering) for now. However, we should probably re-evaluate this testing strategy at some point. For deterministic filters such as these, I think we should have a test image with an expected output and assert that the filter produces this output each time.

kevinyamauchi:
@ttung, @ambrosejcarr, would this be okay for now? I was looking at the GaussianHighPass tests (/starfish/test/pipeline/test_filter.py) and it looks like test_gaussian_high_pass() and test_gaussian_high_pass_apply() are the same. Am I missing something?

ambrosejcarr:
> For deterministic filters such as these, I think we should have a test image with an expected output and assert that the filter produces this output each time.

Agree. I like the checkerboard image used by skimage for some of these things. It should produce an inspectable output.

I think the only thing you need test here is the scientific output. In my open testing profile, I think I hit the rest of the cases that have to do with the API.

kevinyamauchi:
Okay, sounds good. It looks like PR #504 already includes the mean high pass filter, so I think I just need to add the tests in (/starfish/test/pipeline/test_filter.py. Is this correct?

Also, what is the difference between test_gaussian_high_pass() and test_gaussian_high_pass_apply()? It looks like they do the same thing on master.

ttung:
I thought I already did this, but https://github.com/spacetx/starfish/pull/638.

kevinyamauchi:
> I thought I already did this, but #638.

Approved, thanks!

ambrosejcarr:
> Okay, sounds good. It looks like PR #504 already includes the mean high pass filter, so I think I just need to add the tests in (/starfish/test/pipeline/test_filter.py. Is this correct?

Correct.

kevinyamauchi:
Resolved by #655

PR #655, opened 2018-09-27
Add tests for mean high pass filter

kevinyamauchi:
To address issue #636, I added tests for the mean high pass filter a la GaussianHighPass().

### DIFF
--- a/starfish/test/pipeline/test_filter.py
+++ b/starfish/test/pipeline/test_filter.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 
 from starfish.image._filter import gaussian_high_pass
+from starfish.image._filter import mean_high_pass
 from starfish.imagestack.imagestack import ImageStack
 from starfish.types import Number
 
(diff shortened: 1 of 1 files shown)


starfish/test/pipeline/test_filter.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 
 from starfish.image._filter import gaussian_high_pass
```
ttung: Just make these imports on one line, i.e., `from starfish.image._filter import gaussian_high_pass, mean_high_pass`

ttung:
> To address issue #636, I added tests for the mean high pass filter a la GaussianHighPass().

If your github summary includes, "Fixes #xxx", then that issue will automagically get closed when you merge!

starfish/test/pipeline/test_filter.py
```diff
@@ -31,3 +32,22 @@ def test_gaussian_high_pass_3d(sigma: Union[Number, Tuple[Number]]) -> None:
     ghp = gaussian_high_pass.GaussianHighPass(sigma=sigma, is_volume=True)
     result = ghp.run(image_stack)
     assert np.sum(result.numpy_array) < sum_before
+
+@pytest.mark.parametrize('size', (1, (1, 1)))
```
ambrosejcarr: I think we could probably combine these since they have the same test in the end. @ttung, we might be able to apply this to other filters. 

```python3
@pytest.mark.parametrize('size, is_volume', [
    (1, False), 
    ((1, 1), True),
    ((1, 0, 1), False)
])
def test_mean_high_pass(size: Union[Number, Tuple[Number]], is_volume: bool) -> None:
...
```

kevinyamauchi:
> If your github summary includes, "Fixes #xxx", then that issue will automagically get closed when you merge!

Cool! Although, I have to say that closing an issue after merging a PR feels like a victory lap.

starfish/test/pipeline/test_filter.py
kevinyamauchi: Sounds good. I'll open another PR since it will affect other filters.

PR #662, opened 2018-10-01
Make intensity table a package

ambrosejcarr:
Makes intensity table into a package.


starfish/intensity_table/__init__.py
```diff
@@ -0,0 +1 @@
+from .intensity_table import IntensityTable
```
ttung: Don't do this, as it negates the advantage of making it a separate package.  It makes it impossible to modularize IntensityTable into components that can be included without bringing in IntensityTable.

e.g., if you have an `intensity_table/constants.py`, you can not import that without importing `IntensityTable`, which increases the risk of circular imports.

Issue #68, opened 2018-02-12
filters -- remove open cv import


dganguli:
This is no longer needed since we entirely use skimage. Opencv should also be removed from setup.py and requirements.txt.

tractatus:
Could this be reconsidered? If OpenCV is supported both R and golang versions of this package can be made without having to reinvent the wheel for all the heavy lifting.

ttung:
Could that functionality be provided by language bridges (FFI) rather than to maintain a separate identical package?

tractatus:
Yes C/C++ as the common language for all the heavy lifting with image processing. 
FFI calls to C/C++ from both python, R and golang would be easier to maintain. In addition performance with opencv is better than scikit-image. No? Or are you suggetsing R and golang would make FFI to python?

tractatus:
Essentially doing similar what mc-stan did for hamilton monte carlo:
https://github.com/stan-dev
http://mc-stan.org/

But here instead of Bayesian modeling with MCMC/HMC the analogy here being a common interface for basic image processing.

mcstan's approach has been extremely successful with interfaces actually used and maintained to all the most popular data analysis languages:
R, Python, shell, MATLAB, Julia, Stata

And works great across platforms:
Linux, Mac, Windows

ttung:
I would prefer postponing optimizing for performance until we have a robust pipeline and infrastructure.  Fast iteration to get us there is key, and python is good for that.  We also don't want to tie ourselves down to a particular image processing backend (opencv or scikit-image).  The idea is that the interfaces are well-defined, but if people want to build an alternative backend (perhaps opencl for GPU acceleration), there is nothing preventing them for doing so.

It depends on what you need starfish to do.  Do you just want to process images?  Or do you want to develop algorithms?  If you're just processing images, R/golang can just shell out to starfish to do the work.  If you're developing algorithms, that is trickier and I'll have to give it some thought.

tractatus:
**Do you just want to process images? Or do you want to develop algorithms?**
Both preferably, no?

More specifically I have a dataset with 204 FOV, 30 z-stacks per FOV and about 2000-4000 amplicons per cell.
See:
http://www.histo.io/?v=cycle001

This sequencing run is for 30 cycles.

If I'm developing routines for handling these kinds of data I might as well develop in the starfish framework no? 

To get really high reads it is not enough to do registration and segmentation on maximum intensity projection but it has to be done on the stack. Nor is it enough to do registration on single FOV.

[...]

PR #667, opened 2018-10-01
Make tile_data _not_ a property.

ttung:
This reverses what was done with #658 and partly done in #600 because this allows us to pass tile_data as a future.  This is necessary to prevent memory consumption from blowing up.

Note that this does not actually change how we load the data.  A newer version of slicedimage (one that includes #68) is required to actually take advantage of the future.

Test plan: travis.

### DIFF
--- a/examples/get_iss_breast_data.py
+++ b/examples/get_iss_breast_data.py
@@ -37,7 +37,6 @@ class IssCroppedBreastTile(FetchedTile):
         crp = img[40:1084, 20:1410]
         return crp
 
-    @property
     def tile_data(self) -> np.ndarray:
         return self.crop(imread(self.file_path))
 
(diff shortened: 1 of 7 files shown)


ambrosejcarr:
I didn't think properties affected the ability to pass futures, in fact they're explicitly used to memoize in python. You attach the future, and the first time the property gets called, you store the result in a private variable, often with the same name, e.g.: 

```python3
@property
def tile_data(self):
    if self._tile_data is None: 
        self._tile_data = self._tile_data_future()
    return self._tile_data
```

The challenge comes in recycling the future, if you want to be able to evict the data and re-use the original object (which would require another call to the future to get the data back), but this pattern still makes sense, in my mind.

ttung:
> I didn't think properties affected the ability to pass futures, in fact they're explicitly used to memoize in python. 

I'm trying to get the bound method as a future, not that the property is retrieving a future.

Issue #666, opened 2018-10-01
Excessive memory consumption from experiment builder


ttung:
@zperova reported this issue on slack.  It eventually leads to the experiment builder getting killed.

PR #668, opened 2018-10-01
Fix memory consumption issues with merfish formatter

ttung:
Use `set_numpy_array_future` to avoid loading the tile data until we start writing out to disk.

Depends on #667 

Fixes https://github.com/spacetx/starfish/issues/666


ttung:
pip pkg for slicedimage released; ready for review.

PR #679, opened 2018-10-03
RFC: user-facing debugging instructions

ambrosejcarr:
This doc will eventually detail how users should expect to deal with our rapid iteration. It brought up a few questions that I think we should mull over: 

1. **Assuming users are using starfish with data we host, how do they find the right datasets that match the starfish version?** 
    1. Right now, we go by date. This is fine if they clobber their notebooks with ours, but it's not great UX for a user that is creating their own notebooks. Could we replace our date-stamps with version names that match our starfish/slicedimage versions? Presumably a breaking change that requires new data warrants a version change, and so this should be feasible. 

2. **Assuming users are using starfish with data they've generated, how do they update their SpaceTx-format metadata to work with the current starfish version?** 
    1. Here I'm a bit more stuck -- if they used TileFetcher, would it be possible for us to keep that up to date such that they could simply re-run it? I feel like the current subclassing approach is not amenable to this.

### DIFF
--- a/docs/source/debugging.rst
+++ b/docs/source/debugging.rst
@@ -1,0 +1,50 @@
+Debugging Errors
+================
+
+First, thank you for using Starfish and SpaceTx-Format! Feedback you provide on features and the
+user experience is critical to making Starfish a successful tool. Because we iterate quickly on this
+feedback to add new features, things change often, which can result in your code getting out of sync
+with your data. When that happens, you may observe errors.
+
+Most of the time, you can fix this problem by pulling the most recent version of the code,
+reinstalling starfish, and restarting your environment. If you're using starfish with datasets from
+spaceTx located on our cloudfront distribution, we're committed to keeping that data up to date.
+You can find it in a versioned folder, and the version will correspond to the version of starfish
+that it matches.
+
+If you're using your own data with starfish, you may need to re-run your data ingestion workflow
+based on TileFetcher and FetchedTile to generate up-to-date versions of spaceTx-format.
+
+Upgrading to a new version
+--------------------------
+
+If you've installed from pypi, upgrading is as simple as reinstalling starfish.
+
+.. code-block:: bash
+
+    pip3 install --upgrade starfish
+
+If you've installed our development version to take advantage of new features in real time, you'll
+need to fetch changes and reinstall. Assuming you've cloned the respository into ``./starfish``,
+you can install the newest version as follows:
+
+.. code-block:: bash
+
+    cd ./starfish
+    git checkout master
+    git pull
+    pip3 install .
+
+Reporting bugs
+--------------
+
+Bugs can be contributed as issues in the starfish repository. Please check to make sure there
+is no existing issue that describes the problem you have identified before adding your bug.
+
+When reporting issues please include as much detail as possible about your operating system,
+starfish version, slicedimage version, and python version. Whenever possible, please also include a
+brief, self-contained code example that demonstrates the problem, including a full traceback.
+
+We can also be contacted on the SpaceTx slack in the ``#starfish-users`` channel. If you'd like an
+invitation, please feel free to email us. We can usually respond to bug reports same-day, and
+are very appreciative of the time you take to submit them to us.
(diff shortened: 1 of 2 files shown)


docs/source/debugging.rst
```diff
@@ -0,0 +1,50 @@
...
+
+First, thank you for using Starfish and SpaceTx-Format! Feedback you provide on features and the
+user experience is critical to making Starfish a successful tool. Because we iterate quickly on this
+feedback to add new features, things change often, which can result in your code getting out of sync
+with your data. When that happens, you may observe errors.
+
+Most of the time, you can fix this problem by pulling the most recent version of the code,
+reinstalling starfish, and restarting your environment. If you're using starfish with datasets from
+spaceTx located on our cloudfront distribution, we're committed to keeping that data up to date.
+You can find it in a versioned folder, and the version will correspond to the version of starfish
```
ttung: This is not particularly clear.

docs/source/debugging.rst
```diff
@@ -0,0 +1,50 @@
...
+that it matches.
+
+If you're using your own data with starfish, you may need to re-run your data ingestion workflow
+based on TileFetcher and FetchedTile to generate up-to-date versions of spaceTx-format.
```
ttung: link to docs?

docs/source/debugging.rst
```diff
@@ -0,0 +1,50 @@
...
+    pip3 install .
+
+Reporting bugs
+--------------
+
+Bugs can be contributed as issues in the starfish repository. Please check to make sure there
+is no existing issue that describes the problem you have identified before adding your bug.
+
+When reporting issues please include as much detail as possible about your operating system,
+starfish version, slicedimage version, and python version. Whenever possible, please also include a
```
ttung: a lot of this can be captured by running `pip freeze`

docs/source/debugging.rst
```diff
@@ -0,0 +1,50 @@
...
+based on TileFetcher and FetchedTile to generate up-to-date versions of spaceTx-format.
+
+Upgrading to a new version
+--------------------------
+
+If you've installed from pypi, upgrading is as simple as reinstalling starfish.
+
+.. code-block:: bash
+
+    pip3 install --upgrade starfish
```
shanaxel42: If they're using the virtualenv they setup with the starfish setup instrucitons I think this would just be  pip install --upgrade starfish, we should probably be consistent on our setup and debugging instructions to reflect one or the other.

docs/source/debugging.rst
ambrosejcarr: Can we commit to this? Are there problems I'm not seeing with replacing dates with versions? I'm happy to add an example such as: 

```diff
For example, if you are upgrading from starfish 0.1.0 0.1.1, you will need to change 
references to our datasets as follows: 

- http://czi.starfish.data.public.s3.amazonaws.com/browse/formatted/0.1.0/iss_breast/experiment.json
+ http://czi.starfish.data.public.s3.amazonaws.com/browse/formatted/0.1.1/iss_breast/experiment.json
```

ttung: This will likely result in more copies of our data.  Not every version of starfish changes the file format.

ambrosejcarr: Ah, that's a good point. What do you think is the right approach for the user experience?

docs/source/debugging.rst
ambrosejcarr: Reasonable -- I think both work inside the virtualenv, but I've edited this to be consistent.

docs/source/debugging.rst
ambrosejcarr: Added forward-facing references for when this is in the API.

docs/source/debugging.rst
ttung: One option: it's more overhead, but we can create a top-level experiment.json that points to a bunch of shared data files.

ambrosejcarr: It seems like we may need to move in that direction anyway with spaceTx data coming. Our model doesn't seem sustainable when we move into the 100s of Tbs. With that stipulation in mind, does it make sense to go that route? 

I think the right path here is for me to remove it as you say, and for you to add an issue to the backlog to mark the need to revisit this?

Issue #677, opened 2018-10-02
Write an adapter for Napari stack visualization.


ambrosejcarr:
Napari is a python package for visualization of image data. We should write an adapter for it. 
See demo: https://slack-files.com/T8MB1T0MA-FD5SHUAR0-84d2f2dc8b. 

Might be a solution for #603

kevinyamauchi:
I was planning to put this into ImageStack.show_stack(). How does that sound?

ambrosejcarr:
> I was planning to put this into ImageStack.show_stack(). How does that sound? (cc @dganguli )

Do you mean it would replace the existing functionality? If so, does it persist in a notebook? Although the existing approach is a bit clunky, I think there's some benefit in that it is just sitting there when you come back to it, versus opening in another window that the user has to close.

kevinyamauchi:
I don't think it needs to completely replace existing functionality. Perhaps show_stack() can take a Napari argument? Alternatively, we can add something to the viz suite. 

Also, I'm going to ask Kira about plotting. We may be able to also do show_spots() in Napari.

ambrosejcarr:
> I don't think it needs to completely replace existing functionality. Perhaps show_stack() can take a Napari argument? Alternatively, we can add something to the viz suite.

Cool, my sense is that the safest thing to do would be to add a function `show_napari()` for now, and then we can combine/discard the old stuff once we've played around a bit.

PR #680, opened 2018-10-03
RFC: add method to ImageStack to use Napari for show stack

kevinyamauchi:
As discussed in #677, the nascent [Napari](https://github.com/Napari) viewer can view large stacks and has performant scrolling between slices. Also, it automatically generates the appropriate number of sliders for multidimensional stacks. It seems useful for debugging and optimizing starfish pipeines, so I've added an adapter I've added some notes below. Please let me know your thoughts.

* Since the viewer isn't fully featured, I've created a separate method (`show_stack_napari()`), so we can continue to use all current viewers.
* `show_stack_napari()` uses the get_slices method to create the data arrays to pass to Napari, so we can slice the ImageStack. 
* Since scrolling through all dimensions in a full 5D ImageStack is not a possibility, we may want to add an easy option to just pass the whole stack (e.g., `indices='all'`) to Napari rather than giving all indices. 
* I used the instructions in REQUIREMENTS.txt to update it - please let me know if I did that correctly.


kevinyamauchi:
P.S. I didn't encode the test notebook as a py file because I think we'd nuke it before merging.

ttung:
How big are these packages?  I imagine that PyQt is probably ginormous.

One possible solution is to move the import into `show_stack_napari()`, wrap it like:

```
try:
    import napari_gui
except ImportError:
    warnings.warn("Cannot find the napari library. Install it by running \"pip install napari\"")
    return

<rest of show_stack_napari()>
```

kevinyamauchi:
Good point, @ttung. Pardon my ignorance, but in the case that napari_gui is only imported with that method, does napari_gui go in the `REQUIREMENTS.txt`?

ttung:
I don't think we have a good solution for this problem (of bloating requirements) yet.  I would prefer not to block you while we sort things out.  If @ambrosejcarr and @dganguli approve this, I have no problem having this land and we can fix the requirements problem afterwards.

starfish/imagestack/imagestack.py
```diff
@@ -386,6 +387,28 @@ def set_slice(
 
         self._data.values[slice_list] = data
 
+    def show_stack_napari(self, indices: Mapping[Indices, Union[int, slice]]):
```
ambrosejcarr: This is awesome. Is it possible to have napari label the sliders? I find that part of it a bit unintuitive. 

If it's _not_ possible, we should document the order that they appear on the GUI from top to bottom. Based on the MERFISH example, seems like it's the same as the `reordered_array`?

starfish/imagestack/imagestack.py
```diff
@@ -386,6 +387,28 @@ def set_slice(
...
+            Indices to select a volume to visualize. Passed to `Image.get_slice()`.
+            See `Image.get_slice()` for examples.
+
+        Notes
+        -----
+        To use in a Jupyter notebook, use the %gui qt5 magic.
+
+
+        """
+        # Switch axes such that it is indexed [x, y, rnd, c, z]
```
ambrosejcarr: I'd prefer `[x, y, r, c, z]` or `[x, y, round, channel, z]`

PR #684, opened 2018-10-03
RFC: Add Expression Matrix + Links to Seurat & Scanpy

ambrosejcarr:
Early PR for feedback on linking `IntensityTable` to an `ExpressionMatrix` (2d genes x cells) object that can be exported to `loom` and `AnnData` for use in `Scanpy` and `Seurat`

Before I write tests, do we like how this looks? Anything anyone would change or add? 

- [x] wrote constructor from IntensityTable
- [x] ExpressionMatrix saves and loads data in each format, and the files load into scanpy and Seurat
- [ ] Tests for serialization functions

### DIFF
--- a/docs/source/API/core_objects/expression_matrix.rst
+++ b/docs/source/API/core_objects/expression_matrix.rst
@@ -1,0 +1,9 @@
+ExpressionMatrix
+================
+
+Expression Matrix is a wrapper for xarray.DataSet that provides serialization adapters to netCDF,
+loom, and AnnData to enable it to be used with single-cell analysis software packages such as
+scanpy (python) and seurat (R). 
+
+.. automodule:: starfish.expression_matrix.expression_matrix.ExpressionMatrix
+   :members:
--- a/docs/source/API/core_objects/expression_matrix.rst
+++ b/docs/source/API/core_objects/expression_matrix.rst
@@ -1,9 +1,9 @@
 ExpressionMatrix
 ================
 
-Expression Matrix is a wrapper for xarray.DataSet that provides serialization adapters to netCDF,
+Expression Matrix is a wrapper for xarray.DataArray that provides serialization adapters to netCDF,
 loom, and AnnData to enable it to be used with single-cell analysis software packages such as
-scanpy (python) and seurat (R). 
+scanpy (python) and seurat (R).
 
 .. automodule:: starfish.expression_matrix.expression_matrix.ExpressionMatrix
    :members:
--- a/starfish/expression_matrix/expression_matrix.py
+++ b/starfish/expression_matrix/expression_matrix.py
@@ -1,0 +1,146 @@
+import anndata
+import loompy
+import xarray as xr
+
+
+class ExpressionMatrix(xr.DataArray):
+    """Container for expression data extracted from an IntensityTable
+
+    An ExpressionMatrix is a 2-dimensional ``cells`` x ``genes`` tensor whose values are the
+    number of spots observed for each gene observed by the experiment. In addition to the basic
+    xarray methods, IntensityTable implements:
+
+    Methods
+    -------
+    save(filename)
+        save the ExpressionMatrix to netCDF
+
+    save_loom(filename)
+        save the ExpressionMatrix to loom for use in R or python
+
+    save_anndata(filename)
+        save the ExpressionMatrix to AnnData for use in ``Scanpy``
+
+    load(filename)
+        load an ExpressionMatrix from netCDF
+
+    load_loom(filename)
+        load an ExpressionMatrix from loom
+
+    load_anndata(filename)
+        load an ExpressionMatrix from AnnData
+
+
+    Examples
+    --------
+    # TODO ambrosejcarr write examples
+
+    """
+
+    def to_anndata(self) -> anndata.AnnData:
+        """convert ExpressionMatrix to a scanpy-compatible AnnData object"""
+        row_attrs = {k: self['cells'][k].values for k in self['cells'].coords}
+        col_attrs = {k: self['genes'][k].values for k in self['genes'].coords}
+        return anndata.AnnData(self.data, row_attrs, col_attrs)
+
+    def save(self, filename: str) -> None:
+        """Save an ExpressionMatrix as a Netcdf File
+
+        Parameters
+        ----------
+        filename : str
+            Name of Netcdf file
+
+        """
+        self.to_netcdf(filename)
+
+    def save_loom(self, filename: str) -> None:
+        """Save an ExpressionMatrix as a loom file
+
+        Parameters
+        ----------
+        filename : str
+            Name of loom file
+
+        """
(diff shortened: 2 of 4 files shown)


dganguli:
This is awesome! Love it. Would be very curious to hear what @ttung and @shanaxel42 think tho.

starfish/expression_matrix/expression_matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        -------
+        ExpressionMatrix
+
+        """
+        loaded = xr.open_dataarray(filename)
+        expression_matrix = cls(
+            loaded.data,
+            loaded.coords,
+            loaded.dims
+        )
```
shanaxel42: Could this load method detect what kind of file (Netcdf, AnnData, loom) is being passed in and respond accordingly? So you'd only need one load method?

docs/source/API/core_objects/expression_matrix.rst
```diff
@@ -0,0 +1,9 @@
+ExpressionMatrix
+================
+
+Expression Matrix is a wrapper for xarray.DataSet that provides serialization adapters to netCDF,
```
shanaxel42: xr.DataArray not DataSet right?

starfish/intensity_table/intensity_table.py
```diff
@@ -384,6 +386,54 @@ def to_features_dataframe(self) -> pd.DataFrame:
...
+            metadata['area'] = ("cells", np.full(counts.shape[0], fill_value=np.nan))
+
+        # add genes to the metadata
+        metadata.update({"genes": counts.columns.values})
+        metadata.update({"cell_id": ("cells", counts.index.values)})
+
+        mat = ExpressionMatrix(
+            data=counts.values,
+            dims=('cells', 'genes'),
+            coords=metadata,
```
shanaxel42: So coords is a dictionary that has area, x, y, z, cell_id and gene values per cell in it?

starfish/intensity_table/intensity_table.py
```diff
@@ -384,6 +386,54 @@ def to_features_dataframe(self) -> pd.DataFrame:
...
+        Parameters
+        ----------
+        regions: Optional[regional.Many]
+            cell segmentation results that were used to assign points to cells. If not provided, the
+            centers of the cells will be estimated by taking the midpoint between the extreme-valued
+            spots on each axis.
+
+        Returns
+        -------
+        pd.DataFrame :
```
shanaxel42: returns a xr.DataArray?

starfish/expression_matrix/expression_matrix.py
ambrosejcarr: I'm not sure I can trust AnnData and Loom have used magic numbers in their formats. It looks like we'd need to open the file with pytables to inspect the index to try to guess which is which. That could work. More likely we'd need to base it off of file extensions, which I don't really like to do. 

I foresee such a function ending with "Warning: could not autodetect file format, defaulting to `netCDF`. To override, please use a specific `load_<format>` method". 

At least in the short term, I don't know if this is worth the time. Honestly I expect most people to take their matrix, leave, and never return to starfish. :-)

starfish/intensity_table/intensity_table.py
ambrosejcarr: The straightforward answer is that it is used to define a [coordinates](http://xarray.pydata.org/en/stable/data-structures.html#coordinates) object with said information, and a dicts is a way to provide information to that constructor

REQUIREMENTS.txt
```diff
@@ -1,4 +1,5 @@
 # You should not edit this file directly.  Instead, you should edit one of the following files (REQUIREMENTS.txt.in) and run make REQUIREMENTS.txt
+anndata==0.6.10
```
ttung: I still think we should _not_ include these dependencies as exporters are not in the realm of core functionality.  One could easily imagine this getting out of hand.

starfish/expression_matrix/expression_matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        load an ExpressionMatrix from AnnData
+
+
+    Examples
+    --------
+    # TODO ambrosejcarr write examples
+
+    """
+
+    def to_anndata(self) -> anndata.AnnData:
```
ttung: make private?

REQUIREMENTS.txt
ambrosejcarr: Cool, what's the pattern that you'd prefer to see here? An attempt to import where an error is thrown if the dependency is not installed that instructs the user to pip install the thing?

ambrosejcarr:
I'm good to delete the load methods. I don't see why a user would want a count matrix as we've defined it if they already have something they can load into scanpy/seurat.

REQUIREMENTS.txt
ttung: I think starfish-full should install everything used by the codebase and the notebooks.

ttung: > Cool, what's the pattern that you'd prefer to see here? An attempt to import where an error is thrown if the dependency is not installed that instructs the user to pip install the thing?

Yes, that's correct.

ambrosejcarr:
Updated to remove extra requirements, added `try_import` function.

starfish/intensity_table/intensity_table.py
```diff
@@ -384,6 +386,54 @@ def to_features_dataframe(self) -> pd.DataFrame:
...
+                "area": ("cells", [regions[id_].area for id_ in counts.index]),
+                "x": ("cells", [regions[id_].center[0] for id_ in counts.index]),
+                "y": ("cells", [regions[id_].center[1] for id_ in counts.index]),
+                "z": ("cells", np.zeros(counts.shape[0]))
+            }
+        else:
+            grouped = self.to_features_dataframe().groupby(['cell_id'])[['x', 'y', 'z']]
+            min_ = grouped.min()
+            max_ = grouped.max()
+            coordinate_df = min_ + (max_ - min_) / 2
```
ttung: I don't think you need to do this trick as python doesn't really have a number range (it automatically promotes types).

starfish/util/try_import.py
```diff
@@ -0,0 +1,11 @@
+import importlib
+from typing import Any
+
+def try_import(module: str) -> Any:
```
ttung: I have some concerns about this approach given that it makes IDEs and mypy probably very unhappy, so check out #734.

starfish/intensity_table/intensity_table.py
ambrosejcarr: I don't understand. What I'm attempting to capture here is the midpoint between extreme values of spots associated with cells, to build a "center" for each cell.

ttung: so isn't that just (min_ + max_) / 2?

The min_ + (max_ - min_) / 2 trick is one that people do to prevent overflow.

PR #687, opened 2018-10-03
[osmFISH 2/5] Implement Laplacian Filter

dganguli:
This filter enhances spots, and is used by the osmFISH pipeline. Under the hood, it's simply a call to ```scipy.ndimage.gaussian_laplace```

Note that there is no test for this filter. From offline conversations, it's not clear what our testing strategy for filters are. Especially for filters that are pulled in from existing libraries, like this one. Really, what we want to test here is that the Starfish specific wrapping code works as expected.

For now, I'll rely on [osmFISH 5/5] notebook to function as the test of this filter.

### DIFF
--- a/starfish/image/_filter/__init__.py
+++ b/starfish/image/_filter/__init__.py
@@ -9,6 +9,7 @@ from . import bandpass
 from . import clip
 from . import gaussian_high_pass
 from . import gaussian_low_pass
+from . import laplace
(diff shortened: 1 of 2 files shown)


dganguli:
FYI @simone-codeluppi

ambrosejcarr:
Ready for review.

PR #688, opened 2018-10-03
[osmFISH 3/5] Re-name LocalMaxPeakFinder -> TrackpyLocalMaxPeakFinder

dganguli:
osmFISH uses a different peak finding algorithm implementation than the one we implemented for allensmFISH. To avoid confusion, I've re-named the Trackpy one to LocalMaxPeakFinder. 

[osmFISH 4/5] implements another local max peak finding algorithm that comes from skimage. I call that one LocalMaxPeakFinder instead such that we can differentiate.


ttung:
Please update the allen notebook with this refactor.

PR #689, opened 2018-10-03
[osmFISH 4/5] Implement LocalMaxPeakFinder

dganguli:
This SpotDetector uses skimage.local_max_peak_finder, and implements the automatic thresholding algorithm employed by pysmFISH.

Several outstanding issues remain:
- skimage.local_max_peak_finder only works in 3D
- the automatic thresholding code needs some documentation
- it's unclear how to pull out the radius from these algorithms. in fact, the radius is not needed for sequential smFISH anyways. radius is mostly just necessary for non-pixel spot detection based methods for multiplexed techniques
- i'm not sure the CLI hookup was done correctly. it's not tested either.
- there's no test for this. but i think that's easy to implement? for now i rely on the notebook from [osmFISH 5/5]

FYI @simone-codeluppi

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -14,6 +14,7 @@ jsonschema==2.6.0
 kiwisolver==1.0.1
 matplotlib==2.2.3
 more-itertools==4.3.0
+mpmath==1.0.0
 networkx==2.1
 numpy==1.15.1
 packaging==17.1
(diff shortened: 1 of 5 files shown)


ambrosejcarr:
Ready for review.

REQUIREMENTS.txt
```diff
@@ -38,6 +39,7 @@ semantic-version==2.6.0
 showit==1.1.4
 six==1.11.0
 slicedimage==1.0.0
+sympy==1.3
```
joshmoore: An update to environment.yml and or a josh-ping when this lands is appreciated.

PR #691, opened 2018-10-03
Updated test_api_contract.py to use .xarray instead of .numpy_array

ttung:
Test plan: `pytest starfish/test/image/filter/test_api_contract.py`


ttung:
Fixed remaining tests.

PR #696, opened 2018-10-06
fixing float bug

shanaxel42:
I discovered this bug while trying to test something else. But the preprocessing library used by the bandpass filter returns results in float64.

### DIFF
--- a/starfish/image/_filter/bandpass.py
+++ b/starfish/image/_filter/bandpass.py
@@ -81,7 +81,7 @@ class Bandpass(FilterAlgorithmBase):
             image, lshort=lshort, llong=llong, threshold=threshold,
             truncate=truncate
         )
-        return bandpassed
+        return bandpassed.astype(np.float32)
 
     def run(
             self, stack: ImageStack, in_place: bool=False, verbose: bool=False,


starfish/image/_filter/bandpass.py
```diff
@@ -81,7 +81,7 @@ def _bandpass(
             image, lshort=lshort, llong=llong, threshold=threshold,
             truncate=truncate
         )
-        return bandpassed
+        return bandpassed.astype(np.float32)
```
ttung: should this call `preserve_float_range`?

shanaxel42: I think preserve float range only clips or rescales the values, it doesn't cast to float32

ttung:
```
def preserve_float_range(
        array: Union[xr.DataArray, np.ndarray],
        rescale: bool=False) -> Union[xr.DataArray, np.ndarray]:
    """
    Clips values below zero to zero. If values above one are detected, clips them
    to 1 unless `rescale` is True, in which case the input is scaled by
    the max value and the dynamic range is preserved.

    Parameters
    ----------
    array : Union[xr.DataArray, np.ndarray]
        Array whose values should be in the interval [0, 1] but may not be.
    rescale: bool
        If true, scale values by the max.

    Returns
    -------
    array : Union[xr.DataArray, np.ndarray]
        Array whose values are in the interval [0, 1].

    """
    array = array.copy()
    is_xr = isinstance(array, xr.DataArray)
    if np.any(array < 0):
        if is_xr:
            array.values[array.values < 0] = 0
        else:
            array[array < 0] = 0
    if np.any(array > 1):
        if rescale:
            array /= array.max()
        else:
            if is_xr:
                array.values[array.values > 1] = 1
            else:
                array[array > 1] = 1
    return array.astype(np.float32)
```

Final line should cast to float32.

ambrosejcarr: More generally, I like tony's idea here -- if `preserve_float_range` _didn't_ work, I'd wanna fix it. :-)

ttung: Looks like this was landed as part of #697.  Can you comment on this thread? :)

ambrosejcarr: I think this should call `preserve_float_range`.

shanaxel42: Sorry sorry had spaced on this. BAM https://github.com/spacetx/starfish/pull/735

ttung: BAM!

Issue #603, opened 2018-09-19
investigate adding to_tiff() method to ImageStack 


dganguli:
A common pattern I've encountered in using Starfish is:

0. Load data in spaceTx format,
1. Do some filtering,
2. Visualize results.
3. Do thinks make sense? If yes, pass. Else, GOTO 1.
4. Find spots, decode, etc.

For step 2. we expose a show_stack method, which visualizes the data using a Jupyter widget. However, I find myself wanting to pan, zoom, adjust dynamic range, etc. And this is hard to do in a performant way with Jupyter widgets.

Instead, I write out tiff files, and open them in FIJI. This is often painful. I have to convert the float32 images into a tiff specific format. To do this, I often just pick an single tile form the full 5 dimensional stack, convert it to uint16, and pick some ad-hoc tiff-writer to save out data.

I suggest we expose a ImageStack.to_tiff() method. This will take our ImageStack, do all the necessary conversions, and choose a specific tiff format (I vote OME-tiff!) to write to disk. We can potentially some skimage compatibility to do so: http://scikit-image.org/docs/dev/api/skimage.external.tifffile.html

@joshmoore what's the right format for our data? which converter should we use from the Python eco-system?

FYI @ambrosejcarr @ttung @shanaxel42

dganguli:
Here's what I currently do:

```
from skimage import img_as_uint
from skimage.io import imsave
# filt2 is of type ImageStack. Here I'm choosing round 1, channel 1, z-plane 1 to save as a tiff to open #up with FIJI
im_tiff = img_as_uint(filt2.numpy_array[0,0,0,:,:])
imsave('/Users/dganguli/Desktop/fuck.tiff', im_tiff)
```

dganguli:
Would be nice if I could save all dimensions, and use hyperstack functionality of FIJI to look at all the data somehow.

joshmoore:
@dganguli : the three options for writing out OME-TIFFs that occur to me are:

 * **use ome-files-py**: pretty friendly for the user, but the build/deployment burden is pretty high without something like docker, conda or wheels.
 * **call out to Java**: there are libraries like https://github.com/soft-matter/pims that make this more manageable but, again, a deployment burden.
 * **write a simplified ome-tiff writer in python**: I don't imagine this would be too hard and somewhat matches the strategy that I've taken with spacetx-fov-writer where the implementation in one language is designed to support a specific subset of the other language.

[...]

PR #707, opened 2018-10-12
Imagestack Indexing and Slicing 

shanaxel42:


starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+        indices: The (Round, Ch, Z) indices that identify which tile's coordinates we're using
+
+        key: a value or range to index on
+
+        """
+        coord_min, coord_max = self.coordinates(indices, coord)
+        dimension_size = self.xarray.sizes[coord.value[0]]
+        physical_pixel_size = (coord_max - coord_min) / dimension_size
+        new_min, new_max = coord_min, coord_max
+        min_pixel_index = key if type(key) is int else key.start
```
ambrosejcarr: This is disambiguating between an int and a `slice`?

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+
+                    new_coords[indices].loc[
+                        dict(physical_coordinate=[
+                            PhysicalCoordinateTypes.X_MIN,
+                            PhysicalCoordinateTypes.X_MAX,
+                            PhysicalCoordinateTypes.Y_MIN,
+                            PhysicalCoordinateTypes.Y_MAX]
+                        )] = [xmin, xmax, ymin, ymax]
+
+    def _rescale_physical_coordinate(self, coord: Coordinates, indices: Mapping[Indices, int],
```
ambrosejcarr: I don't think we're ever scaling in the sense that "scale" connotes a multiplication by a constant value -- we're "sub-setting" or "slicing", or something like that, right?

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+                                     key) -> Tuple:
+        """Calculates rescaled coordinates
```
ambrosejcarr: I think we should be more verbose here. Talk about the implications of this.

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+        physical_pixel_size = (coord_max - coord_min) / dimension_size
+        new_min, new_max = coord_min, coord_max
+        min_pixel_index = key if type(key) is int else key.start
+        max_pixel_index = key if type(key) is int else key.stop
+        if min_pixel_index:
+            # check for negative index
+            min_pixel_index = min_pixel_index + dimension_size if min_pixel_index < 0 \
+                else min_pixel_index
+            new_min = (physical_pixel_size * min_pixel_index) + coord_min
+        if max_pixel_index:
```
ambrosejcarr: When would max be None/0?

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+            max_pixel_index = max_pixel_index + dimension_size if max_pixel_index < 0 \
+                else max_pixel_index
+            new_max = (physical_pixel_size * (max_pixel_index + 1)) + coord_min
+        return new_min, new_max
+
+    def _needs_coords_resacling(self, indexers: dict) -> bool:
```
ambrosejcarr: Spelling, but also this is probably not "rescaling" per above comment.

starfish/imagestack/imagestack.py
shanaxel42: Yea when the indexing comes in as a slice like [:5] it's turned into a python Slice object which is basically just a tuple but the object has start and stop attributes that are the start and stop of the slicing operation.

starfish/imagestack/imagestack.py
shanaxel42: Yea good point, well we are doing some scaling operations to determine the corresponding physical coordinate of a new pixel one...but there's probably a better name for that. Maybe like calculate_new_pyhsical_coordinate? ...I'm open to suggestions haha

starfish/imagestack/imagestack.py
shanaxel42: So if the slicing operation if from start to somewhere [:3] or end to somewhere [3:] the python Slice object that comes in has a start or stop attribute of None. So we just wanna keep that coordinate value as is.

ambrosejcarr: Cool, please immortalize this in a comment. 😁

starfish/imagestack/imagestack.py
ambrosejcarr: Cool, store as a line comment please. :-)

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
 
+    def __getitem__(self, key) -> "ImageStack":
+        # index xarray (keep dims) and create new stack
+        indexed_data = self._index_keep_dimensions(self.xarray, key)
+        stack = self.from_numpy_array(indexed_data.data)
+        # set coords on new stack
+        stack._coordinates = self._calc_new_coords(indexers=self._item_key_to_dict(key))
+        return stack
+
+    def _item_key_to_dict(self, key):
```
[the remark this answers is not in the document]
ttung: I actually think it would be easier if you take each index value, and expanded it to the actual values, i.e., `slice(None)` -> `slice(0, sz-1)`.  your code for recalculating the coordinates would probably be isolated from the weirdness of slice.

joshmoore:
Ah! This is exactly what I was looking for re: gh-603. If so, I'd _not_ add the parameters to https://github.com/joshmoore/starfish/commit/a4b73d0966c10dc84342ec053b336febb81da24e which get forwarded to `get_slice` and would instead plan on using `stack[...].write()` for @dganguli's purposes.

ambrosejcarr:
> I'm not so enthusiastic about __getitem__ taking an unlabeled key. It assumes a certain geometry for the composed xarray. If we discover that our algorithms benefit from a particular geometry for the xarray, this will be hard to change. Much preferred: stack[dict(Indices.Round: 5, ...)]

I agree on the labeled key. Sorry I didn't think about that when describing the use case. I'd like us to take a look at the `xarray` indexing options and see if there are any they have which might fit our use cases. Fundamentally, I think we're trying to accomplish the same thing. Also, I think their UX is really nice. They've obviously spent some time tuning it. @ttung @shanaxel42 Do you think any of these options would work? 

http://xarray.pydata.org/en/stable/indexing.html

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +290,131 @@ def xarray(self) -> xr.DataArray:
...
+        # store original dims
+        original_dims = data.dims
+        # index
+        data = data[key]
+        # find missing dims
+        missing_dims = set(original_dims) - set(data.dims)
+        # Add back in missing dims
+        data = data.expand_dims(tuple(missing_dims))
+        # Reorder to correct format
+        return data.transpose(*original_dims)
```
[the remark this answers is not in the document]
dganguli: so fresh and so clean clean

dganguli:
This is awesome! In terms of UX, I think it may also be cool to have a way to index the slice with named indices:

```
from starfish import ImageStack

stack = ImageStack.synthetic_stack() # <starfish.ImageStack (r: 4, c: 4, z: 12, y: 50, x: 40)>`

stack.get({Indices.C: 0}) #should be same as stack[:, 0, :, : , :]

stack.get({Indices.R: [0:3]}) #should be same as stack[0:3, :, :, :, :]

stack.get({Indices.C:0, Indices.R: [0:3]}) # should be same as stack[0, 0:3, :, : , :]
```
@ambrosejcarr what do you think?

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +291,37 @@ def xarray(self) -> xr.DataArray:
         """Retrieves the image data as an xarray.DataArray"""
         return self._data
 
+    def _convert_to_indexers_dict(self, pos_kwargs):
```
ttung: It would be helpful to add a docblock to this method.

ttung: Also, I don't understand what the value of this as a method is.  It is only used once, and both the caller and this method are pretty simple methods...

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +291,37 @@ def xarray(self) -> xr.DataArray:
...
+        return_dict = {ind.value: slice(None, None) for ind in Indices}
+        for key, value in pos_kwargs.items():
+            if isinstance(value, tuple):
+                value = slice(value[0], value[1])
+            return_dict[key] = value
+        return return_dict
+
+    def sel(self, indexers: Mapping[Indices, Union[int, tuple]]=None):
```
ttung: Docblock this please.

starfish/imagestack/imagestack.py
```diff
@@ -26,6 +26,7 @@
 from starfish.errors import DataFormatWarning
 from starfish.experiment.builder import build_image, TileFetcher
 from starfish.experiment.builder.defaultproviders import OnesTile, tile_fetcher_factory
+from starfish.imagestack import physical_coordiante_calculator
```
ttung: s/physical_coordiante_calculator/physical_coordinate_calculator/

starfish/imagestack/physical_coordiante_calculator.py
```diff
@@ -0,0 +1,114 @@
...
+    indexers : a dictionary of dim:index where index is the value
+    or range to index the dimension
+
+    """
+    new_coords = image_stack._coordinates.copy()
+    # index by R, CH, V
+    key = {'r': indexers[Indices.ROUND.value],
+           'c': indexers[Indices.CH.value],
+           'z': indexers[Indices.Z.value]}
+    new_coords = image_stack.index_keep_dimensions(new_coords, key)
```
ttung: `ImageStack.sel` itself calls `index_keep_dimensions`, and then it also calls `calc_new_physical_coords`, which calls `index_keep_dimensions` again.  This doesn't seem right.

I propose that `calc_new_physical_coords` takes:
1. named shape of the original 5D tensor
2. named physical coordinates of the original 5D tensor
3. named indexers

From that, you should be able to calculate the coordinates of the indexed array.

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +291,37 @@ def xarray(self) -> xr.DataArray:
...
+        indexers = self._convert_to_indexers_dict(indexers)
+        indexed_data = self.index_keep_dimensions(self.xarray, indexers)
+        stack = self.from_numpy_array(indexed_data.data)
+        # set coords on new stack
+        stack._coordinates = physical_coordiante_calculator.calc_new_physical_coords(
+            self, indexers)
+        return stack
+
+    @staticmethod
+    def index_keep_dimensions(data: xr.DataArray, indexers) -> xr.DataArray:
```
ttung: proposal: this method can stand alone outside of `ImageStack`.  it's a static method anyway.

starfish/test/image/test_imagestack_index.py
```diff
@@ -0,0 +1,134 @@
+from collections import OrderedDict
+
+import numpy as np
+
+from starfish.imagestack.imagestack import ImageStack
+from starfish.types import Coordinates, Indices
+
+
+def test_indexing_by_r_ch_z():
```
ttung: Most of your testing can center around `calc_new_physical_coords` to make the tests simpler and easier to comprehend.

starfish/imagestack/physical_coordiante_calculator.py
shanaxel42: Well I'm calling index keep dimensions on two different xarrays. One is Imagestack.xarray in sel just indexing by whatever is passed in and the second is the coordinates xarray that should also be indexed by the same R, CH, V values before then calculating the actual coordinate values. Which part doesn't seem right? Both xarrays need to be passed into index_keep_dimensions at some point, I guess I could do both in the sel method but my thinking was that calc_new_physical_coords was doing everything to "calculate a new physical coords xarray"

starfish/imagestack/imagestack.py
shanaxel42: It does three things 1.) creates an empty slice(None, None) entry for every dimensions in indexers so that we don't have to add a bunch of "If Indices.X in indexers" statements down the line. 2.) Converts the dict keys from Indices to their string values which is needed to call xarray.sel(indexers). 3.) converts tuple range values to slice objects. Do you mean I should just add all this code to the sel method?

starfish/test/image/test_imagestack_index.py
shanaxel42: I'll still need to create Imagestacks with coordinates if I'm just testing calc_new_physical_coords,  which I think is what takes up a lot of the complexity...the test flow is just 1.) Create an imagestack 2.) index it 3.) verify that the new shape is what we expect. And then for  test_indexing_by_x_y 1.) create and image stack with simple coords 2.) index it 3.)verify that the new shape and coords are what we were expecting. I tried to add a lot of comments to explain what was going on, but are you saying I shouldn't test the full indexing of an imagestsck at all? and just create an imagestack with coords and verify that calc_new_coords is working?

starfish/imagestack/imagestack.py
shanaxel42: Does this mean move this to it's own directory?

starfish/imagestack/imagestack.py
shanaxel42: Also if we ever did want to add indexing with kwargs or another ux for indexing it would be easier to have this separate to do all the index parsing.

starfish/imagestack/imagestack.py
shanaxel42: Is there a specific place you were thinking? Or should I add a new file to the util directory

starfish/imagestack/physical_coordiante_calculator.py
ttung: Ahh, I see.  I totally misread the code.  I think that furthers the argument for moving `index_keep_coordinates` out of `ImageStack`.

You're totally right that we need to slice both the image data and the coordinates.  However, even after better understanding the code, I think this organization is confusing.

starfish/imagestack/imagestack.py
ttung: So you should put the explanation in the docblock.  And possibly inline this in the `sel` method.

starfish/test/image/test_imagestack_index.py
```diff
@@ -0,0 +1,134 @@
...
+    expected_shape = OrderedDict([(Indices.ROUND, 1), (Indices.CH, 1),
+                                  (Indices.Z, 1), (Indices.Y, 190), (Indices.X, 190)])
+    assert indexed_stack.shape == expected_shape
+    expected_coords = {Coordinates.X: (1, 1.955),
+                       Coordinates.Y: (4, 5.91),
+                       Coordinates.Z: (1, 3)}
+    check_coodinate_values(indexed_stack, expected_coords)
+
+
+def check_coodinate_values(stack, expected_coords):
```
ttung: s/check_coodinate_values/check_coordinate_values/

starfish/imagestack/imagestack.py
ttung: > Also if we ever did want to add indexing with kwargs or another ux for indexing it would be easier to have this separate to do all the index parsing.

understood.  then just a request for docblock.

starfish/imagestack/imagestack.py
ttung: No, there is a typo. :)

starfish/imagestack/imagestack.py
ttung: starfish/imagestack/??? 

I'm truly awful at naming things, so I won't even try. :)

starfish/test/image/test_imagestack_index.py
ttung: I think there's value in making `calc_new_physical_coords` easily testable (and indeed build tests for that).  And there's value in testing the `sel` functionality.  Both are very crucial components that deserve some tests.

starfish/imagestack/imagestack.py
shanaxel42: OHHH my bad haha

starfish/imagestack/imagestack.py
shanaxel42: haha hmmm indexing_utils?

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    for _round in range(coords_array.sizes[Indices.ROUND]):
+        for ch in range(coords_array.sizes[Indices.CH]):
+            for z in range(coords_array.sizes[Indices.Z]):
+                indices = {
+                    Indices.ROUND: _round,
+                    Indices.CH: ch,
+                    Indices.Z: z
+                }
+                xmin, xmax = _recalculate_physical_coordinate_range(
```
ttung: I have a preference for:

```suggestion
                xmin, xmax, ymin, ymax = coords_array[indices].loc[
                    dict(physical_coordinate=[
                        PhysicalCoordinateTypes.X_MIN,
                        PhysicalCoordinateTypes.X_MAX,
                        PhysicalCoordinateTypes.Y_MIN,
                        PhysicalCoordinateTypes.Y_MAX
                    ])]
                xmin, xmax = _recalculate_physical_coordinate_range(
                    xmin, xmax,
                    stack_shape[Indices.X.value],
                    indexers[Indices.X.value])
                ymin, ymax = _recalculate_physical_coordinate_range(
                    ymin, ymax,
                    stack_shape[Indices.Y.value],
                    indexers[Indices.Y.value])
                coords_array[indices].loc[
                    dict(physical_coordinate=[
                        PhysicalCoordinateTypes.X_MIN,
                        PhysicalCoordinateTypes.X_MAX,
                        PhysicalCoordinateTypes.Y_MIN,
                        PhysicalCoordinateTypes.Y_MAX
                    ])] = [xmin, xmax, ymin, ymax]
```

And `_recalculate_physical_coordinate_range` would accept the following arguments:
1. physical_coordinate_start
2. physical_coordinate_end
3. pixel_size
4. pixel_slice

Rationale: when I do cropping _before_ `ImageStack` is loaded, I probably won't have `coords_array`.  If you structure `_recalculate_physical_coordinate_range` this way, I can use it without modification.  Probably... :)

starfish/test/image/test_physical_coordinate_calculator.py
```diff
@@ -0,0 +1,184 @@
...
+                num_pixels=self.stack_shape[Indices.Y])
+
+        # Index on single value of X, range of y
+        self.indexers[Indices.Y], self.indexers[Indices.X] = 100, slice(None, 100)
+        new_coords = physical_coordinate_calculator.calc_new_physical_coords_array(
+            physical_coordinates=self.coords_array,
+            stack_shape=self.stack_shape,
+            indexers=self.indexers)
+
+        expected_xmin = physical_coordinate_calculator.calculate_physical_pixel_value(
```
ttung: I wonder if it makes sense to calculate these by hand and store the values in the test file.

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +292,43 @@ def xarray(self) -> xr.DataArray:
         """Retrieves the image data as an xarray.DataArray"""
         return self._data
 
+    def sel(self, indexers: Mapping[Indices, Union[int, tuple]]=None):
+        """Given a dictionary mapping the index name to either a value or a slice range, return an
+        Imagestack with each dimension indexed accordingly
+
+        Parameters
+        ----------
+        indexers : Dict[Indices, (int/tuple)]
```
ambrosejcarr: is (int/tuple) valid mypy? If so, I love that.

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +292,43 @@ def xarray(self) -> xr.DataArray:
...
+            A dictionary of dim:index where index is the value or range to index the dimension
+
+        Examples
+        --------
+        Imagestack stack with shape (5, 5, 15, 200, 200)
```
ambrosejcarr: I'd love these to be real code examples (see e.g. https://github.com/spacetx/starfish/blob/master/starfish/intensity_table/intensity_table.py#L42)

starfish/imagestack/imagestack.py
```diff
@@ -290,6 +292,43 @@ def xarray(self) -> xr.DataArray:
...
+        and the imagestack's physical coordinates
+        xarray also indexed and recalculated according to the x,y slicing.
+
+        Returns
+        -------
+        ImageStack :
+            a new image stack indexed by given value or range.
+        """
+
+        # convert indexers to dict <str:(int, slice) format
```
ambrosejcarr: I'm not sure what `<str` refers to.

starfish/imagestack/indexing_utils.py
```diff
@@ -0,0 +1,36 @@
+import xarray as xr
+
+from starfish.types import Indices
+
+
+def convert_to_indexers_dict(indexers):
```
ambrosejcarr: indexers needs a type and this function needs a return value.

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,173 @@
...
+                                           tile_indices: Mapping[Indices, int],
+                                           key
+                                           ) -> Tuple:
+    """Given the dimension size and pixel coordinate indexes calculate the corresponding
+    coordinates in physical space
+
+    Parameters
+    ----------
+    coords_array: xarray
+        xarray that holds the min/max values of physical coordinates per tile (Round, Ch, Z).
```
ambrosejcarr: List dimensions here for clarity: (3, 2) or (2, 3)?

starfish/imagestack/imagestack.py
shanaxel42: A typo!

starfish/imagestack/physical_coordinate_calculator.py
shanaxel42: OH this is a really good call.

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,173 @@
...
+def calculate_physcial_pixel_size(coord_max, coord_min, num_pixels):
+    """Calculate the size of a pixel in physical space"""
+    return (coord_max - coord_min) / num_pixels
+
+
+def calculate_physical_pixel_value(physcial_pixel_size, index, start_of_range):
+    """Calculate the physical pixel value at the given index"""
+    if index:
+        # Check for negative index
+        if index < 0:
```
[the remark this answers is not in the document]
shanaxel42: I think it's kinda nice to have for not too much code. I always use -1 to get the last item of something when I don't know how long it is. @ambrosejcarr do you have an opinion?

starfish/imagestack/imagestack.py
shanaxel42: I think I found somewhere that's official sphinx format...let me try and find it

starfish/test/image/test_physical_coordinate_calculator.py
shanaxel42: haaaaa I had a version of this file that did that but thought you might tell me to just use the method. Predicted wrong :(

starfish/imagestack/physical_coordinate_calculator.py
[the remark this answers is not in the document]
ttung: I think that makes sense, but wouldn't your calculation then require the range of your space (in px)?

starfish/test/image/test_physical_coordinate_calculator.py
ttung: I can see advantages to both, which is why I'm just wondering! :)

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,195 @@
...
+    stack_shape : Dict[str, int]
+        The shape of the image tensor by categorical index (channels, imaging rounds, z-layers)
+    indexers : Dict[str, (int/slice)]
+        A dictionary of dim:index where index is the value or range to index the dimension
+
+    Returns
+    -------
+    A coordinates xarray indexed by R, CH, V and values recalculated according to indexing on X/Y
+    """
+    new_coords = physical_coordinates.copy()
```
ttung: Would prefer an explicit deepcopy (i.e., `.copy(deep=True)` or `copy.deepcopy(physical_coordinates)`

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,195 @@
...
+           Indices.CH.value: indexers[Indices.CH.value],
+           Indices.Z.value: indexers[Indices.Z.value]}
+    new_coords = indexing_utils.index_keep_dimensions(new_coords, key)
+    # check if X or Y dimension indexed, if so recalculate physcal coordinate min/max values
+    if _needs_coords_recalculating(indexers[Indices.X.value], indexers[Indices.Y.value]):
+        _recalculate_physical_coordinate_ranges(stack_shape, indexers, new_coords)
+    return new_coords
+
+
+def _needs_coords_recalculating(x_indexers, y_indexers) -> bool:
```
ttung:
```suggestion
def _needs_coords_recalculating(x_indexers: Union[int, slice], y_indexers: Union[int, slice]) -> bool:
```

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,195 @@
...
+                coords_array[indices].loc[
+                    dict(physical_coordinate=[
+                        PhysicalCoordinateTypes.X_MIN,
+                        PhysicalCoordinateTypes.X_MAX,
+                        PhysicalCoordinateTypes.Y_MIN,
+                        PhysicalCoordinateTypes.Y_MAX
+                    ])] = [xmin, xmax, ymin, ymax]
+
+
+def _calculate_physcial_pixel_size(coord_max: float, coord_min: float, num_pixels: int):
```
ttung: would prefer `Number` in place of `float`

ttung: and return value of `Number`.

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -0,0 +1,195 @@
...
+                                           dimension_size: int,
+                                           indexer
+                                           ) -> Tuple[Number, Number]:
+    """Given the dimension size and pixel coordinate indexes calculate the corresponding
+    coordinates in physical space
+
+    Parameters
+    ----------
+    coord_min: float
+        the minimun physical coordinate value
```
ttung:
```suggestion
        the minimum physical coordinate value
```

starfish/test/image/test_physical_coordinate_calculator.py
```diff
@@ -0,0 +1,184 @@
...
+from collections import OrderedDict
+
+import numpy as np
+import xarray as xr
+
+from starfish.imagestack import physical_coordinate_calculator
+from starfish.types import Indices, PHYSICAL_COORDINATE_DIMENSION, PhysicalCoordinateTypes
+
+
+class TestPhysicalCoordinateCalculator(unittest.TestCase):
```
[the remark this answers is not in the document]
ttung: Also, @ambrosejcarr do you have any concerns about `unittest` tests mixed in with `pytest` tests?

starfish/test/image/test_physical_coordinate_calculator.py
```diff
@@ -0,0 +1,132 @@
...
+                      self.physical_coords[PhysicalCoordinateTypes.X_MAX],
+                      self.physical_coords[PhysicalCoordinateTypes.Y_MIN],
+                      self.physical_coords[PhysicalCoordinateTypes.Y_MAX],
+                      self.physical_coords[PhysicalCoordinateTypes.Z_MIN],
+                      self.physical_coords[PhysicalCoordinateTypes.Z_MAX]])
+
+        self.stack_shape = OrderedDict([(Indices.ROUND, 1), (Indices.CH, 1),
+                                        (Indices.Z, 1), (Indices.Y, 200), (Indices.X, 200)])
+
+    def test_calc_new_physical_coords_array(self):
```
ttung: these tests are great.  would split each of these cases into its own method.

PR #712, opened 2018-10-15
Reimplement relabel

ambrosejcarr:
- Re-implements `re-label` and moves it into `watershed.py`. 
- Removes `munge.py`, which no longer has functions that are used by starfish codebase. 

Testing strategy: 
- sorry, I just compared the new and old strategies in the ISS notebook and asserted that the outcomes were the same. They are. 

image


starfish/image/_segmentation/watershed.py
```diff
@@ -321,6 +321,28 @@ def watershed(self, markers: np.ndarray, watershed_mask: np.ndarray) -> np.ndarr
...
+            not be sequential integers.
+
+        Returns
+        -------
+        image : np.ndarray[np.uint32]
+            same as input, but the values are re-labled as sequential integers
+        num_labels : int
+            number of unique objects
+        """
+        output = np.zeros_like(image)
```
ttung: empty_like?

starfish/image/_segmentation/watershed.py
```diff
@@ -321,6 +321,28 @@ def watershed(self, markers: np.ndarray, watershed_mask: np.ndarray) -> np.ndarr
...
+        for i, v in enumerate(np.unique(image)):
```
ttung: also, [here](https://stackoverflow.com/questions/16992713/translate-every-element-in-numpy-array-according-to-key)'s some people benchmarking alternate solutions.  the vectorize solution might be applicable to our use case.

starfish/image/_segmentation/watershed.py
ambrosejcarr: The "not an object" default for each pixel is zero, so `np.zeros` is the correct method for this.

starfish/image/_segmentation/watershed.py
ambrosejcarr: Here's my translation of the vectorized version for our use case. 

```python3
In [11]: def relabel_vectorize(image):
    ...:     keys = np.unique(image)
    ...:     mapping = dict(zip(keys, range(len(keys))))
    ...:     return np.vectorize(mapping.__getitem__)(image)
```

and mine: 
```python3
In [10]: def relabel_image(image: np.ndarray) -> np.ndarray:
    ...:     output = np.empty_like(image)
    ...:     for i, v in enumerate(np.unique(image)):
    ...:         output[np.where(image == v)] = i
    ...:     return output, i
```

Tested against my implementation on a poisson-generated random data
```python
In [15]: a = np.random.poisson(0.5, size=1000000).reshape(1000, 1000)

In [16]: %timeit relabel_image(a)
62.9 ms ± 823 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [17]: %timeit relabel_vectorize(a)
191 ms ± 3.56 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [18]: a = np.random.poisson(0.5, size=10000000).reshape(1000, 10000)

In [19]: %timeit relabel_image(a)
694 ms ± 16.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [20]: %timeit relabel_vectorize(a)
1.97 s ± 16.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

The latter size is similar to our standard case, so I think we can safely stick with the loop case (which is also clearer to understand, I think).

PR #713, opened 2018-10-16
docker: Pass ignore-installed for conda

joshmoore:
Installing starfish in conda now fails with:

```
Cannot uninstall 'certifi'. It is a distutils installed project and thus we
cannot accurately determine which files belong to it which would lead to only a
partial uninstall.
```

Temporarily ignoring the existing packages in docker since we can assume they
should be there.

see: https://github.com/pypa/pip/issues/5247 (et al)


joshmoore:
Currently testing at https://travis-ci.org/spacetx/starfish/builds/442216878

joshmoore:
Likely to close this with gh-719 merged. First, though, I'll look into how bioconda reacts to having `certifi` pinned.

ttung:
I kind of like this solution more.  Which version is used when `--ignore-installed` is specified?

joshmoore:
`conda create -n tmp2 python=3.6` (version 4.5.10) uses 2018.10.15-py36_0. Updating to 4.5.11 (`conda update -n base -c defaults conda`)  does the same. (Pip 10.0.1 for reference)

Running `pip install starfish` in either then leads to:

```
(tmp2) $ pip install starfish==0.0.27
...
Collecting certifi==2018.8.24 (from starfish)
  Using cached https://files.pythonhosted.org/packages/df/f7/04fee6ac349e915b82171f8e23cee63644d83663b34c539f7a09aed18f9e/certifi-2018.8.24-py2.py3-none-any.whl
...
Cannot uninstall 'certifi'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

(Same for 0.0.26 and 0.0.25)

ttung:
I am pretty docker and conda un-savvy. Does this mean `--ignore-installed` does not work?

joshmoore:
No, it worked.

PR #716, opened 2018-10-16
IntensityTable concatenate

ambrosejcarr:
Enables concatenation of `IntensityTable` objects in cases where: 
- the experiment has multiple fields of view
- the experiment processed rounds separately

These changes _should_ be future-proof in terms of @shanaxel42 's impending coordinate changes.


starfish/intensity_table/concatenate.py
```diff
@@ -0,0 +1,34 @@
...
+    Returns
+    -------
+        merged IntensityTable. Missing values are filled with np.NaN
+
+    See Also
+    --------
+    Sparse Arrays in xarray: https://github.com/pydata/xarray/issues/1375
+    Combine_first: http://xarray.pydata.org/en/stable/combining.html#combine
+    """
+    return xr.concat(list(intensity_tables), Features.AXIS)
```
shanaxel42: Oh this is chill.

starfish/intensity_table/concatenate.py
```diff
@@ -0,0 +1,34 @@
...
+def concatenate(intensity_tables: Iterable[IntensityTable]) -> IntensityTable:
+    """Concatenate IntensityTables produced for different fields of view or across imaging rounds
+
+    IntensityTables are concatenated along the features axis, and the resulting arrays are stored
+    densely, even if the underlying data is sparse, since xarray does not yet support sparse array
+    structures. This means that spots that are identified in different rounds and channels will
+    be identified as separate features, even if they have exactly identical coordinates.
+
+    To merge spots that share coordinates across rounds and channels into single features amenable
+    to decoding, use IntensityTable.combine_first()
```
shanaxel42: How come you didn't want to just add IntensityTable.combine_first() at the end of this method?

ambrosejcarr: Because I don't think we really need that use case -- I would advocate for a separate function, otherwise it could be a bit confusing what this is designed to do.

PR #718, opened 2018-10-17
remove spurious merge markers

ttung:


ttung:
Landing this because it seems simple enough.... :)

PR #722, opened 2018-10-18
Click migration

joshmoore:
The general call flow here (which likely needs documenting somewhere for devs) is:

 1. `python -m starfish` imports `starfish.starfish` which:
    - imports (if they aren't already) pipeline components (filter, decoder, etc.) which each import their concrete algorithms causing them to be initiated and prepared for registering via the new `PipelineCompenent._cli_register` method
    - defines `starfish.starfish.starfish` which is now a `click.group`
    - registers commands using the `@starfish.command` decorator
    - registers other commands using the `starfish.add_command` method, each of which already has its subcommands from the use of `_cli_register`
2. invocation begins with calling sf.sf.sf; when it completes
3. each subcommand executes from left to right (if a cleanup method is needed, then use `ctx.call_on_close`)

In the concrete example of a pipeline component like `filter` and an algorithm like `clip` invoked using:

```
starfish ... filter ... clip ...
```

 * `starfish` is only responsible for `--profile` (at the moment)
 * `filter` takes `--input` and `--output` via the `_cli()` method at the *bottom* of the module
 * `filter._cli` checks & opens resources as necessary, then sets them as a dictionary in `ctx.obj`
 * `clip._cli` is then called (also defined at the bottom of its module)
 * `clip._cli` is mostly responsible for the instantiation of `Clip`.
 * `clip._cli` then delegates itself to the classmethod `Filter._cli_run` as `instance`
 * `filter._cli_run` gathers the inputs from `ctx.obj`, calls `instance.run`, and saves any results to `output` as necessary.

In this PR:
 * [x] Remove all uses of `argparse` in `starfish/*` (except starfish.util.argparse)
 * [x] i.e. *not* updating `examples/get_*` for the moment
 * [x] Add a `version` command
 * [x] Add a mechanism `command.noart = True` to prevent artwork from being printed
 * [x] Add support for `python -m starfish` invocations to simplify testing
 * [x] Replace the argparse.FsExistsType

Out-of-scope:
 * Updating uses of `print()` etc. for better click integration
 * Integrating click with mypy

Test plan: 
 * `python -svxk cli`
 * General usability of `starfish ...`


ambrosejcarr:
>  Replace the argparse.FsExistsType

On this point, I believe click has a built-in type to support this: https://click.palletsprojects.com/en/7.x/arguments/#file-arguments

joshmoore:
> On this point, I believe click has a built-in type to support this: 

Definitely. My question was whether it was intended that the `--input` arguments could be remote, in which case a test is needed for path vs. url.

joshmoore:
Ah, I was assuming something more complex. `starfish.util.FsExistsType` exists since `argparse.FileType` returns an open file as opposed to a path. Pushed.

starfish/image/_filter/__init__.py
```diff
@@ -19,38 +19,30 @@
...
-
-        print('Filtering images ...')
-        stack = ImageStack.from_path_or_url(args.input)
-        instance = args.filter_algorithm_class(**vars(args))
-        output = instance.run(stack)
-        output.write(args.output)
+    def _cli_run(cls, ctx, instance):
+        output = ctx.obj["output"]
+        stack = ctx.obj["stack"]
+        instance.run(stack)
```
ambrosejcarr:
```suggestion
        filtered = instance.run(stack)
        filtered.write(output)
```

ambrosejcarr: I think this should fix the dartfish bug.

ambrosejcarr:
Passing all tests: https://github.com/spacetx/starfish/tree/ajc-click-migration

joshmoore:
@ambrosejcarr : green.

ttung:
Can this be rebased?  There's a lot of changes in the segmentation code that I feel probably doesn't belong...

joshmoore:
> Can this be rebased? There's a lot ...

Not sure why GitHub is showing the changes from the merges of master. A result of cooperating via merges, I assume. I didn't think the time spent on rebasing was going to be worth it, so I've opened a new PR with this context squash-merged. Alternatively, `git diff origin/master...josh/click-migration` should show you the correct diff for this PR.

PR #733, opened 2018-10-23
Fix docs build

ambrosejcarr:
To fix the docs build we need to: 
1. install from `REQUIREMENTS-DEV.txt`
2. update `REQUIREMENTS-DEV.txt` to include new modules (m2r)

Testing strategy: 
1. [Docs](https://spacetx-starfish.readthedocs.io/en/ajc-docs-fix/) build on this branch
2. Latest build passes: https://readthedocs.org/projects/spacetx-starfish/builds/7988751/


setup.py
```diff
@@ -14,7 +14,9 @@
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
 
-install_requires = [line.rstrip() for line in open(os.path.join(os.path.dirname(__file__), "REQUIREMENTS.txt"))]
+install_requires = [
+    line.rstrip() for line in open(os.path.join(os.path.dirname(__file__), "REQUIREMENTS-DEV.txt"))
```
ttung: If we do this, what's the point of `REQUIREMENTS.txt` and `REQUIREMENTS-DEV.txt`?  Why not just trash the former?

ambrosejcarr: Agree, it loses its importance. How do you feel about this change?

ttung: I still want starfish-lite and starfish-ezmode. :)

ambrosejcarr: Ok. Practically, how do we get the docs building again in the short term? Do you want me to rename `REQUIREMENTS-DEV.txt` and remove `REQUIREMENTS.txt`?

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
 max-line-length=100
-ignore: E252, E301, E302, E305, E731, F811
+ignore: E252, E301, E302, E305, W503, W605, E731, F811
```
ambrosejcarr: W503 & W504 are mutually exclusive, one must be ignored: 
W503: binary operator must come before break
W504: binary operator must come after break. 

pep8 prefers W504, so we ignore W503

W605 breaks on the starfish art. 

E252 complains if you don't have spaces around equals sign in parameter kwargs.

ttung: If W605 breaks only on the starfish art, let's just fix that.

ttung: Also, my guess is that the warnings in this file are not useful any more, since we specify them on the command line.

ambrosejcarr:
Fixed the art, didn't understand your comment about warnings.

ttung:
I think the warnings exclusions on the cli command (in the Makefile) supercede the warnings exclusions in `.flake8`

ambrosejcarr:
So I removed it, and that did indeed work. I'm confused why we have both files now, though. The Makefile is just a superset. What's the logic here?

.flake8
ttung: #780

PR #736, opened 2018-10-24
Fix docs for apply and transform.

ttung:
I suck with words, so would appreciate tips on improving the verbiage.

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -62,8 +62,11 @@ class ImageStack:
         retrieve a slice of the image tensor
     set_slice(indices, data, axes=None)
         set a slice of the image tensor
-    apply(func, is_volume=False, in_place=False, verbose=False, n_processes=None)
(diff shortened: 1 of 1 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -614,24 +617,24 @@ def _iter_tiles(
...
+            func: Callable,
+            split_by: Set[Indices]=None,
             in_place=False,
             verbose: bool=False,
             n_processes: Optional[int]=None,
             **kwargs
     ) -> "ImageStack":
-        """Apply func over all tiles or volumes in self
+        """Apply func over all N-dimensional chunks of this image, yielding identically shaped
+        chunks, which are constituted into an ImageStack and returned.
```
shanaxel42: Oh man not sure about "chunks" what if it was just "Apply func over all N-dimensions of this image, and return the resulting imagestack"

ttung: See #740 :)

PR #738, opened 2018-10-25
Tests for passing a multiprocessing.Array backing a numpy array between processes

ttung:
1. Added the shmem module, which wraps some very ugly logic necessary to pass multiprocessing.Array objects between processes.
2. Add tests for a multiprocessing.Array backing a numpy array to be passed between processes.

Test plan: wrote tests!

### DIFF
--- a/starfish/multiprocessing/shmem.py
+++ b/starfish/multiprocessing/shmem.py
@@ -1,0 +1,41 @@
+"""
+Because multiprocessing.Pool demands that any shared-memory constructs be passed in as initializer
+arguments, and not as arguments to multiprocessing.Pool.map or multiprocessing.Pool.apply, we need a
+stub initializer that accepts the shared-memory construct and store it in a global for retrieval.
+
+Does not work:
+  arrays = [multiprocessing.Array(..)]
+  with multiprocessing.Pool() as pool:
+    pool.map(some_func, arrays)
+
+Does work:
+  array = multiprocessing.Array(..)
+  data = range(10)
+  with multiprocessing.Pool(initializer=some_function, initargs=(array,)) as pool:
+    pool.map(some_func, data)
+
+This wraps the ugliness of a global variable inside a class that's only used for this purpose.  An
+example of how this might be used would be:
+
+  def worker(val):
+    array = SharedMemory.get_payload()
+    # do something with the shared memory
+
+  array = multiprocessing.Array(..)
+  data = range(10)
+  with multiprocessing.Pool(initializer=SharedMemory.initalizer, initargs=(array,)) as pool:
+    pool.map(worker, data)
+"""
+from typing import Any
+
+
+class SharedMemory:
+    _payload = None
+
+    @staticmethod
+    def initializer(payload: Any) -> None:
+        SharedMemory._payload = payload
+
+    @staticmethod
+    def get_payload() -> Any:
+        return SharedMemory._payload
--- a/starfish/test/test_multiprocessing.py
+++ b/starfish/test/test_multiprocessing.py
@@ -1,0 +1,138 @@
+"""
+These tests ensure that we can share an np-array or something that encapsulates an np-array among
+processes.  The data is shared such that any process can write to the array and it is visible to all
+the other processes.  It is up to the developer that uses this mechanism to ensure data integrity is
+maintained.
+"""
+
+
+import ctypes
+import multiprocessing
+from functools import partial
+# Even though we import multiprocessing, mypy can't find the Array class.  To avoid sprinkling
+# ignore markers all over the file, we explicitly import the symbol and put the ignore marker here.
+from multiprocessing import Array as mp_array  # type: ignore
+from typing import Any, Callable
+
+import numpy as np
+
+from starfish.multiprocessing.shmem import SharedMemory
+
+
+def test_numpy_array(nitems: int=10):
+    """
+    Try to share a numpy array directly.  This should fail as numpy is by default a copy-on-write
+    object.  The worker process will write to array and we should not see the changes.
+    """
+    array = np.zeros(shape=(nitems,), dtype=np.uint8)
+    _start_process_to_test_shmem(
+        array_holder=array,
+        decoder=_decode_numpy_array_to_numpy_array,
+        nitems=nitems)
+    for ix in range(nitems):
+        assert array[ix] == 0
+
+
+def _decode_numpy_array_to_numpy_array(array):
+    """
+    When passing naked numpy arrays into subprocesses, no work is needed to retrieve the array. See
+    :py:method:`_decode_array_to_numpy_array` for an example where decoding is necessary.
+    """
+    return array
+
+
+def test_shmem_numpy_array(nitems: int=10):
+    """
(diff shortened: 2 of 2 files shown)


starfish/test/test_multiprocessing.py
```diff
@@ -0,0 +1,113 @@
+"""
+These tests ensure that we can share an np-array or something that encapsulates an np-array among
+processes.  The data is shared such that any process can write to the array and it is visible to all
+the other processes.  It is up to the developer that uses this mechanism to ensure data integrity is
```
ambrosejcarr: Any suggestions about how a developer would ensure this? Could be useful to include here.

starfish/test/test_multiprocessing.py
```diff
@@ -0,0 +1,113 @@
...
+    array = decoder(shmem.get_payload())
+    array[position] = position
+
+
+def _start_process_to_test_shmem(
+        array_holder: Any,
+        decoder: Callable[[Any], np.ndarray],
+        nitems: int) -> None:
+    """
+    Starts a process and passes an object to its initializer.  The object is passed to
```
ambrosejcarr: Would replace docs with the following, which is a bit more verbose about the rationale for implementing this function. 

```
    This method leverages multiprocessing.Pool to test process initializers and 
    shared memory array decoders. 

    Starts a process Pool and passes :py:method:`shmem.initializer` as its initializer, 
    which will store `array_holder` as a global variable inside the initialized process. 
    
    Creates `bound_func`, which will be executed by the subprocesses of Pool. 
    `bound_func` is a partial function constructed from 
    :py:method:`_write_to_shared_memory_array` that exposes `decoder`, a method 
    that gets a numpy array out of array_holder and then writes a value to it 
    (inside the spawned process). 

    This enables writing of tests that check whether the written values are available
    in the parent processes, which depends on `shmem.initializer`, `array_holder`, 
    and `decoder`.  
```

starfish/test/test_multiprocessing.py
```diff
@@ -0,0 +1,113 @@
...
+from starfish.multiprocessing import shmem
+
+
+def test_numpy_array(nitems: int=10):
+    """
+    Try to share a numpy array directly.  This should fail as numpy is by default a copy-on-write
+    object.  The worker process will write to array and we should not see the changes.
+    """
+    array = np.zeros(shape=(nitems,), dtype=np.uint8)
+    _start_process_to_test_shmem(array, _decode_numpy_array_to_numpy_array, nitems)
```
ambrosejcarr: suggestion: use `**kwargs` in functions defined here, this corner of python is a bit circuitous, naming things will make coming back to these tests later a bit easier

starfish/test/test_multiprocessing.py
```diff
@@ -0,0 +1,113 @@
...
+    for ix in range(nitems):
+        assert array[ix] == 0
+
+
+def _decode_numpy_array_to_numpy_array(array):
```
ambrosejcarr:
```suggestion
def _decode_numpy_array_to_numpy_array(array):
    """
    When passing naked numpy arrays into subprocesses, no work is needed to retrieve the 
    array. See :py:method:`_decode_array_to_numpy_array` for an example where decoding is 
    necessary.
    """

File: starfish/test/test_multiprocessing.py [hunk 5]
```diff
@@ -0,0 +1,113 @@
...
+    buffer = mp_array(ctypes.c_uint8, nitems)
+    array = _decode_array_to_numpy_array(buffer)
+    array.fill(0)
+    _start_process_to_test_shmem(
+        TestWrappedArray(buffer), _decode_wrapped_array_to_numpy_array, nitems)
+    for ix in range(nitems):
+        assert array[ix] == ix
+
+
+def _decode_wrapped_array_to_numpy_array(wrapped_array):
```
ambrosejcarr:
```suggestion
def _decode_wrapped_array_to_numpy_array(wrapped_array) -> np.ndarray:
    """Reconstitute a numpy array from a shared memory buffer."""
```

ambrosejcarr: Could wrapped array have a type?

starfish/multiprocessing/shmem.py
```diff
@@ -0,0 +1,19 @@
...
+stub initializer that accepts the shared-memory construct and store it in a global for retrieval.
+This ugly bit of logic is best kept away from other code.
+"""
+from typing import Any
+
+
+_global = None
+
+
+def initializer(payload: Any) -> None:
```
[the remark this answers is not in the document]
ttung: Will do.

starfish/multiprocessing/shmem.py
```diff
@@ -0,0 +1,19 @@
...
+    global _payload
+    _payload = payload  # type: ignore
+
+
+def get_payload() -> Any:
```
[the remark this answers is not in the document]
ttung: Will fix.

starfish/test/test_multiprocessing.py
ttung: I can't think of any general guidelines to ensure this. :(

starfish/test/test_multiprocessing.py
ttung: I don't understand this suggestion.

starfish/test/test_multiprocessing.py
ttung: Will apply.

starfish/test/test_multiprocessing.py
ttung: Will apply, and add a type to `wrapped_array`

starfish/test/test_multiprocessing.py
```diff
@@ -0,0 +1,113 @@
...
+    return np.frombuffer(wrapped_array.array.get_obj(), dtype=np.uint8)
+
+
+def _applied_func(decoder: Callable[[Any], np.ndarray], position: int) -> None:
```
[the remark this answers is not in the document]
ttung: will do.

starfish/test/test_multiprocessing.py
ttung: Applied.

starfish/test/test_multiprocessing.py
ambrosejcarr: Ok. No problem.

starfish/test/test_multiprocessing.py
ambrosejcarr: Sorry, I meant: 

```suggestion
    _start_process_to_test_shmem(
        array_holder=array, 
        decoder=_decode_numpy_array_to_numpy_array, 
        nitems=nitems
    )
```

ambrosejcarr:
Love this, thanks for adding the comments. 

... when (if?) we have free time, this would make for a nice blog post. The documentation on how to use multiprocessing with numpy is really bad.

ttung:
> ... when (if?) we have free time, this would make for a nice blog post. The documentation on how to use multiprocessing with numpy is really bad.

I look forward to you ghostwriting that blog post.

starfish/test/test_multiprocessing.py
ttung: Applied.

PR #739, opened 2018-10-25
Make MPDataArray actually shared-memory

ttung:
https://github.com/spacetx/starfish/pull/737 created the wrapper class for xr.DataArray, but fundamentally, it was still a copy-on-write construct that did not allow multiple writers across processes to operate on it.  This swaps out the underpinnings with the constructs introduced and tested in https://github.com/spacetx/starfish/pull/738, such that processes that we fork out in `ImageStack` can cooperatively work on multiple parts of the tensor concurrently.

Whenever we construct a `MPDataArray`, we now construct a `multiprocessing.Array`, and layer a `numpy` object on top of that memory.  Then we construct a `xr.DataArray` around that `numpy` object.  Fun, right?

We freely copy xr.DataArray around when we want to run `ImageStack.apply` not in-place, so we have a `__deepcopy__` method in `MPDataArray` to copy correctly (i.e., create a new `multiprocessing.Array`, and layer everything on top of that).

We also intercept `xr.DataArray``s copy method, such that if someone attempts to deepcopy the xarray directly, we warn them that it has consequences.

Test plan: Added a few tests:
1. a test that exercises `xr_deepcopy`
2. a test that exercises `copy.deepcopy(..)` on an ImageStack.
3. a test that exercises `copy.deepcopy(..)` on an ImageStack's xarray (which will work but will trigger a warning).

Depends on #737, #738

### DIFF
--- a/starfish/imagestack/_mp_dataarray.py
+++ b/starfish/imagestack/_mp_dataarray.py
@@ -1,8 +1,12 @@
-import copy
-from typing import Sequence
+import functools
+import warnings
+from collections import OrderedDict
+from multiprocessing import Array as mp_array  # type: ignore
+from typing import Callable, Sequence, Tuple
 
 import numpy as np
 import xarray as xr
+from xarray import Variable
 
 from starfish.types import Number
 
@@ -19,18 +23,20 @@ class MPDataArray:
     multiprocessing.Array object back out of the numpy array or the xarray.  Therefore, we need to
     explicitly maintain a reference to it and keep the two items together.
     """
-    def __init__(self, data: xr.DataArray) -> None:
+    def __init__(self, data: xr.DataArray, backing_mp_array: mp_array) -> None:
         self._data = data
+        self._backing_mp_array = backing_mp_array
 
     @classmethod
     def from_shape_and_dtype(
             cls, shape: Sequence[int], dtype, initial_value: Number=None, *args, **kwargs
     ) -> "MPDataArray":
-        np_array = np.empty(shape=shape, dtype=dtype)
+        np_array, backing_mp_array = np_array_backed_by_mp_array(shape, dtype)
         if initial_value is not None:
             np_array.fill(initial_value)
         xarray = xr.DataArray(np_array, *args, **kwargs)
-        return MPDataArray(xarray)
+        xarray.copy = functools.partial(replacement_copy, xarray.copy)
+        return MPDataArray(xarray, backing_mp_array)
 
     @property
     def data(self) -> xr.DataArray:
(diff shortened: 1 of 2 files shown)


starfish/imagestack/_mp_dataarray.py
```diff
@@ -39,4 +46,55 @@ def __setitem__(self, key, value):
...
+    )
+
+    return result, backing_array
+
+
+def replacement_copy(orig_copy: Callable, deep=True):
+    if deep:
+        warnings.warn(
+            "calling deepcopy with ImageStack's xarrays results in arrays that cannot be shared "
+            "across processes.  use mp_dataarray.xr_deepcopy if you need that behavior."
```
ambrosejcarr:
```suggestion
            "across processes.  Use mp_dataarray.xr_deepcopy if you need that behavior."
```

starfish/imagestack/_mp_dataarray.py
```diff
@@ -12,18 +17,20 @@ class MPDataArray:
...
     def from_shape_and_dtype(
             cls, shape: Sequence[int], dtype, initial_value: Number=None, *args, **kwargs
     ) -> "MPDataArray":
-        np_array = np.empty(shape=shape, dtype=dtype)
+        np_array, backing_mp_array = np_array_backed_by_mp_array(shape, dtype)
         if initial_value is not None:
             np_array.fill(initial_value)
         xarray = xr.DataArray(np_array, *args, **kwargs)
-        return MPDataArray(xarray)
+        xarray.copy = functools.partial(replacement_copy, xarray.copy)
```
ambrosejcarr: Nice. 👍

starfish/test/test_multiprocessing.py
```diff
@@ -81,6 +84,64 @@ def _decode_wrapped_array_to_numpy_array(wrapped_array):
...
+        assert np.allclose(imagestack_copy.xarray[ix], ix)
+
+
+def _decode_imagestack_array_to_numpy_array(
+        shape: Tuple[int, ...], dtype, buffer) -> np.ndarray:
+    unshaped_np_array = np.frombuffer(buffer.get_obj(), dtype=dtype)
+    return unshaped_np_array.reshape(shape)
+
+
+def test_imagestack_xarray_deepcopy(nitems: int=10) -> None:
```
ambrosejcarr: Is it worth explicitly demonstrating that improperly copying does _not_ work with multiprocessing?

starfish/imagestack/_mp_dataarray.py
```diff
@@ -39,4 +46,55 @@ def __setitem__(self, key, value):
...
+    """Replacement for xr.DataArray's deepcopy method.  Returns a deep copy of the input xarray
+    backed by a multiprocessing.Array buffer.
+    """
+    shaped_np_array, backing_array = np_array_backed_by_mp_array(
+        source.variable.shape, source.variable.dtype)
+
+    shaped_np_array[:] = source.variable.data
+
+    variable = Variable(
+        copy.deepcopy(source.variable.dims),
```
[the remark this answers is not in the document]
ttung: I don't, you're right!  Updated.

starfish/test/test_multiprocessing.py
ttung: If you don't use the proper copy, you don't have a `multiprocessing.Array` to pass to `multiprocessing.Pool`.  In that case, you're just passing a numpy array to `multiprocessing.Pool`, which is `test_numpy_array` (first case of `starfish/test/test_multiprocessing.py`)

PR #740, opened 2018-10-25
Rename split_by to group_by.

ttung:
Split by is ambiguous because it could be axes along which we are splitting the image, or it could be the dimensions that remain.  Right now, it means the latter.

Group by, however, is less ambiguous.

Updated all the callers in the process.

Depends on https://github.com/spacetx/starfish/pull/736

Test plan: make -j fast run_notebooks

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -63,14 +63,18 @@ class ImageStack:
         retrieve a slice of the image tensor
     set_slice(indices, data, axes=None)
         set a slice of the image tensor
-    apply(func, split_by={Indices.X, Indices.Y}, in_place=False, verbose=False, n_processes=None)
-        split the image tensor into chunks and apply a function across each of the chunks to yield
-        an image tensor
-    transform(func, split_by={Indices.X, Indices.Y}, verbose=False, n_processes=None)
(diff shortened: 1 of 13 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -62,14 +62,17 @@ class ImageStack:
...
-    transform(func, split_by={Indices.X, Indices.Y}, verbose=False, n_processes=None)
-        split the image tensor into chunks and apply a function across each of the chunks
+    apply(func, group_by={Indices.ROUND, Indices.CH, Indices.Z}, in_place=False, verbose=False,
+            n_processes=None)
+        split the image tensor along one or more axes and apply a function across each of the
+        components to yield an image tensor
+    transform(func, group_by={Indices.ROUND, Indices.CH, Indices.Z}, verbose=False,
+            n_processes=None)
+        split the image tensor along one or more axes and apply a function across each of the
+        components
```
```suggestion
        components. Results are returned as a List with length equal to the number of times 
        the image tensor is split.
```

ambrosejcarr: The return style is the key difference between apply and transform -- since sphinx doesn't let us state return values, I think it's important to be explicit here.

ttung: applied.

PR #741, opened 2018-10-25
Pass the entire tensor to child processes as well as the slice indices

ttung:
Currently, we are slicing the tensor in the parent process, and passing the slices to each of the child processes.  To share memory more efficiently, we want to pass the entire tensor (and the backing shared memory construct) to the child processes, and instead just pass the slice indices to the child process.  The child process will slice out the part of the tensor it is interested in, and then pass it to the callable (filter, spot calling code, etc.).

`_multiprocessing_workflow` is the entry point for the child processes.  It retrieves the tensor using the constructs provided by https://github.com/spacetx/starfish/pull/738, and slices the data.

We also add a finalizer callable, which is something defined by the caller, has access to the entire tensor, and the slice indices.  This is to be used to merge the results into the original tensor for the `apply` case.  This will be sent out in the next PR. :)

Remove `_iter_tiles` as it is no longer needed.

Depends on #736, #737, #738, #739, #740

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -6,8 +6,18 @@ from copy import deepcopy
 from functools import partial
 from itertools import product
 from typing import (
-    Any, Callable, Iterable, Iterator, List, Mapping, MutableSequence,
-    Optional, Sequence, Set, Tuple, Union
+    Any,
+    Callable,
+    Container,
+    Iterator,
+    List,
+    Mapping,
+    MutableSequence,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
 )
 
 import matplotlib.pyplot as plt
@@ -27,6 +37,7 @@ from starfish.errors import DataFormatWarning
 from starfish.experiment.builder import build_image, TileFetcher
 from starfish.experiment.builder.defaultproviders import OnesTile, tile_fetcher_factory
 from starfish.intensity_table.intensity_table import IntensityTable
+from starfish.multiprocessing import shmem
 from starfish.types import (
     Coordinates,
     Indices,
@@ -629,25 +640,6 @@ class ImageStack:
             a = zip(indices, items)
             yield {ind: val for (ind, val) in a}
 
-    def _iter_tiles(
-            self, indices: Iterable[Mapping[Indices, Union[int, slice]]]
-    ) -> Iterable[np.ndarray]:
-        """Given an iterable of indices, return a generator of numpy arrays from self
-
-        Parameters
-        ----------
-        indices, Iterable[Mapping[str, int]]
-            Iterable of indices that map a dimension (str) to a value (int)
-
-        Yields
-        ------
-        np.ndarray
-            Numpy array that corresponds to provided indices
-        """
-        for inds in indices:
-            array, axes = self.get_slice(inds)
-            yield array
-
     def apply(
             self,
             func: Callable,
@@ -711,6 +703,11 @@ class ImageStack:
             group_by: Set[Indices]=None,
             verbose=False,
             n_processes: Optional[int]=None,
+            finalizer_callable: Optional[
+                Callable[[np.ndarray,
+                          Any,
+                          Mapping[Indices, int],
+                          Tuple[Union[int, slice], ...]], Any]]=None,
             **kwargs
     ) -> List[Any]:
         """Split the image along a set of axes, and apply a function across all the components.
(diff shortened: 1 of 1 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -699,6 +696,12 @@ def transform(
         n_processes : Optional[int]
             The number of processes to use for apply. If None, uses the output of os.cpu_count()
             (default = None).
+        finalizer_callable : Callable
```
ambrosejcarr: Missing: why someone would want to use this, and examples of usage.

starfish/imagestack/imagestack.py
```diff
@@ -711,18 +714,53 @@ def transform(
...
+                          Mapping[Indices, int],
+                          Tuple[Union[int, slice], ...]], Any]],
+            indices_and_slice_list: Tuple[Mapping[Indices, int],
+                                          Tuple[Union[int, slice], ...]],
+    ):
+        backing_mp_array, shape, dtype = shmem.get_payload()
+        unshaped_numpy_array = np.frombuffer(backing_mp_array.get_obj(), dtype=dtype)
+        numpy_array = unshaped_numpy_array.reshape(shape)
+
+        sliced = numpy_array[indices_and_slice_list[1]]
```
ambrosejcarr: So, we're doing something to `sliced`, and `finalizer_callable` is enabling us to mess around with the shared numpy array (e.g., enables in-place operations?)

Wouldn't that technically also be available within worker_callable() if it did any in-place operation?

starfish/imagestack/imagestack.py
```diff
@@ -1121,3 +1159,37 @@ def un_squeeze(self, stack):
         new_shape = (self.num_rounds, self.num_chs, self.num_zlayers) + self.tile_shape
         res = stack.reshape(new_shape)
         return res
+
+
+class _StripArguments:
+    """
+    Class to strip out arguments to an argument call.  This is spiritually the opposite of
```
```suggestion
    Class to strip out arguments to a function call.  This is spiritually the opposite of
```

starfish/imagestack/imagestack.py
```diff
@@ -1121,3 +1159,37 @@ def un_squeeze(self, stack):
...
+    functools.partial.  When this proxy is called, it removes some arguments and passes it to the
```
ambrosejcarr: define "it" here -- the stripped method?

starfish/imagestack/imagestack.py
```diff
@@ -1121,3 +1159,37 @@ def un_squeeze(self, stack):
...
+    original callable.
+
+    This is used in scenarios where factory methods and lambdas cannot be used, such as dispatching
```
ambrosejcarr: I think I'm looking for examples of usage so that I can figure out why this is needed.

starfish/imagestack/imagestack.py
```diff
@@ -711,18 +714,53 @@ def transform(
...
         if verbose:
-            tiles = tqdm(self._iter_tiles(indices))
-        else:
-            tiles = self._iter_tiles(indices)
+            indices_and_slice_list = tqdm(indices_and_slice_list)
 
-        applyfunc: Callable = partial(func, **kwargs)
+        applyfunc: Callable = partial(
+            self._multiprocessing_workflow,
+            _StripArguments(partial(func, **kwargs), (1, 2)),
```
ambrosejcarr: "remove 2nd and 3rd arguments of this partial, which probably _come from kwargs_" is a pretty indecipherable piece of code. :|

starfish/imagestack/imagestack.py
ttung: `func` gets the sliced data.  `finalizer_callable` gets the original data array, the results, and the slice list.

starfish/imagestack/imagestack.py
ttung: haha, it's actually more complicated than that.  arguments 2 & 3 come from `_multiprocessing_workflow`.  kwargs are bound, so they can't be stripped.

starfish/imagestack/imagestack.py
ttung: the sliced data is available in `worker_callable`.  the entire original data set is not available.  i guess we could conceivably limit the view to the slice as far as the finalizer is concerned.  i was trying to generalize it, but I can definitely see that it may be premature generalization.

starfish/imagestack/imagestack.py
ttung: will do.

starfish/imagestack/imagestack.py
ttung: Will reword.

starfish/imagestack/imagestack.py
ttung: simply put: one cannot pass a lambda or a method returned from another method because lexical scopes cannot be pickled.

ttung:
Replaced by #754

PR #742, opened 2018-10-25
Apply to save results in-place

ttung:
Rather than have apply calculate the output, and pass it back to the parent process, we merge the results back into the original tensor.  This is possible because all the child processes use a shared memory construct provided by https://github.com/spacetx/starfish/pull/738.  This is done in the finalizer callable provided by https://github.com/spacetx/starfish/pull/754.

Depends on #736, #737, #738, #739, #740, #754 

Test plan: make -j fast run_notebooks

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -732,13 +732,21 @@ class ImageStack:
                 func,
                 group_by=group_by, in_place=True, verbose=verbose, n_processes=n_processes, **kwargs
             )
+        bound_func = partial(ImageStack._in_place_apply, func)
 
-        results = self.transform(func, group_by=group_by, verbose=verbose,
-                                 n_processes=n_processes, **kwargs)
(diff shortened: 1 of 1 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -1193,3 +1196,12 @@ def __call__(self, *args, **kwargs):
         for keyword_argument_removed in self.keyword_arguments_removed:
             kwargs.remove(keyword_argument_removed)
         return self.func(*called_args, **kwargs)
+
+
+def apply_finalizer_callable(
```
ambrosejcarr: This is only ever used in our code to write to an array, right? I suspect we can be more specific in our naming to make clearer what's going on here. 

```suggestion
def write_result_to_shmem_array(
```

ttung: This is no longer necessary, given the changes I made to #754

PR #757, opened 2018-10-31
Methods to build iterators across a discrete n-dimensional space

ttung:
Given a discrete n-dimensional space, return an iterator that varies across that space.  The user picks the ordering of how dimensions vary, from slowest varying to fastest varying.

Test plan: see unit tests

PR #758, opened 2018-10-31
Dimension iteration order can be configured in the experiment builder.

ttung:
We can vary which dimensions vary the fastest.

Depends on #757 

Test plan: `python examples/get_iss_data.py /tmp/starfish/raw /tmp/starfish/formatted --d 1`

### DIFF
--- a/starfish/experiment/builder/__init__.py
+++ b/starfish/experiment/builder/__init__.py
@@ -1,6 +1,6 @@
 import json
 import os
-from typing import Any, BinaryIO, Callable, Dict, Mapping, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


ttung:
> I think itertools.product simplifies a lot of what is implemented in this and # 757.

I considered `itertools.product` but decided against it since it produces unlabeled output.  If you feel strongly about it, I can rebuild this using `itertools.product`.

ttung:
> I think itertools.product simplifies a lot of what is implemented in this and # 757.

In case you missed it, I updated this to use `itertools.product` and still have named data.

Issue #449, opened 2018-08-18
ImageStack methods should generate new ImageStacks, not numpy arrays


ambrosejcarr:
Right now, many methods (max_proj, get_slice) extract numpy arrays or xarrays from ImageStack. This is problematic because they're no longer fully-featured image objects. You can't write them, for example. 

I think longer term it would be nice for objects to at least have an option of generating a new ImageStack. Since ImageStacks all expose the underlying array, this feels like an increase in flexibility to me, and would simplify get_slice because the axes labels would remain attached to the object (no need to emit a tuple). The main hiccup would be how to deal with projected axes (e.g. does a max proj of all categorical values produce a (1, 1, 1, y, x) ImageStack?

I like the way that xarray deals with this, which would be (1, 1, 1, y, x) with an option to squeeze out the extra axes (y, x). In the latter case, maybe we emit a non-ImageStack object. 

@ttung @dganguli @shanaxel42  Interested in your thoughts.

ttung:
@shanaxel42 has done this for `max_proj`.  Does it make sense to do this for `get_slice`?  attn @ambrosejcarr  @dganguli

ttung:
If not, we can close this.

I suspect that what you really want to use is the crop methods when you want to produce another ImageStack, which is already present or in PR.

PR #765, opened 2018-11-02
Max Proj Returns ImageStack

shanaxel42:
Max proj refactored to return an Imagestack

A private _squeezed_numpy() method added to Imagestack that returns the ImageStack's data as a squeezed numpy array (what max_proj used to return)

Code that used to use max_proj updated to use new max_proj and _squeezed_numpy() when needed. 

If there are places where we think we should refactor stuff to use ImageStacks instead of using the _squeezed_numpy() method let me know. 

Test Plan:  make -j run_notebooks

fixes: #449


notebooks/py/MERFISH_Pipeline_-_U2O2_Cell_Culture_-_1_FOV.py
```diff
@@ -245,6 +245,7 @@
     ax1.axes.set_axis_off()
 
     mp = scaled_image.max_proj(Indices.ROUND, Indices.CH, Indices.Z)
```
ttung: the pattern you took with some of the other files were:

xx_max = xx.max_proj(..)
xx = xx_max._squeezed_numpy(..)

do you want to keep with that pattern here? or alternatively, do what you did here for the other cases?

starfish/image/_registration/fourier_shift.py
```diff
@@ -63,12 +63,14 @@ def run(self, image: ImageStack, in_place: bool=False) -> Optional[ImageStack]:
 
         # TODO: (ambrosejcarr) is this the appropriate way of dealing with Z in registration?
         mp = image.max_proj(Indices.CH, Indices.Z)
+        mp_numpy = mp._squeezed_numpy(Indices.CH, Indices.Z)
```
ttung: this is yet another model.  i kind of like this one, actually.

starfish/imagestack/imagestack.py
```diff
@@ -992,8 +992,16 @@ def max_proj(self, *dims: Indices) -> np.ndarray:
...
-        return max_projection
+        max_projection = self._data.max([dim.value for dim in dims])
+        max_projection = max_projection.expand_dims(tuple(dim.value for dim in dims))
+        max_projection = max_projection.transpose(*self.xarray.dims)
+        max_proj_stack = self.from_numpy_array(max_projection.values)
+        return max_proj_stack
+
+    def _squeezed_numpy(self, *dims: Indices):
+        """return this ImageStack's data as a squeezed numpy array"""
+        return xr.DataArray.squeeze(self.xarray,
```
ttung: I would just do: `self.xarray.squeeze().values`

it's not clear `_squeezed_numpy` needs to limit dims (is there a case where we need it?)

shanaxel42: If there's an Imagestack that has a dimension of size 1 other then the ones you max project over calling squeeze on the result will accidentally also remove that dimension (because it was already size 1). This came up in the ISS notebook test I think, at one point the imagestack had a shape like (r:1, c:4, z:4, x:40, y:50) but we only did a max projection over ch and z so the resulting stack was (r:1, c:1, z:1, x:40, y:50) and calling squeeze on that incorrectly removed the round dimension. So I added the dims params so you could limit the dimensions to squeeze

ttung: Ah! That makes sense.

I think you can still do `self.xarray.squeeze((dims...)).values` though.

ttung: ok.  but why aren't you just calling `self.xarray.squeeze` instead of `xr.DataArray.squeeze(xarray)`

notebooks/py/assay_comparison.py
```diff
@@ -87,25 +88,29 @@ def curl(dest_path, link):
...
-merfish_nuclei = experiment.fov()['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)
-
+merfish_nuclei_mp = experiment.fov()['nuclei'].max_proj(Indices.CH, Indices.ROUND, Indices.Z)._squeezed_numpy()
+merfish_nuclei__mp_numpy = merfish_nuclei_mp._squeezed_numpy(Indices.CH, Indices.ROUND, Indices.Z)
 # merfish doesn't have a dots image, and some of the channels are stronger than others.
 # We can use the scale factors to get the right levels
 merfish_background = experiment.fov()[FieldOfView.PRIMARY_IMAGES].max_proj(Indices.CH, Indices.ROUND)
-merfish_background = np.reshape(merfish_background, (1, 1, *merfish_background.shape))
-merfish_background = ImageStack.from_numpy_array(merfish_background)
+# merfish_background_numpy = merfish_background._squeezed_numpy(Indices.CH, Indices.ROUND)
```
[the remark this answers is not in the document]
shanaxel42: oh! I realized the commented out lines were basically doing everything max_proj now does to get an imagestack so we don't need them anymore. I'll delete them and verify all the notebooks run though

PR #772, opened 2018-11-06
[easy] stylistic cleanup of physical_coordinate_calculator.py

ttung:
1. `physcial` -> `physical`.
2. Reorder min before max for `_calculate_physical_pixel_size`.
3. Small line wrapping tweaks.
4. Fix annotations and docblock for `_recalculate_physical_coordinate_ranges`.

### DIFF
--- a/starfish/imagestack/physical_coordinate_calculator.py
+++ b/starfish/imagestack/physical_coordinate_calculator.py
@@ -1,4 +1,4 @@
-from typing import Any, Mapping, Tuple, Union
+from typing import Any, Mapping, Optional, Tuple, Union
 
 import xarray as xr
 
@@ -12,9 +12,11 @@ from starfish.types import (
 )
 
 
-def calc_new_physical_coords_array(physical_coordinates: xr.DataArray,
-                                   stack_shape: Mapping[Indices, int],
-                                   indexers: Mapping[str, Union[int, slice]]) -> xr.DataArray:
+def calc_new_physical_coords_array(
+        physical_coordinates: xr.DataArray,
+        stack_shape: Mapping[Indices, int],
+        indexers: Mapping[str, Union[int, slice]],
+) -> xr.DataArray:
     """Calculates the resulting coordinates array from indexing on each dimension in indexers
 
     Parameters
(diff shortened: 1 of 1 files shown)


starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -121,11 +123,12 @@ def _pixel_offset_to_physical_coordinate(physical_pixel_size: Number,
...
-                                           coord_max: float,
-                                           dimension_size: int,
-                                           indexer
-                                           ) -> Tuple[Number, Number]:
+def _recalculate_physical_coordinate_range(
+        coord_min: float,
+        coord_max: float,
+        dimension_size: int,
+        indexer
+) -> Tuple[Number, Number]:
```
shanaxel42: So if there's listed params like this I should always start them on the next line?

starfish/imagestack/physical_coordinate_calculator.py
```diff
@@ -42,8 +44,8 @@ def calc_new_physical_coords_array(physical_coordinates: xr.DataArray,
     return new_coords
 
 
-def _needs_coords_recalculating(x_indexers: Union[int, slice], y_indexers: Union[int, slice]
-                                ) -> bool:
+def _needs_coords_recalculating(
+        x_indexers: Union[int, slice], y_indexers: Union[int, slice]) -> bool:
     if isinstance(x_indexers, int) or isinstance(y_indexers, int):
```
shanaxel42: If the return definition goes over the line limit you'd prefer to just have everything on the next line?

ttung: Yup.  The rationale is the same: it minimizes LOC churn if we rename methods.

starfish/imagestack/physical_coordinate_calculator.py
ttung: So my preference is that for multiline params (calls and method declarations), that we start on the next line (basically the second option of https://www.python.org/dev/peps/pep-0008/#indentation).

Rationale: If the method name changes, we do not need to change the indentation for every line.

PR #773, opened 2018-11-06
Adding tiles aligned attribute to Imagestack

shanaxel42:
This is primarily for #682 but I think may be useful info to have elsewhere.


starfish/test/image/test_imagestack_coordinates.py
```diff
@@ -153,3 +157,44 @@ def test_scalar_coordinates():
...
+
+    @property
+    def format(self) -> ImageFormat:
+        return ImageFormat.TIFF
+
+    def tile_data(self) -> np.ndarray:
+        return np.ones((HEIGHT, WIDTH), dtype=np.float32)
+
+
+def test_aligend_coordinates():
```
dganguli: maybe add a test that asserts that tiles are not aligned?

shanaxel42:
Going to just combine this with the error PR since the changes are small

Issue #760, opened 2018-11-01
All filters should accept boolean is_volume


joshmoore:
see https://github.com/spacetx/starfish/pull/748#discussion_r229096569

gh-748 marked a few locations where is_volume was not being passed.

PR #775, opened 2018-11-07
adding is_volume flag to filters

shanaxel42:
Fixes: #760


starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -6,15 +6,23 @@
...
+    gaussian_kernel,
+    preserve_float_range
+)
 
 
 class DeconvolvePSF(FilterAlgorithmBase):
 
     def __init__(
-            self, num_iter: int, sigma: Number, clip: bool=True) -> None:
+            self,
```
ttung: btw, if this fits on one line, you can go ahead and make it all on one line.  my general order of preferences:
1. fit on one line.
2. fit on two lines.
3. one param per line.

starfish/image/_filter/zero_by_channel_magnitude.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, thresh: int, normalize: bool) -> None:
         """
         self.thresh = thresh
         self.normalize = normalize
+        self.is_volume = is_volume
```
ttung: doesn't look like anyone uses this...?

starfish/image/_filter/richardson_lucy_deconvolution.py
```diff
@@ -27,6 +35,7 @@ def __init__(
         clip : bool (default = False)
             if True, pixel values below -1 and above 1 are clipped for skimage pipeline
             compatibility
+            :param is_volume:
```
ttung: this is not numpy doc style.

shanaxel42: oh man I think pycharm auto did this

starfish/image/_filter/zero_by_channel_magnitude.py
joshmoore: @shanaxel42 : I wasn't sure either which is why I passed the buck. 😉

ttung: My intuition is that we should not include this as it implies it does something it does not.  I think there are filters that do not accept this format.

Second choice: raise ValueError if is_volume is True.

shanaxel42: Updated to not include the flag here

PR #777, opened 2018-11-08
[Easy] Add default threshold parameter for Bandpass filter

dganguli:
The Bandpass filter wraps trackpy's bandpass filter. This underlying filter accepts a None value for the threshold parameter. This None value then internally gets set to 1/255. This makes sense for uint8 images, but not for our float32 images which contain small values in [0,1]. This caused some pain for @berl who is used to using trackpy in this manner. 

To counteract this issue, I set the default for the threshold to 0 (which is effectively a noop) and raise an exception if the user passes in a None value.

### DIFF
--- a/starfish/image/_filter/bandpass.py
+++ b/starfish/image/_filter/bandpass.py
@@ -14,7 +14,7 @@ from .util import determine_axes_to_group_by, preserve_float_range
 class Bandpass(FilterAlgorithmBase):
 
     def __init__(
-            self, lshort: Number, llong: int, threshold: Number, truncate: Number=4,
+            self, lshort: Number, llong: int, threshold: Number=0, truncate: Number=4,
             is_volume: bool=False) -> None:
         """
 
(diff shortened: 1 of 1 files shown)


dganguli:
nice catches! pushed fixes.

starfish/image/_filter/bandpass.py
```diff
@@ -25,15 +25,19 @@ def __init__(
...
             truncate the gaussian kernel, used by the gaussian filter, at this many standard
-            deviations
+            deviations (default 4)
         is_volume : bool
             If True, 3d (z, y, x) volumes will be filtered. By default, filter 2-d (y, x) planes
         """
         self.lshort = lshort
         self.llong = llong
+
+        if threshold is None:
```
ttung: necro-comment: this wasn't in the original PR.  curious why this is necessary.

starfish/image/_filter/bandpass.py
```diff
@@ -25,15 +25,19 @@ def __init__(
         llong : int
             filter frequencies above this odd integer value
         threshold : float
```
ttung: didn't notice this in the first round, but this (and `truncate`) are declared as `floats` in the doc, but as `Number` in the params.

PR #780, opened 2018-11-09
Fix E605 warnings.

ttung:
1. Put back E605 as a valid warning.
2. Use r-style strings to not quote backslashes.

Test plan: `make -j lint`, waiting on travis now...


ttung:
this is so ez i'm going to land this.

PR #787, opened 2018-11-14
Refactoring GaussianSpotDetector

### LINKED ISSUE #727: add support for multiple skimage blob detectors
Currently only `blog_log` from `skimage.feature` is supported through the `GaussianSpotDetector`. `blob_dog` and `blob_doh` also from `skimage.feature` have the same api, but different tradeoffs around speed and what they detect. It would be nice to expose all three blob detectors to starfish users. One way to do this would be to rename `GaussianSpotDetector` something like `BlobDetector` and include an additional argument, `algorithm` which has to be one of `blog_log`, `blob_dog`, or `blob_doh` and then call the appropriate function inside the detector.

shanaxel42:
Modified the signature to take in one of 3 different skimage blob detection options

Changed name from GaussianSpotDetector to BlobDetector

fixes #727

### DIFF
--- a/docs/source/usage/iss/iss_pipeline.py
+++ b/docs/source/usage/iss/iss_pipeline.py
@@ -24,7 +24,7 @@ def iss_pipeline(fov, codebook):
     filtered = filt.run(registered, verbose=True, in_place=False)
 
     # detect spots using laplacian of gaussians approach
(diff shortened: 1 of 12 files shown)


joshmoore:
Note: rename propagated to docs in https://github.com/spacetx/starfish/pull/886

PR #790, opened 2018-11-15
[NO-TILESET 2/3] Implement ImageStack write without depending on the source ImageStack

ttung:
Rather than writing the data back to the source ImageStack and writing that out to disk, make a new TileSet, and set it up appropriately.

This makes explicit what is needed for writing the ImageStack, and what is used to write the ImageStack.

Depends on #782 

Test plan: `make -j fast run_notebooks`

### DIFF
--- a/examples/get_cli_test_data.py
+++ b/examples/get_cli_test_data.py
@@ -21,11 +21,11 @@ if __name__ == "__main__":
     for fov in exp.fovs():
         fov_dir = pathlib.Path(args.output_dir, fov.name)
         fov_dir.mkdir()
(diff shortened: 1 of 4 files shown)


ttung:
Prior to this PR, we would write the image data back to the original TileSet that the ImageStack came from, and then write the TileSet to disk.  That ties the implementation of ImageStack to TileSet, and creates some awkwardness in the code.  For instance, in `from_numpy_array`, we have to construct a TileSet, copy the data to the TileSet, and then create an ImageStack from that.

The new approach simply states we want to write a TileSet.  The code creates a TileSet, copies the image data to it, and writes it to disk.

ttung:
I'm going to rename `write` to `export_to_tileset` to be 100% explicit about what the method does.

PR #795, opened 2018-11-16
[easy] order fovs by name by default

joshmoore:
Order all fovs by name by default. (Other idea from starfish sync was to order by physical coordinates)

cc: @berl

### DIFF
--- a/starfish/experiment/experiment.py
+++ b/starfish/experiment/experiment.py
@@ -269,8 +269,8 @@ class Experiment:
     ) -> FieldOfView:
         """
         Given a callable filter_fn, apply it to all the FOVs in this experiment.  Return the first
-        FOV such that filter_fn(FOV) returns True.  Because there is no guaranteed order for the
-        FOVs, use this cautiously.
+        FOV such that filter_fn(FOV) returns True. The order of the filtered FOVs will be determined
+        by the key_fn callable. By default, this matches the order of fov.name.
 
         If no FOV matches, raise LookupError.
         """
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @berl. I've added docs & tests now. Landing.

starfish/experiment/experiment.py
```diff
@@ -265,39 +265,48 @@ def verify_version(cls, semantic_version_str: str) -> Version:
     def fov(
             self,
             filter_fn: Callable[[FieldOfView], bool]=lambda _: True,
+            key_fn: Callable[[FieldOfView], str]=lambda fov: fov.name,
```
ttung: Is there a compelling reason for any alternative key?  I'm ok with making the fovs always sorted by name.

In the "distant" past, fovs were not guaranteed to have names, but as they are now, it might just make sense to always sort by name.

joshmoore: > Is there a compelling reason for any alternative key? I'm ok with making the fovs always sorted by name.

The only other sorting that occurred to me was by physical coordinates in the case that the names are not helpful. `fov.name` was certainly easier in terms of a quick, stable sorting.

ttung: Yeah, physical coordinates seemed like the only other plausible option, but that still seems unlikely since sorting a 3D space (R,C,Z) into 1D is just ... weird.

joshmoore: @ttung: would you prefer the argument to be reverted?

ttung: No, it does seem weird to me, but I'm okay leaving it in there. :)

PR #796, opened 2018-11-16
click logging of pipelines

joshmoore:
Generate a record of all top-level build steps (in click land this is currently equivalent to those methods which are wrapped with `pass_context`).

Also:
 * Add --quiet flag to block art (config: `cli.quiet`)
 * Add --record flag for a file path or `-` (config: `cli.record`)
 * Record all cli steps in the ctx.obj["record"] array
 * Print the records to the given path on exit

Current usage with output:

```
$ python -m starfish build --record=- --fov-count=1 --hybridization-dimensions='{"z":1}' out2
[{"build": {"params": {"fov_count": 1, "hybridization_dimensions": {"r": 1, "c": 1, "z": 1}, "output_dir": "out2", "nuclei_dimensions": null, "dots_dimensions": null}}}]
```

A solid next refactoring step might be to build a general `Recorder` object that accessible by other methods for further recording. (This may or may not be best handled by re-using `logging`)
 
cc: @berl @dganguli @shanaxel42

### DIFF
--- a/sptx_format/cli.py
+++ b/sptx_format/cli.py
@@ -1,12 +1,13 @@
 import click
 
 from sptx_format import validate_sptx
+from starfish.util.click import pass_context_and_record
 
 
 @click.command()
 @click.option("--experiment-json", required=True, metavar="JSON_FILE_OR_URL")
 @click.option("--fuzz", is_flag=True)
-@click.pass_context
+@pass_context_and_record
 def validate(ctx, experiment_json, fuzz):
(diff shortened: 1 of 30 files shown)


ambrosejcarr:
@joshmoore is this still active? Do you need anything from me here?

joshmoore:
I really defer to what everyone else would like to see. Closing for the moment, but happy to re-open on top of @shanaxel42's provenance work.

PR #797, opened 2018-11-17
Adding extra state info LocalMaxPeak.run()

### LINKED ISSUE #708: SpotFinders do not have state. They should be able to return results for debug/QC
For any spot detection algorithm that implements ```SpotFinderAlgorithmBase```, it's not possible to add private method variables to maintain state. This is because of the way the ```run``` method works, which calls ```detect_spots``` which calls ```Stack.apply``` which calls ```SpotFinderAlgorithmBase.image_to_spots``` on each tile of the stack. The latter method is the contains the actual spot finding code. Everything else is just glue.

If the objects ```image_to_spots``` method updates state, this state is not actually kept.  ```Stack.apply``` works by spawning multiple worker processes. But setting variables in the worker processes does not propagate back to the main process. As such, spot detectors cannot maintain state.

I propose, we change the method signature for the argument in detect.detect_spots (spot_finding_method)

from:
```  spot_finding_method: Callable[..., SpotAttributes]```
to:
```Callable[..., Tuple[SpotAttributes, Optional[SpotDetectorResults]]]```

this will _not_ fix the ability of spot detectors to maintain state, but _will_ allow them to return useful quantities for diagnosing issues with and performing quality control for spot detectors.

### PR DESCRIPTION
**Pull request by shanaxel42**
The goal of this PR is to provide a way for people to access the state information of LocalMaxPeakFinder for debugging after spot finding. The approach is to add a new felid to SpotAttributes called 'extras' where any extra information about spot finding method can be added. Then the info can be accessed by calling SpotAttibutes.extras after an image_to_spots call. 

In this PR:
*New class called spot_detector_restults where NamedTuple objects can be added for any spot detection method
*Added an extras field to SpotAttributes 
*Edited local_max image_to_spots to include extra state info on the SpotAttributes object
*Added a test that checks to make sure the info persists through multiprocessing 

fixes #708

### DIFF
--- a/starfish/spots/_detector/local_max_peak_finder.py
+++ b/starfish/spots/_detector/local_max_peak_finder.py
@@ -261,17 +261,25 @@ class LocalMaxPeakFinder(SpotFinderAlgorithmBase):
                Features.SPOT_ID: np.arange(self._spot_coords.shape[0]),
                Features.INTENSITY: data_image[self._spot_coords[:, 0],
                                               self._spot_coords[:, 1],
-                                              self._spot_coords[:, 2]]
-               }
+                                              self._spot_coords[:, 2]],
 
-        return SpotAttributes(pd.DataFrame(res))
+               }
+        lmr = LocalMaxFinderResults(
+            self._thresholds,
+            self._spot_counts,
+            self._grad,
+            self._spot_props,
+            self._labels
+        )
+
+        return SpotAttributes(pd.DataFrame(res), lmr)
 
     def run(
             self,
             data_stack: ImageStack,
             blobs_image: Optional[Union[np.ndarray, xr.DataArray]] = None,
             reference_image_from_max_projection: bool = False,
-    ) -> Tuple[IntensityTable, LocalMaxFinderResults]:
+    ) -> IntensityTable:
         """Find spots.
 
         Parameters
@@ -299,15 +307,7 @@ class LocalMaxPeakFinder(SpotFinderAlgorithmBase):
             measurement_function=self.measurement_function,
             radius_is_gyration=False)
 
-        lmr = LocalMaxFinderResults(
-            self._thresholds,
-            self._spot_counts,
-            self._grad,
-            self._spot_props,
-            self._labels
-        )
-
-        return intensity_table, lmr
+        return intensity_table
 
     @staticmethod
     @click.command("LocalMaxPeakFinder")
--- a/starfish/spots/_detector/local_max_peak_finder.py
+++ b/starfish/spots/_detector/local_max_peak_finder.py
@@ -15,12 +15,14 @@ from starfish.intensity_table.intensity_table import IntensityTable
 from starfish.types import Features, Indices, Number, SpotAttributes
 from ._base import SpotFinderAlgorithmBase
 from .detect import detect_spots
+from .spot_detector_results import LocalMaxFinderResults
 
 
 # TODO ambrosejcarr, ttung: https://github.com/spacetx/starfish/issues/708
 # Currently, spot finders cannot propagate state, which makes the flow for this
 # spot finder confusing. One would expect to have access to the private parameters
 # however, they are lost due to the memory-space forking induced by multi-processing.
+
 
 class LocalMaxPeakFinder(SpotFinderAlgorithmBase):
     def __init__(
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: starfish/spots/_detector/local_max_peak_finder.py [hunk 1]
```diff
@@ -297,7 +299,15 @@ def run(
             measurement_function=self.measurement_function,
             radius_is_gyration=False)
 
-        return intensity_table
+        lmr = LocalMaxFinderResults(
+            self._thresholds,
```
dganguli: my understanding is that the self.* variables never get set in the first place when running this w/ multiple processes? i think what we need to verify that this works with a unit test. the test plan is to ensure that the returned lmr indeed has values for the desired variables. 

only thing need to make this work is:
1. a realistic synthetic dataset example
2. a code path that goes through this part of detect_spots:

```
    else:  # don't use a reference image, measure each
        spot_finding_method = partial(spot_finding_method, **spot_finding_kwargs)
        spot_attributes_list = data_stack.transform(
            func=spot_finding_method,
            group_by=group_by
        )
        intensity_table = concatenate_spot_attributes_to_intensities(spot_attributes_list)
```

dganguli: the offending piece of the above code is data_stack.transform, which does this:

```        with multiprocessing.Pool(
                n_processes,
                initializer=SharedMemory.initializer,
                initargs=((self._data._backing_mp_array,
                           self._data._data.shape,
                           self._data._data.dtype),)) as pool:
            results = pool.imap(applyfunc, indices_and_slice_list)
            return list(zip(results, indices))```

Comment (shanaxel42) [reply starfish/spots/_detector/local_max_peak_finder.py]: Ahh this changes things. Wait why does the test need to go through this section of detect spots? `    else:  # don't use a reference image, measure each
        spot_finding_method = partial(spot_finding_method, **spot_finding_kwargs)
        spot_attributes_list = data_stack.transform(
            func=spot_finding_method,
            group_by=group_by
        )
        intensity_table = concatenate_spot_attributes_to_intensities(spot_attributes_list)` I thought that if someone wanted the extra info they would call LocalMaxPeakFinder.run() directly like they do for PIxelSpotDetector in the dartfish notebook? `spot_intensities, results = psd.run(zero_norm_stack)`

File: starfish/types/_spot_attributes.py [hunk 2]
```diff
@@ -16,15 +16,17 @@ class SpotAttributes(ValidatedTable):
...
     }
 
-    def __init__(self, spot_attributes: pd.DataFrame) -> None:
+    def __init__(self, spot_attributes: pd.DataFrame, extras=None) -> None:
         """Construct a SpotAttributes instance
 
         Parameters
         ----------
         spot_attributes : pd.DataFrame
+        :param extras:
```
ambrosejcarr: Please document + fix your pycharm auto-settings :-)

Ref: https://www.jetbrains.com/help/pycharm/settings-tools-python-integrated-tools.html

starfish/spots/_detector/spot_detector_results.py
```diff
@@ -0,0 +1,11 @@
+from typing import List, NamedTuple, Optional, Union
```
ambrosejcarr: I think it would be appropriate to refactor the `ConnectedComponentDecodingResult` from `combine_adjacent_features.py` to also live in this file.

starfish/test/spots/detector/test_spot_detection.py
```diff
@@ -170,3 +172,22 @@ def test_spot_finding_no_reference_image(
     expected = [0.00793712] * 4
     assert np.allclose(intensity_table.sum((Indices.ROUND, Indices.CH)).values, expected), \
         "wrong spot intensities detected"
+
+
+@pytest.mark.parametrize('data_stack, spot_detector, radius_is_gyration', [
+    (data_stack, local_max_spot_detector, False)
+])
+def test_spot_finding_maintains_state_info(
```
ambrosejcarr: I would be more comfortable if this also tested the spot finder `run` method.

ambrosejcarr:
@shanaxel42 is this ready for re-review? This has been here a while. :-)

shanaxel42:
> @shanaxel42 is this ready for re-review? I don't know if Deep is going to get to this for a while ;-)

Ha yea, I actually think it'll be obsolete when I merge the multiproceses n=1 changes because I believe this was all for debugging purposes anyway. I'll delete this PR once the other one's up!

ambrosejcarr:
>  I believe this was all for debugging purposes anyway.

I think that's not quite right -- there are some additional data that are returned by this approach, akin to the `ConnectedComponentDecodingResult` returned by the `PixelSpotDecoder`. We want those results regardless of whether they are calculated on the main process or in subprocesses, and right now I believe the data are lost in the subprocesses because the `SpotFinder` objects can't pass state back to the main thread. 

There's potentially a question of algorithm design here (e.g. do we avoid statefulness and reprogram the method using a functional approach?)

PR #802, opened 2018-11-23
Extract the TileSet → ImageStack code into a separate module.

ttung:
1. `_get_dimension_size` gets moved to the new module as it is no longer needed in ImageStack.
2. Rename `tile_metadata` to `tile_data` as it represents both the image data and the metadata now.

Depends on #791

This is part of RFC #776

Test plan: make -j run_notebooks

### DIFF
--- a/starfish/imagestack/parser/tileset/_parser.py
+++ b/starfish/imagestack/parser/tileset/_parser.py
@@ -1,40 +1,114 @@
 """
 This module parses and retains the extras metadata attached to TileSet extras.
 """
+import warnings
+from typing import Collection, MutableMapping, Tuple
 
-from typing import Collection, Mapping, MutableMapping
+from slicedimage import Tile, TileSet
 
-from slicedimage import TileSet
-
+from starfish.errors import DataFormatWarning
+from starfish.imagestack.dataorder import AXES_DATA
 from starfish.imagestack.parser import TileCollectionData, TileKey
 from starfish.types import Indices
 
 
 class TileSetData(TileCollectionData):
     """
-    This class parses the data, including extras, from a TileSet and its constituent tiles.
+    This class presents a simpler API for accessing a TileSet and its constituent tiles.
     """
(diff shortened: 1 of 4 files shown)


starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -1,40 +1,114 @@
...
+    Tuple[Tuple[int, int], TileSetData] :
+        A tuple consisting of the following:
+            1. The (y, x) size of each tile.
+            2. A :py:class:`starfish.imagestack.tileset.TileSetData` that can be queried to obtain
+               the image data and extras metadata of each tile, as well as the extras metadata of
+               the entire :py:class:`slicedimage.TileSet`.
+    """
+    tile_data = TileSetData(tileset)
+
+    # NOTE: (ttung) this is highly inefficient as it forces us to decode the tiles multiple times.
```
ambrosejcarr: I'll keep an eye out. :-)

joshmoore:
Not directly related to this PR, but as a general heads-up while working on related factorings: is there a way to access the source `path_or_url` for a given tile from within the imagestack? This is related to gh-603 and particularly in trying to have a version of gh-893 which doesn't re-write the data.

While working on the validation propagation (via StarfishConfig), I wondered if that would be made easier by refactoring the top-level logic like that in `Experiment.from_json` into each individual constructor/builder method.

ttung:
> Not directly related to this PR, but as a general heads-up while working on related factorings: is there a way to access the source path_or_url for a given tile from within the imagestack? This is related to gh-603 and particularly in trying to have a version of gh-893 which doesn't re-write the data.

I think this is somewhat complicated because we could in fact have an ImageStack constructed entirely in-memory.

You could grab the backing `TileCollectionData`, and if it's an instance of `TileSetData`, you can grab the TileSet from it, and then you can grab the individual Tile objects, and inspect the `_numpy_array_future` field.

It breaks a zillion abstraction boundaries though. :(

PR #803, opened 2018-11-23
Dynamically evaluate the tile shape and dtype as we decode.

ttung:
There is an inefficiency in the code that has us decoding the image _twice_, once to find the common dtype of the data and once to constitute the 5D tensor.

#802 exacerbates the inefficiency even more, by calling numpy_array multiple times, each time triggering a decode.

This PR changes the code so that we just decode the image once and storing it in a proxy object that looks a lot like a slicedimage Tile object.  Unlike a slicedimage Tile object, however, it retains the image for the lifetime of the object.  We tightly constrain the lifetime of these objects in the ImageStack constructor to avoid bloating memory requirements.

This PR also introduces an `_Expectations` class, which tracks the tile shapes and dtypes of the tiles seen during a decode of a TileSet.  If the shapes are not all identical, it will trigger a ValueError.  If the kind of dtypes are not all identical, it will trigger a TypeError.  Additionally, if the dtypes are all of the same kind, but not all of the same size, it will trigger a :py:class:`starfish.errors.DataFormatWarning`.

Depends on #802

This is part of RFC #776

Test plan: `make -j run_notebooks`

### DIFF
--- a/starfish/imagestack/parser/tileset/_parser.py
+++ b/starfish/imagestack/parser/tileset/_parser.py
@@ -2,14 +2,127 @@
 This module parses and retains the extras metadata attached to TileSet extras.
 """
 import warnings
-from typing import Collection, MutableMapping, Tuple
+from typing import Collection, Mapping, MutableMapping, Optional, Tuple
 
+import numpy as np
 from slicedimage import Tile, TileSet
 
 from starfish.errors import DataFormatWarning
 from starfish.imagestack.dataorder import AXES_DATA
-from starfish.imagestack.parser import TileCollectionData, TileKey
-from starfish.types import Indices
+from starfish.imagestack.parser import TileCollectionData, TileData, TileKey
+from starfish.types import Coordinates, Indices, Number
+
+
+class _TileSetConsistencyDetector:
+    """
+    This class tracks all the tile shapes and the dtypes seen thus far during the decode of a
+    :py:class:`slicedimage.TileSet`.  If the shapes are not all identical, it will trigger a
+    ValueError.  If the kind of dtypes are not all identical, it will trigger a TypeError.
+    Additionally, if the dtypes are all of the same kind, but not all of the same size, it will
+    trigger a :py:class:`starfish.errors.DataFormatWarning`.
+    """
+    def __init__(self) -> None:
+        self.tile_shape: Optional[Tuple[int, ...]] = None
+        self.kind = None
+        self.dtype_size = None
+
+    def report_tile_shape(self, r: int, ch: int, z: int, tile_shape: Tuple[int, ...]) -> None:
+        """As each tile is parsed, the tile shape should be reported via this method.  If an
+        inconsistency is detected, a ValueError exception will be raised.
+
+        Parameters
+        ----------
+        r, ch, z : int
+            The indices of the tile, which is used to generate the exception text.
+        tile_shape : Tuple[int, ...]
+            The shape of the tile.
+        """
+        if self.tile_shape is not None and self.tile_shape != tile_shape:
+            raise ValueError(
+                f"Tile (R: {r} C: {ch} Z: {z}) has shape {tile_shape}, which is different from one "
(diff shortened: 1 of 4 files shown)


starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -2,14 +2,101 @@
...
+    ValueError.  If the kind of dtypes are not all identical, it will trigger a TypeError.
+    Additionally, if the dtypes are all of the same kind, but not all of the same size, it will
+    trigger a :py:class:`starfish.errors.DataFormatWarning`.
+    """
+    def __init__(self) -> None:
+        self.tile_shape: Optional[Tuple[int, ...]] = None
+        self.kind = None
+        self.dtype_size = None
+
+    def report_tile_shape(self, r: int, ch: int, z: int, tile_shape: Tuple[int, ...]) -> None:
```
ambrosejcarr: I wasn't 100% sure what this class did, I think it could use documentation of: 
1. why it's called
2. what it raises and when
3. why r, ch, and z are necessary, since they're not used.

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -2,14 +2,101 @@
...
+        if self.tile_shape is not None and self.tile_shape != tile_shape:
+            raise ValueError("Starfish does not support tiles that are not identical in shape")
+        self.tile_shape = tile_shape
+
+    def report_dtype(self, r: int, ch: int, z: int, dtype) -> None:
```
ambrosejcarr: similarly, I'd move some of the documentation from the class docs about raising the warning down to this command. Here it's a bit clearer why r, ch, and z are needed, but it could be good to doc them as: 

```
r, ch, z : int
    Incorporated in the triggered warning to identify which Tile has a different shape
```

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -2,14 +2,101 @@
...
+                f"more of the toher tiles.",
+                DataFormatWarning)
+
+        self.kind = dtype.kind
+        self.dtype_size = dtype.itemsize
+
+
+class SlicedImageTile(TileData):
+    """
+    This wraps a :py:class:`slicedimage.Tile`.  The key behavioral difference between this and
```
ambrosejcarr: This is the key difference -- what are the other differences? Might be good to spell them out. I might also transfer some of the logic from the PR description here -- we need this to efficiently load data, etc etc.

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -2,14 +2,101 @@
...
+    :py:class:`slicedimage.Tile` is that this class does cache the image data upon load.  It is
+    therefore incumbent on the consumers of these objects to discard them as soon as it is
+    reasonable to do so to free up memory.
+    """
+    def __init__(
```
ambrosejcarr: I read this and think "what's `inner_tile`?". I don't have a great grasp of what goes on in `slicedimage` -- do you expect users to know that library to read this code? If no, I'd doc this constructor.

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -25,6 +112,7 @@ def __init__(self, tileset: TileSet) -> None:
                 z=tile.indices.get(Indices.Z, 0))
             self.tiles[key] = tile
         self._extras = tileset.extras
+        self._expectations = _Expectations()
```
ambrosejcarr: I'd doc here why you're initializing `_Expectations()` and what the expected behavior is.

starfish/imagestack/parser/tileset/_parser.py
shanaxel42: maybe it work make more sense if it was named TileSetValidator? Or something like that. That's at least my understanding of what it does...compares incoming tiles to this info to see if the tileset is valid

ttung:
I think I hit all the issues. :)

PR #804, opened 2018-11-27
fixing value error and documentation in blob.py

shanaxel42:


starfish/spots/_detector/blob.py
```diff
@@ -178,7 +178,7 @@ def run(
     @click.option(
         "--detector_method", default='blob_log',
         help="str ['blob_dog', 'blob_doh', 'blob_log'] name of the type of "
-             "detection method used from skimage.feature"
+             "detection method used from skimage.feature. Default: blob_log"
```
ttung: `click` should print the default in the help text.  `argparse` does, so I would assume that `click` can too.

unfortunately, it seems I cannot invoke the help for `BlobDetector`: https://spacetx.slack.com/archives/CAR91U354/p1543306613255300

joshmoore: Writing up the problem&solutions under https://github.com/spacetx/starfish/issues/808

Issue #808, opened 2018-11-27
Click help for subcommands broken


joshmoore:
Some reading material:

 * https://github.com/pallets/click/issues/50 "eager required options suppress help"
 * https://github.com/pallets/click/issues/108#issuecomment-255547347 "Support for shared arguments?" (tl;dr - not gonna do it)
 * https://github.com/pallets/click/issues/295 "Can't access help of subcommands without going through parent's prompts"

Possible solutions:
 * Move top level args to the subcommands (`detect_spots -i X BlobDetector` becomes `detect_spots BlobDetector -i X`)
 * Mark top level args `required=False` and check for their value at run-time
 * Make `--help` "[eager](https://click.palletsprojects.com/en/3.x/options/#callbacks-and-eager-options)" (not yet sure how to do this)
 * Add own group subclass ([SO](https://stackoverflow.com/questions/47972638/how-can-i-define-the-order-of-click-sub-commands-in-help))
 * Add own argument subclass ([SO](https://stackoverflow.com/questions/47437472/in-python-click-how-do-i-see-help-for-subcommands-whose-parents-have-required)) -- fix in progress

ttung:
Is this done?

joshmoore:
Yup.

PR #809, opened 2018-11-27
click: implement option subclass (see #808)

joshmoore:
This workarounds the issue spelled out in gh-808. The same fix will need to be applied to each of:

```
$ git grep required=True $(git grep -l click.group)
starfish/image/_filter/__init__.py:    @click.option("-o", "--output", required=True)
starfish/image/_registration/__init__.py:    @click.option("-o", "--output", required=True)
starfish/image/_segmentation/__init__.py:    @click.option("--primary-images", required=True, type=click.Path(exists=True))
starfish/image/_segmentation/__init__.py:    @click.option("--nuclei", required=True, type=click.Path(exists=True))
starfish/image/_segmentation/__init__.py:    @click.option("-o", "--output", required=True)
starfish/spots/_decoder/__init__.py:    @click.option("-i", "--input", required=True, type=click.Path(exists=True))
starfish/spots/_decoder/__init__.py:    @click.option("-o", "--output", required=True)
starfish/spots/_decoder/__init__.py:    @click.option("--codebook", required=True, type=click.Path(exists=True))
starfish/spots/_detector/__init__.py:    @click.option("-i", "--input", required=True, cls=RequiredParentOption,
starfish/spots/_detector/__init__.py:    @click.option("-o", "--output", required=True, cls=RequiredParentOption)
starfish/spots/_target_assignment/__init__.py:    @click.option("--label-image", required=True, type=click.Path(exists=True))
starfish/spots/_target_assignment/__init__.py:    @click.option("--intensities", required=True, type=click.Path(exists=True))
```

Background reading: https://stackoverflow.com/questions/47437472/in-python-click-how-do-i-see-help-for-subcommands-whose-parents-have-required

Test Plan:
 * tests should be green
 * output of the various `--help` commands should meet expectations


joshmoore:
Note: this will slightly conflict with gh-796 since I also added `starfish.util.click` there.

joshmoore:
> Does it make sense to build a proxy click library where the option method automatically pulls this off?

`from starfish.util.click import (
    command,
    group,
    option,
)`

would be straight-forward enough. I don't see anything particularly harmful in having the specialized `--help` parsing regardless of whether or not the top-level option is required.

ttung:
> would be straight-forward enough. I don't see anything particularly harmful in having the specialized --help parsing regardless of whether or not the top-level option is required.

theoretically, you can inspect to see if the caller requests `required=True`, right?

joshmoore:
> theoretically, you can inspect to see if the caller requests required=True, right?

Sure. But thinking about it, I'm going to guess in a case where the top-level _weren't_ required, we're probably getting the wrong help. `TestHelp` will need to be updated to also check for concrete strings like `BlobDetector`

ttung:
How come?  If starfish.util.click.option gets `required=True`, it just adds the `class=RequiredParentOption` parameter and calls the actual click.option method.

joshmoore:
For the state of this PR, definitely. I just have the suspicion that we need it unconditionally.

ttung:
Ah, I see.

by the way, I don't understand the statement `TestHelp will need to be updated to also check for concrete strings like BlobDetector`.  Could you explain?

joshmoore:
Similar to the fuzz test (but more important), I think `TestHelp` should have a copy of the stdout/stderr that a user expects to see and verify that it's being produced so if a refactoring breaks user-space, we're informed.

joshmoore:
I think that's everything addressed except for adding all click functionality to `starfish.util.click` (i.e. never `import click`'ing _except_ in s.u.c

starfish/util/click.py
```diff
@@ -0,0 +1,47 @@
+import click
```
ttung: idea: could we import all of click into this module, and then replace the option symbol?

i.e.,
```
from click import *

click_option = option
def option(*args, **kwargs):
    kwargs["cls"] = RequiredParentOption
    return click_option(*args, **kwargs)
```

And then where we currently import click, we would just do:

`from starfish.util import click`

and then the rest of the code should basically be unmodified.

joshmoore: I don't see any idiom for doing `import *`; I'll import explicitly and push.

joshmoore: Pushed.

joshmoore:
Corrected conflicts, ready for review.

starfish/test/full_pipelines/cli/test_help.py
```diff
@@ -0,0 +1,120 @@
...
+  filter
+  registration
+  segment
+  target_assignment
+  validate           invokes validate with the parsed commandline...
+  version
+"""
+        assert_diff(actual, expected)
+
+    def test_second(self):
```
ambrosejcarr: These tests seem like they might be a bit annoying to refactor. Would it be adequate to code up something like the following for the purposes of these tests? 
```python3
help_string = "  --help                          Show this message and exit."
assert help_string in actual
```

joshmoore: You're right that this is pretty brittle. For example, the test will need to be updated for gh-904 already. I defer on to what degree we want the help output to count as public API which can't break. If we want to go for a less-strict solution, I'd imagine the `Usage:` line would be best, since that is what really broke with this PR, `starfish a b --help` was producing `starfish a --help b` output.

joshmoore: Pushed the simplify tests. Waiting on green.

PR #811, opened 2018-11-27
[easy] Remove the axes parameter from set_slice.

ttung:
It does not appear to be used.


ttung:
I am super wrong.  This is important, and it's actually an oversight that we never used this.  Will fix.

PR #813, opened 2018-11-28
Metaclass Logging 

shanaxel42:


starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,29 @@
...
+            result = func(*args, **kwargs)
+            # Spot detection returns a tuple, filtering returns an Imagestack
+            if isinstance(result, ImageStack):
+                # Only run on second run of function if in_place is True
+                in_place = 'in_place' in kwargs and kwargs['in_place'] is True
+                if not in_place:
+                    result.update_log(args[0])
+            return result
+        return helper
+
```
shanaxel42: So right now this will only add to the Imagestack log if a pipeline component returns an Imagestack, (all filters do). But it doesn't capture spot detection. Do we care? Another idea I had was to make the log an external class from Imagestack that both filtering and detection could log too throughout a pipeline.

starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,29 @@
...
+            cls.run = AlgorithmBaseType.run_with_logging(cls.run)
+
+    @staticmethod
+    def run_with_logging(func):
+        def helper(*args, **kwargs):
+            result = func(*args, **kwargs)
+            # Spot detection returns a tuple, filtering returns an Imagestack
+            if isinstance(result, ImageStack):
+                # Only run on second run of function if in_place is True
+                in_place = 'in_place' in kwargs and kwargs['in_place'] is True
```
ttung:
```suggestion
                in_place = kwargs.get('in_place',  False)
```

starfish/pipeline/algorithmbase.py
ttung: My 2c: probably not.  But it may make sense to propagate the logged data from ImageStack to IntensityTable like you suggested (on slack, I think?).

ambrosejcarr: I think we probably do want to record, on the intensity table, the set of steps that were executed to produce it (proagate, as Tony says). 

Since spot detection doesn't modify the ImageStack, I think it's correct not to log the execution of spot detection to the ImageStack. 

For completeness, we probably want to propagate from intensity table to expression matrix, too.

joshmoore:
Don't know if there's general interest, but I imagine that if a driver from the CLI for what happens to the record in this PR is needed, it could be plugged in at https://github.com/spacetx/starfish/pull/796/files#diff-bb077f1b04bf4d58f2e5b8b51e452a39R48

If that's not particularly needed, I'll close gh-796 for the time being.

shanaxel42:
Going to close out this PR in lieu of #885 that includes adding the log to the IntensityTable

PR #814, opened 2018-11-28
Make _iter_indices more robust.

ttung:
1. Default args should not be mutable values.
2. The existing implementation depends on set iteration order being the same across multiple runs.  While this appears to be true with cpython, as far as I can tell, there is no guarantee that this is always to be the case.

Test plan: make -j test lint mypy

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -659,14 +659,13 @@ class ImageStack:
 
         return tuple(slice_list), axes
 
-    def _iter_indices(self, indices: Set[Indices]={Indices.ROUND, Indices.CH}
-                      ) -> Iterator[Mapping[Indices, int]]:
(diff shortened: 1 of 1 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -659,24 +659,26 @@ def _build_slice_list(
...
 
         Yields
         ------
         Dict[str, int]
             Mapping of dimension name to index
 
         """
-        ranges = [np.arange(self.shape[ind]) for ind in indices]
+        if indices is None:
+            indices = {Indices.ROUND, Indices.CH}
```
shanaxel42: how is this different from have the default value be {Indices.ROUND, Indices.CH}?

ttung: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

shanaxel42: oh whaaaa wow I'm learning so much about python today

PR #815, opened 2018-11-28
Add support for labeled indices in imagestack / experiment builders

ttung:
The IDs for rounds/chs/zlayers for imagestack / experiment builders are now taken from a sequence rather than ranging from 0..(n-1).  At all existing call sites, we simply provide a range, so it's a functional noop.  However, this allows us to build tests for labeled indices.

Test plan: `make -j test lint mypy`

Depends on #803

### DIFF
--- a/starfish/experiment/builder/orderediterator.py
+++ b/starfish/experiment/builder/orderediterator.py
@@ -4,24 +4,24 @@ from typing import Iterator, Mapping, Sequence, Tuple
 from starfish.types import Indices
 
 
-def join_dimension_sizes(
+def join_dimension_labels(
         dimension_order: Sequence[Indices],
         *,
-        size_for_round: int,
-        size_for_ch: int,
-        size_for_z: int,
-) -> Sequence[Tuple[Indices, int]]:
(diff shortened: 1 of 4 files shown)


starfish/experiment/builder/orderediterator.py
```diff
@@ -31,12 +31,12 @@ def join_dimension_sizes(
...
 
 def ordered_iterator(
-        dimension_sizes: Sequence[Tuple[Indices, int]]
+        dimension_ids: Sequence[Tuple[Indices, Sequence[int]]]
 ) -> Iterator[Mapping[Indices, int]]:
     """
-    Given a sequence of tuples of dimensions and its respective size, return an iterator that steps
-    through all the possible points in the space.  The sequence is ordered from the slowest varying
-    dimension to the fastest varying dimension.
+    Given a sequence of tuples of dimensions and its respective sequence of ids, return an iterator
```
shanaxel42: I'm a little confused about the terminology, are ids different then labels? Maybe I should wait till the glossary though : )

PR #823, opened 2018-12-01
Refactor ImageStack construction from numpy arrays.

ttung:
This now uses the same abstractions from starfish.imagestack.parser to construct the ImageStack from numpy arrays.  This will allow us to propagate information about the labeled axes in a coherent way (in a future PR).

Test plan: make -j fast

Depends on #803


ttung:
> ah so if we create an imagestack from a numpy array we're just going to fill in fake coordinates for every tile?

`synthetic_tile` already does that.

https://github.com/spacetx/starfish/blob/f694c776b3e65e79543b88256fac348a57ea1d66/starfish/imagestack/imagestack.py#L1111
https://github.com/spacetx/starfish/blob/f694c776b3e65e79543b88256fac348a57ea1d66/starfish/experiment/builder/defaultproviders.py#L57

ttung:
> Ah, now I get where this is all going. Is this functionality tested somewhere?

`synthetic_stack` and `max_proj` exercise this code path.

PR #829, opened 2018-12-05
add docs for PipelineComponent and AlgorithmBase

### LINKED ISSUE #233: PipelineComponents need some kind of class docs
e.g., right now the classes in `__init__.py` for each `pipeline_component` look something like this: 

```python
class PixelSpotDetector(PipelineComponent):

    pixel_finder_algorithm_class = None
    pixel_finder_group: argparse.ArgumentParser

    @classmethod
    def implementing_algorithms(cls):
        return _base.PixelFinderAlgorithmBase.__subclasses__()

    @classmethod
    def add_to_parser(cls, subparsers):
        """Adds the pixel finder component to the CLI argument parser."""
        ...
```

We could use some Docs explaining how these work!

ambrosejcarr:
Adds documentation to close #233 

@ttung I took a stab at this, but your edits here would be welcome.

### DIFF
--- a/starfish/pipeline/algorithmbase.py
+++ b/starfish/pipeline/algorithmbase.py
@@ -1,6 +1,24 @@
-
-
 class AlgorithmBase:
+    """
+    This is the base class of any algorithm that starfish exposes. This base class is retrieved by
+    the PipelineComponent class to register the Algorithm to the CLI and API.
+
+    New algorithm classes, like "segmentation" must subclass AlgorithmBase. See, e.g.
+    starfish.image._segmentation._base.SegmentationAlgorithmBase
+
+    All algorithm implementations, like "Watershed", must then subclass their AlgorithmBase. In this
+    case, SegmentationAlgorithmBase. See, e.g.
+    starfish.image._segmentation.watershed.Watershed
+
+    This pattern combines with a PipelineComponent, in this case named "Segmentation" to enable
+    any implementing methods to be accessed from the API as:
+
+    starfish.image.Segmentation.<implementing algorithm (Watershed)>
+
+    and from the CLI as:
+
+    $> starfish segmentation watershed
+    """
     @classmethod
     def _get_algorithm_name(cls):
         """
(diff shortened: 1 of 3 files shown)


starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,24 @@
+class AlgorithmBase:
+    """
+    This is the base class of any algorithm that starfish exposes. This base class is retrieved by
```
ttung: this is not entirely true.  it's more like subclass of PipelineComponent retrieve subclasses of its corresponding AlgorithmBase.

starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,24 @@
...
+    the PipelineComponent class to register the Algorithm to the CLI and API.
 
+    New algorithm classes, like "segmentation" must subclass AlgorithmBase. See, e.g.
```
ttung: This is confusing to me.

ambrosejcarr:
Attempted to clarify.

starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,39 @@
...
+
+    This pattern exposes the API as follows:
+
+    `starfish.image.Segmentation.<implementing algorithm (Watershed)>`
+
+    and the CLI as:
+
+    `$> starfish segmentation watershed`
+
+    To create an entirely new algorithm class, like `Segmentation`, a new subclass of both
```
ttung: This is confusing.  Segmentation is not an algorithm class in the programming sense of the word class.  It might make sense if you're talking about a class of algorithms in the non-programming sense, but then it just seems like we should use a non-ambiguous term.

starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,39 @@
...
+    `AlgorithmBase` and `PipelineComponent` must be created.
+
+    To add to an existing class, an algorithm implementation must subclass the corresponding
```
ttung: This has the same problem.  I would find a different term other than "class".

PR #879, opened 2018-12-11
[EASY] Pass ImageFormat to ImageStack.export

joshmoore:
Allows passing `ImageFormat` to the export method. This does *not* reproduce the error seen by @ambrosejcarr that NUMPY files were being written with ".tiff" names. Both tests here currently produce files with the expected file-ending.

prerequisite for: tiff writing (gh-603)

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -1045,8 +1045,8 @@ class ImageStack:
             tileset,
             filepath,
             pretty=True,
-            tile_format=tile_format,
-            tile_opener=tile_opener)
+            tile_opener=tile_opener,
+            tile_format=tile_format)
 
     def max_proj(self, *dims: Indices) -> "ImageStack":
         """return a max projection over one or more axis of the image tensor
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -29,7 +29,13 @@ from scipy.ndimage.filters import gaussian_filter
 from scipy.stats import scoreatpercentile
 from skimage import exposure
 from skimage import img_as_float32, img_as_uint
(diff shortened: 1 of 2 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -1034,6 +1045,7 @@ def tile_opener(tileset_path, tile, ext):
             tileset,
             filepath,
             pretty=True,
+            tile_format=tile_format,
```
ttung: nit: would reorder these two, to match the call param order in slicedimage.

starfish/test/image/test_imagestack.py
```diff
@@ -239,3 +240,18 @@ def test_imagestack_to_intensity_table_no_noise(synthetic_spot_pass_through_stac
...
+))
+def test_imagestack_export(tmpdir, format, ending, count, recwarn):
+    """
+    Save a synthetic stack to files and check the results
+    """
+    stack = ImageStack.synthetic_stack()
+    stack_json = tmpdir / "output.json"
+    stack.export(str(stack_json), tile_format=format)
+    assert ImageStack.from_path_or_url(str(stack_json))
+    assert count == len([x for x in tmpdir.listdir() if str(x).endswith(ending)])
```
ttung: this test would not detect the scenario that @ambrosejcarr described (a file that ends with .tif but is actually in .npy format).

if you wanted to cover that case, then you could verify that one of the data files can be opened using ImageFormat's `reader_func`.

ttung: also, you don't need to provide `ending`.  ImageFormat exposes a property called `file_ext` that should cover this case.

joshmoore:
Suggestions pushed.

joshmoore:
Waiting on green and I'll land it.

PR #880, opened 2018-12-11
Rework how we parameterize the image format for starfish.experiment.builder

ttung:
While slicedimage supports images where the individual tiles are different formats, there isn't an API that actually allows _writing_ images of that nature.  We can just provide the tile format at the top level `write_experiment_json` call and write the data accordingly.

Test plan: `python examples/get_iss_breast_data.py /Users/ttung/Downloads/starfish_data/mignardi_breast_1/raw/ ../starfish-formatted-data/breast/ 1` and visual inspection.

### DIFF
--- a/starfish/experiment/builder/cli.py
+++ b/starfish/experiment/builder/cli.py
@@ -1,6 +1,7 @@
 import json
 
 import click
+from slicedimage import ImageFormat
 
 from starfish.types import Indices
 from . import AUX_IMAGE_NAMES, write_experiment_json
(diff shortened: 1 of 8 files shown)


starfish/experiment/builder/cli.py
```diff
@@ -47,8 +48,9 @@ def dimensions_option(name, required):
 
 def build(output_dir, fov_count, hybridization_dimensions, **kwargs):
     write_experiment_json(
-        output_dir, fov_count, hybridization_dimensions,
-        kwargs
+        output_dir, fov_count, ImageFormat.TIFF,
```
shanaxel42: is TIFF the default?

ttung: I am declaring it so. :)

I suspect it makes sense because it allows for a quick visualization of the stack after building the experiment object.  Any thoughts to the contrary?

ttung:
I'm going to land this, but feel free to raise objections to my default post-land.

PR #885, opened 2018-12-14
Provenance Logging

shanaxel42:
This PR uses a meta class in algorithm base to add provenance logging to the imagestack and intensity table. For the Imagestack the log is stored as a property named 'log'. For the IntensityTable the log is stored in the attrs['log']. After some discussion with @ambrosejcarr we decided that for more complicated runtime objects (imagestacks, codebooks) just serializing the result of their __repr__ methods was sufficient for writing the log to disk. 

Work that still needs to be done:
-Going to wait for Tony's various decoupling index values PRs to land before adding the log to Imagestack.export since they contain considerable changes to the method
-zero_by_channel refactor to use apply so that it doesn't get logged twice in an in_place instance. 

Added tests in ISS and DARTFISH pipelines that write the IntensityTable to a temp file then loads in back into memory to verify the JSON serialization / deserialization


starfish/pipeline/algorithmbase.py
```diff
@@ -1,6 +1,39 @@
...
+
+    @staticmethod
+    def run_with_logging(func):
+        def helper(*args, **kwargs):
+            result = func(*args, **kwargs)
+            # Spot detection returns a tuple or IntensityTable, filtering returns an Imagestack
+            if isinstance(result, ImageStack):
+                result.update_log(args[0])
+            # Spot detection
+            elif isinstance(result, tuple) or isinstance(result, IntensityTable):
```
ambrosejcarr: There are some methods that take IntensityTables and return IntensityTables (decoders, segmentation assignment). I don't think those are covered here yet, unless I'm missing something. 

I think this class deserves a docstring to explain how different approaches are treated. Some of the context from the PR explanation could probably go here.

starfish/test/full_pipelines/api/test_iss_api.py
```diff
@@ -119,5 +125,11 @@ def test_iss_pipeline_cropped_data():
     lab = TargetAssignment.Label()
     assigned = lab.run(label_image, decoded)
 
+    pipeline_log = assigned.attrs[LOG]
+
+    assert pipeline_log[0]['method'] == 'WhiteTophat'
+    assert pipeline_log[1]['method'] == 'FourierShiftRegistration'
+    assert pipeline_log[2]['method'] == 'BlobDetector'
```
ambrosejcarr: This is a clever way to test. 👍 

There are substrates here to evaluate the missing components I highlighted above.

starfish/pipeline/algorithmbase.py
shanaxel42: I think they'd fall under the 'isinstance(result, IntensityTable)' case, but you're right gotta add that docstring

ambrosejcarr: My read here is that if the result is an IntensityTable, but the second arg is not an ImageStack, nothing will happen. By my read, the cases I suggest fall into the "nothing happens" category.

shanaxel42: ahhhhhh I see what your saying, yes you're right

starfish/types/_constants.py
```diff
@@ -21,6 +21,9 @@ class Coordinates(AugmentedEnum):
 
 
 PHYSICAL_COORDINATE_DIMENSION = "physical_coordinate"
+
+LOG = 'log'
+
 """
```
ttung: This is the documentation for `PHYSICAL_COORDINATE_DIMENSION`.

You need to move the definition for LOG down, and also document it.

starfish/types/_constants.py
```diff
@@ -21,6 +21,9 @@ class Coordinates(AugmentedEnum):
 
 
 PHYSICAL_COORDINATE_DIMENSION = "physical_coordinate"
+
+LOG = 'log'
```
ttung:
```suggestion
LOG = "log"
```

starfish/intensity_table/intensity_table.py
```diff
@@ -172,6 +173,9 @@ def save(self, filename: str) -> None:
             Name of Netcdf file
 
         """
+        if LOG in self.attrs:
+            # if log info, json encode it
+            self.attrs[LOG] = LogEncoder().encode(self.attrs[LOG])
```
ttung: This is kind of odd because if you save it twice, the second time, you're going to be writing some weird stuff...

starfish/intensity_table/intensity_table.py
```diff
@@ -233,6 +237,9 @@ def load(cls, filename: str) -> "IntensityTable":
             loaded.coords,
             loaded.dims
         )
+        # maintain log
+        if LOG in loaded.attrs:
+            intensity_table.attrs[LOG] = loaded.attrs[LOG]
```
ttung: does this need to be deserialized?

shanaxel42: does it? maybe I should just always store then json encoded info in both the Imagestack and Intensitytable. Since it's just for reading do we care about maintaining the original objects?

starfish/types/_constants.py
shanaxel42: does it? maybe I should just always store then json encoded info in both the Imagestack and Intensitytable. Since it's just for reading do we care about maintaining the original objects?

shanaxel42:
Here's an example of the log from the end of the dartfish pipeline after it's been written and loaded from disk.

`<class 'list'>:[  
   {  
      'method':'ScaleByPercentile',
      'arguments':{  
         'p':100,
         'is_volume':False
      }
   },
   {  
      'method':'ZeroByChannelMagnitude',
      'arguments':{  
         'thresh':0.05,
         'normalize':False
      }
   },
   {  
      'method':'ZeroByChannelMagnitude',
      'arguments':{  
         'thresh':0.05,
         'normalize':False
      }
   },
   {  
      'method':'PixelSpotDetector',
      'arguments':{  
         'codebook':'"<xarray.Codebook (target: 96, c: 3, r: 6)>\\narray([[[0, 1, ..., 0, 0],\\n        [0, 0, ..., 1, 0],\\n        [1, 0, ..., 0, 0]],\\n\\n       [[0, 1, ..., 1, 0],\\n        [1, 0, ..., 0, 0],\\n        [0, 0, ..., 0, 0]],\\n\\n       ...,\\n\\n       [[0, 0, ..., 0, 0],\\n        [0, 1, ..., 1, 0],\\n        [1, 0, ..., 0, 0]],\\n\\n       [[0, 0, ..., 0, 0],\\n        [0, 0, ..., 0, 1],\\n        [1, 1, ..., 0, 0]]], dtype=uint8)\\nCoordinates:\\n  * target   (target) object \'MYH11\' \'RELN\' \'NR4A2\' \'COL20A1\' \'IGFBP5\' ...\\n  * c        (c) int64 0 1 2\\n  * r        (r) int64 0 1 2 3 4 5"',
         'metric':'euclidean',
         'distance_threshold':3,
         'magnitude_threshold':0.5,
         'min_area':5,
         'max_area':30,
         'norm_order':2,
         'crop_x':0,
         'crop_y':0,
         'crop_z':0
      }
   },
   {  
      'method':'PixelSpotDetector',
      'arguments':{  
         'codebook':'"<xarray.Codebook (target: 96, c: 3, r: 6)>\\narray([[[0, 1, ..., 0, 0],\\n        [0, 0, ..., 1, 0],\\n        [1, 0, ..., 0, 0]],\\n\\n       [[0, 1, ..., 1, 0],\\n        [1, 0, ..., 0, 0],\\n        [0, 0, ..., 0, 0]],\\n\\n       ...,\\n\\n       [[0, 0, ..., 0, 0],\\n        [0, 1, ..., 1, 0],\\n        [1, 0, ..., 0, 0]],\\n\\n       [[0, 0, ..., 0, 0],\\n        [0, 0, ..., 0, 1],\\n        [1, 1, ..., 0, 0]]], dtype=uint8)\\nCoordinates:\\n  * target   (target) object \'MYH11\' \'RELN\' \'NR4A2\' \'COL20A1\' \'IGFBP5\' ...\\n  * c        (c) int64 0 1 2\\n  * r        (r) int64 0 1 2 3 4 5"',
         'metric':'euclidean',
         'distance_threshold':0.6,
         'magnitude_threshold':0.5,
         'min_area':5,
         'max_area':30,
         'norm_order':2,
         'crop_x':0,
         'crop_y':0,
         'crop_z':0
      }
   }
]`

shanaxel42:
Going to make the imagestack exporting and loading bit a seperate PR

starfish/test/full_pipelines/api/test_dartfish.py
```diff
@@ -126,7 +127,28 @@ def test_dartfish_pipeline_cropped_data():
...
+
+    pipeline_log = spot_intensities.attrs['log']
+
+    assert pipeline_log[0]['method'] == 'ScaleByPercentile'
+    assert pipeline_log[1]['method'] == 'ZeroByChannelMagnitude'
+
+    # Test serialization / deserialization of IntensityTable log
+
+    fp = tempfile.NamedTemporaryFile()
+    spot_intensities.save(fp.name)
```
ttung: doesn't save convert the log into a json-encoded string? can you assert on the value of the log here?

shanaxel42: I do right below! I reload and assert the value.

ttung: I believe that if you reload and assert on the value, the value will be correct.  My concern is that if you encode it on save, and continue to manipulate the IntensityTable, the log will now be serialized, and _not_ correct.

shanaxel42: ahhhh I understand what you mean. I'll test for that

shanaxel42:
Changed the log format on the IntensityTable to be json encoded to avoid the saving funkiness. Going to go ahead and merge this one!

starfish/imagestack/imagestack.py
```diff
@@ -889,6 +890,36 @@ def tile_metadata(self) -> pd.DataFrame:
...
+         ex.
+            [('GaussianHighPass', {'sigma': (3, 3), 'is_volume': False}),
+            ('GaussianLowPass', {'sigma': (1, 1), 'is_volume': False}])]
+
+        Means that this imagestack was created by applying a GaussianHighPass Filter then
+        a GaussianLowPass Filter to a starting imagestack
+
+        Returns
+        -------
+        List[Tuple[str, dict]]
```
ttung: This should be `List[dict]`

starfish/imagestack/imagestack.py
```diff
@@ -889,6 +890,36 @@ def tile_metadata(self) -> pd.DataFrame:
...
+        """
+        return self._log
+
+    def update_log(self, class_instance):
```
ttung: I think you can do `     def update_log(self, class_instance: AlgorithmBase): -> None`

starfish/intensity_table/intensity_table.py
```diff
@@ -233,6 +239,9 @@ def load(cls, filename: str) -> "IntensityTable":
             loaded.coords,
             loaded.dims
         )
```
ttung: Why not just:

```
intensity_table = cls(
    loaded.data,
    loaded.coords,
    loaded.dims,
    attrs=loaded.attrs,
)

you can nuke the statement below.

File: starfish/util/JSONenocder.py [hunk 11]
```diff
@@ -0,0 +1,9 @@
+from json import JSONEncoder
+
+
+class LogEncoder(JSONEncoder):
+    def default(self, o):
```
ttung: ideally, this should have some class docs.

Issue #834, opened 2018-12-07
ImageStack multiprocessing problematic on Windows


ttung:
On Windows, multiprocessing is accomplished through `CreateProcess`, which is more like an exec.  If the `func` method of ImageStack depends on any in-process state (i.e., state that is accessible through globals instead of values serialized across the process boundary), then it will not be present in child processes.  See https://bugs.python.org/issue8713 for more details.

Possible solutions:
1. Ban the use of global state.  This is not possible to do programmatically.  Instead, we will have to use tests to verify that we don't violate this rule.
2. Require Windows users to use WSL.  POSIX fork() is supported in this mode.

PR #889, opened 2018-12-15
Multiprocessing heap patch

### LINKED ISSUE #833: ImageStack performance abysmally slow on OSX
Initializing an ImageStack (i.e., even before loading the data) is very slow on OSX.  This is because multiprocessing.Array on (python 3, OSX) is backed by a [non-anonymous file mmap](https://github.com/python/cpython/blob/c9566b8c454120e3d0ddb5ab970f262a6cd80077/Lib/multiprocessing/heap.py#L80).  This can be solved by [reverting the Heap class to the old functionality and forcing consistent use of `fork`](https://bugs.python.org/issue30919#msg298344).

shanaxel42:
fixes #833 

Patching multiprocessing.heap.Arena.__init__ so that it uses anonymous memory mapping instead of creating a zero filled temp file.

Significantly reduces Imagestack initialization and processing time (no longer freezes my computer, yay!)


starfish/imagestack/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def anonymous_arena_init(self, size, fd=-1):
+    """
+    Patching multiprocessing.heap.Arena.__init__ so that it uses anonymous memory mapping
+    instead of creating a temp file for shared memory amongst subprocess.
+    """
+    self.size = size
+    self.fd = fd  # still kept but is not used !
+    self.buffer = mmap.mmap(-1, self.size)
```
ttung: I'm guessing this will further complicate things re. https://github.com/spacetx/starfish/issues/834

starfish/imagestack/__init__.py
```diff
@@ -0,0 +1,15 @@
+import mmap
+from multiprocessing.heap import Arena
+
+
+def anonymous_arena_init(self, size, fd=-1):
+    """
+    Patching multiprocessing.heap.Arena.__init__ so that it uses anonymous memory mapping
+    instead of creating a temp file for shared memory amongst subprocess.
```
ttung: Can you reference https://bugs.python.org/issue30919 in the comments?

PR #892, opened 2018-12-16
Created specifications for output formats

ambrosejcarr:
Bug Fix: 
- Corrected API references for `IntensityTable` and `Codebook` 

New Docs: 
- Add Specification for `IntensityTable`
- Add Specification for `ExpressionMatrix`
- Add Placeholder Specification for `SegmentationMask` while we research what object we want to use for this.

### DIFF
--- a/docs/source/sptx-format/index.rst
+++ b/docs/source/sptx-format/index.rst
@@ -1,7 +1,12 @@
-.. _sptx_format:
+SpaceTx Format
+==============
 
-.. mdinclude:: ../../../sptx_format/README.md
+.. toctree::
+   :maxdepth: 1
+   :caption: Contents:
 
-.. _schema:
+.. toctree::
+   input_formats/index.rst
 
-.. mdinclude:: ../../../sptx_format/schema/README.md
+.. toctree::
+   output_formats/index.rst
--- a/docs/source/sptx-format/input_formats/SpaceTxFormat/index.rst
+++ b/docs/source/sptx-format/input_formats/SpaceTxFormat/index.rst
@@ -1,0 +1,3 @@
+.. _sptx_format:
+
+.. mdinclude:: ../../../../../sptx_format/README.md
--- a/docs/source/sptx-format/input_formats/Validation/index.rst
+++ b/docs/source/sptx-format/input_formats/Validation/index.rst
@@ -1,0 +1,3 @@
+.. _schema:
+
+.. mdinclude:: ../../../../../sptx_format/schema/README.md
--- a/docs/source/sptx-format/input_formats/index.rst
+++ b/docs/source/sptx-format/input_formats/index.rst
@@ -1,0 +1,12 @@
+Input Data Formats
+==================
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Contents:
+
+.. toctree::
+    SpaceTxFormat/index.rst
+
+.. toctree::
+    Validation/index.rst
--- a/docs/source/sptx-format/output_formats/ExpressionMatrix/index.rst
+++ b/docs/source/sptx-format/output_formats/ExpressionMatrix/index.rst
@@ -1,0 +1,46 @@
+.. _ExpressionMatrixSpecification:
+
+ExpressionMatrix
+================
+The :py:class:`ExpressionMatrix` is a 2-dimensional :code:`cells (x)` by :code:`genes (y)` array
+whose values contain the expression of a gene in a particular cell. The :py:class:`ExpressionMatrix`
+is additionally annotated with the :code:`x, y, z` pixel coordinates of the centroid of the cell in
+a pixel space, and :code:`xc, yc, zc` in physical coordinate space. Additional metadata may be added
+at the user's convenience to either the cells or genes.
+
+Data
+----
+Gene expression are stored as numeric values, typically as integers in Image-based transcriptomics
+experiments, since they represent counted fluorescent spots, each corresponding to a single detected
+RNA molecule.
+
+Metadata
+--------
+
+:code:`cells; cell_id (int):` cell identifier
+
+:code:`cells; x, y, z (int):` coordinates of cell centroid in pixel space
+
+:code:`cells; xc, yc, zc (int):` coordinates of cell centroid in global coordinate space (um)
+
+:code:`genes; gene_id (int):` GENCODE gene ID
+
+:code:`genes; gene_name (int):` Human-readable gene symbol (e.g. HGNC gene symbol for human data)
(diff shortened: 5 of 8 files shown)


docs/source/sptx-format/output_formats/ExpressionMatrix.rst
```diff
@@ -0,0 +1,46 @@
...
+:code:`cells; xc, yc, zc (int):` coordinates of cell centroid in global coordinate space (um)
+
+:code:`genes; gene_id (int):` GENCODE gene ID
+
+:code:`genes; gene_name (int):` Human-readable gene symbol (e.g. HGNC gene symbol for human data)
+
+Implementation
+--------------
+Starfish Implements the :py:class:`ExpressionMatrix` as an :code:`xarray.DataArray` object to take
+advantage of `xarray's`_ high performance, flexible metadata storage capabilities, and Serialization
```
ttung:
```suggestion
advantage of `xarray's`_ high performance, flexible metadata storage capabilities, and serialization
```

docs/source/sptx-format/output_formats/IntensityTable.rst
```diff
@@ -0,0 +1,85 @@
+IntensityTable
+==============
+
+The :py:class:`IntensityTable` summarizes information about RNA features that are detected in
+Image-based transcriptomics experiments. The most common features are spots, pixels, or connected
```
ttung:
```suggestion
image-based transcriptomics experiments. The most common features are spots, pixels, or connected
```

docs/source/sptx-format/output_formats/IntensityTable.rst
```diff
@@ -0,0 +1,85 @@
...
+pixels that result from the fluorescence of a probe that has been experimentally attached to an RNA
+molecule.
+
+To gather information on the RNA present in a tissue slice, Image-based transcriptomics experiments
+require the imaging of the same RNA molecules over multiple rounds and across multiple fluorescence
+channels. In multiplex assays, the identity of these spots is only determined after measuring the
+spot's intensity across all rounds and channels. In sequential experiments, each round and channel
```
ttung:
```suggestion
spot's intensity across all rounds and channels. In sequential assays, each round and channel
```

docs/source/sptx-format/output_formats/IntensityTable.rst
```diff
@@ -0,0 +1,85 @@
+IntensityTable
+==============
+
+The :py:class:`IntensityTable` summarizes information about RNA features that are detected in
+Image-based transcriptomics experiments. The most common features are spots, pixels, or connected
+pixels that result from the fluorescence of a probe that has been experimentally attached to an RNA
+molecule.
+
+To gather information on the RNA present in a tissue slice, Image-based transcriptomics experiments
```
ttung:
```suggestion
To gather information on the RNA present in a tissue slice, image-based transcriptomics experiments
```

ttung:
There's enough important stuff here that may justify @dganguli 's review, fwiw.

docs/source/sptx-format/output_formats/ExpressionMatrix.rst
```diff
@@ -0,0 +1,46 @@
...
+================
+The :py:class:`ExpressionMatrix` is a 2-dimensional :code:`cells (x)` by :code:`genes (y)` array
+whose values contain the expression of a gene in a particular cell. The :py:class:`ExpressionMatrix`
+is additionally annotated with the :code:`x, y, z` coordinates of the centroid of the cell in
+a pixel space, and :code:`xc, yc, zc` in global coordinate space. Additional metadata may be added
+at the user's convenience to either the cells or genes.
+
+Data
+----
+Gene expression are stored as numeric values, typically as integers in Image-based transcriptomics
```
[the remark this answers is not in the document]
ambrosejcarr: Unfortunately yes.

ambrosejcarr:
@ttung I made your changes. I think I updated the structure of the folders in between, so couldn't apply your changes directly (there were conflicts with Josh's update to the validation documentation).

joshmoore:
> there were conflicts with Josh's update to the validation documentation

Doh, sorry about that. This might be a candidate for getting exposed on RTD then so we have multiple eyes on it.

ambrosejcarr:
closes #840,#891

joshmoore:
After gh-925, I can't help but wonder if codebooks shouldn't also fall into this category of serialized xarray objects.

ambrosejcarr:
> After gh-925, I can't help but wonder if codebooks shouldn't also fall into this category of serialized xarray objects.

If we're comfortable with (for now) netcdf as a cross-platform standard, we could go this route. In the future, I imagine we'd dump them as Zarr? Could you flesh out how you'd suggest we approach this @joshmoore ?

joshmoore:
> If we're comfortable with (for now) netcdf as a cross-platform standard, we could go this route.

You're _less_ comfortable with the support for codebook compared to say the intensity table because the codebook is more core, e.g. an input rather than an output? Might need to think that through... but I imagine if we're looking at _any_ use of netcdf then the cross-platformness issue will need to be dealt with regardless. (I currently assume it will be fine)

My motivation was largely coming from a mixture of the fact that `Codebook` is-a `DataArray` and the sizes that are possible.

> In the future, I imagine we'd dump them as Zarr?

There's likely a question here of whether or not to support the more general xarray io methods or to restrict to a subset. In the former case, you'd get json, netcdf, and eventually zarr directly.

> Could you flesh out how you'd suggest we approach this @joshmoore ?

Really, for any of the classes that are-a `xarray` type, having a single format (essentially, `DataArray` and `Dataset`) feels safe enough.

In my case, having installed the netcdf libraries, using `c.to_netcdf("codebook.nc")` and reading it back in with `xr.open_dataset("codebook.nc")` produced something I could immediately relate to:

```
Out[9]:
<xarray.Dataset>
Dimensions:                        (c: 2, r: 8, target: 140)
Coordinates:
  * target                         (target) object 'Blank-1' 'Blank-10' ...
  * c                              (c) int64 0 1
  * r                              (r) int64 0 1 2 3 4 5 6 7
Data variables:
    __xarray_dataarray_variable__  (target, c, r) uint8 ...
```

Other than the somewhat unfortunate `__xarray_dataarray_variable__` this actually seems quite portable, even outside the starfish library.

That being said, during that investigation, `json.store(c.to_json(), open("codebook.json"))` proved to be reasonably fast as well and products the same results without the opaque `nc` file, which is what made me wonder if codebooks "schema" should essentially be just the xarray json serialization.

PR #888, opened 2018-12-14
Ajc seqfish tilefetcher 2

joshmoore:
Minor click updates to prevent null pointer exceptions (and deal with not having the codebook)

PR #894, opened 2018-12-16
Add SeqFISH data to Starfish

ambrosejcarr:
This PR adds SeqFISH data and test data to Starfish: 
* Adds the `examples/format_seqFISH.py` `TileFetcher`
* Adds the `SeqFISH.ipynb` notebook to demonstrate data loading, and add it to run in Travis. 
* Adds `Starfish.data.SeqFISH`

This triggers the following errors that I don't understand. It seems like the types I'm passing should be compatible, unless I don't understand how Union works:
```bash
$ make mypy
mypy --ignore-missing-imports starfish examples sptx_format
examples/format_seqFISH.py:92: error: Argument 2 to "SeqFISHTile" has incompatible type "Mapping[Coordinates, Tuple[float, float]]"; expected "Mapping[Union[str, Coordinates], Union[int, float, Tuple[Union[int, float], Union[int, float]]]]"
examples/format_seqFISH.py:177: error: Argument "primary_image_dimensions" to "write_experiment_json" has incompatible type "Dict[Indices, int]"; expected "Mapping[Union[str, Indices], int]"
make: *** [mypy] Error 1
```


joshmoore:
Pushed commits from gh-888 and a potential fix for mypy.

examples/format_seqFISH.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, input_dir: str) -> None:
         self.input_dir = input_dir
 
     @property
-    def coordinates(self) -> Mapping[Coordinates, Tuple[float, float]]:
+    def coordinates(self) -> Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]:
```
ambrosejcarr: If this is the fix, mypy is being a bit overactive here, no?

joshmoore: I was certainly surprised myself. I tried a couple of different ways of assigning `X` to `Union[X, Y]` to no avail. I'm certainly no mypy expert though. There were suggestions of making the union an alias used everyone, which isn't really what you're trying to achieve. Then there's [@overload](https://stackoverflow.com/a/49101404/56887) but since it's not used in the current codebase, I didn't try that route.

ttung: No, this makes sense.  Parameters for subclasses can accept a broader type definition, but not a narrower one.  If mypy is built correctly, return values should behave in the opposite way.

examples/format_seqFISH.py
```diff
@@ -0,0 +1,197 @@
...
+    ):
+        self._file_path = file_path
+        self._z = z
+        self._ch = ch
+        self._coordinates = coordinates
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        """Gets image shape directly from the data"""
+        return self.tile_data().shape
```
ttung: The downside of this approach is that you will decode the file twice.

Option 1: don't care, just decode twice.  easiest to read code at the cost of perf.
Option 2: cache the shape from the first tile, return it for all subsequent invocations of shape.  when `tile_data` is eventually called, verify that the shape matches the tile's shape.  this has significantly more complicated code.

examples/format_seqFISH.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, input_dir: str) -> None:
+        """Implement a TileFetcher for a single SeqFISH Field of View."""
+        self.input_dir = input_dir
+
+    @property
+    def coordinates(self) -> Mapping[Union[str, Coordinates], Union[Number, Tuple[Number, Number]]]:
+        """Returns dummy coordinates for this single-FoV TileFetcher"""
+        return {
+            Coordinates.X: (0., 0.),
```
ttung: LOL.

on a serious note, even when it's a single FOV, it should have a non-zero sizes in physical coordinate space.  i suspect assigning a physical coordinate of a spot will require dividing by the size of the tile in physical coordinate space.

examples/format_seqFISH.py
```diff
@@ -0,0 +1,197 @@
...
+        aux_name_to_dimensions={},
+        primary_tile_fetcher=primary_tile_fetcher,
+        tile_format=ImageFormat.TIFF,
+        # postprocess_func=post_process_func,
+        dimension_order=(Indices.ROUND, Indices.CH, Indices.Z)
+    )
+
+    if codebook_csv:
+        with open(os.path.join(output_dir, "codebook.json"), "w") as f:
+            codebook = parse_codebook(codebook_csv)
```
ttung: IMO, we should be constructing a Codebook object and using the code in that class to serialize.

examples/format_seqFISH.py
```diff
@@ -0,0 +1,197 @@
...
+    primary_tile_fetcher = SeqFISHTileFetcher(os.path.expanduser(input_dir))
+
+    # This is hardcoded for this example data set
+    primary_image_dimensions: Mapping[Union[str, Indices], int] = {
+        Indices.ROUND: 5,
+        Indices.CH: 12,
+        Indices.Z: 29,
+    }
+
+    # def post_process_func(experiment_json_doc):
```
ttung: ehhh?

examples/format_seqFISH.py
ambrosejcarr: Oh I see. That makes a lot of sense, thanks @ttung . I wasn't considering the subclass angle.

examples/format_seqFISH.py
ambrosejcarr: Given the `lru_cache` that's already in place around `imread`, the only thing that should be redone is the slicing of the loaded data by z and channel. That has a minimal performance loss. 

Am I interpreting the cache's functionality incorrectly?

examples/format_seqFISH.py
ambrosejcarr: Sorry. 😞

ambrosejcarr:
Changes addressed, ready for re-review.

starfish/data/__init__.py
```diff
@@ -83,3 +83,30 @@ def ImagingMassCytometry(use_test_data: bool=False):
...
+    See Also
+    --------
+    Manuscript for Intron-SeqFISH: https://doi.org/10.1016/j.cell.2018.05.035
+
+    """
+    suffix = "-TEST" if use_test_data else ""
+    url = (
+        f"https://d2nhj9g34unfro.cloudfront.net/browse/formatted/20181211/"
+        f"seqfish{suffix}/experiment.json"
+    )
```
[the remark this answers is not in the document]
ambrosejcarr: 😀

examples/format_seqFISH.py
ttung: We read the shape on the first loop, and then we read the image data on a second loop.  This means we will read it twice.  As I said, this may not matter, but you should commit this code with that knowledge. :)

Issue #825, opened 2018-12-03
transfer_physical_coordinates... user error or bug?  let's find out.


berl:
```from starfish.intensity_table.intensity_table_coordinates import transfer_physical_coords_from_imagestack_to_intensity_table as tpc
f0=experiment.fovs_by_name("fov_050")[0]

it0 = starfish.IntensityTable.load(os.path.join(basepath,"fov_050.ncdf"))

tpc(f0[FieldOfView.PRIMARY_IMAGES], it0)
```

fails with the traceback below.  it looks like it's related to z indexing/coordinates or I'm just not doing it right.   FWIW, I'd actually be happy with just x and y at the moment...

> ---------------------------------------------------------------------------
```KeyError                                  Traceback (most recent call last)
~/anaconda3/envs/starfishpy36/lib/python3.6/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   3077             try:
-> 3078                 return self._engine.get_loc(key)
   3079             except KeyError:

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index_class_helper.pxi in pandas._libs.index.Int64Engine._check_type()

KeyError: 6.145187601957586

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-139-eb7bb30de975> in <module>
      4 it0 = starfish.IntensityTable.load(os.path.join(basepath,"fov_050.ncdf"))
      5 
----> 6 tpc(f0[FieldOfView.PRIMARY_IMAGES], it0)

~/anaconda3/envs/starfishpy36/lib/python3.6/site-packages/starfish-0.0.30-py3.6.egg/starfish/intensity_table/intensity_table_coordinates.py in transfer_physical_coords_from_imagestack_to_intensity_table(image_stack, intensity_table)
     48                 pixel_x,
     49                 pixel_y,
---> 50                 image_stack._tile_shape)
     51             # Assign to coordinates arrays
     52             intensity_table[Coordinates.X.value][ind] = physical_coords[0]

~/anaconda3/envs/starfishpy36/lib/python3.6/site-packages/starfish-0.0.30-py3.6.egg/starfish/imagestack/physical_coordinate_calculator.py in get_physical_coordinates_of_spot(coords_array, tile_indices, pixel_x, pixel_y, tile_shape)
[...]

PR #898, opened 2018-12-18
 Converting pixel index values to ints in TrackpyLocalMaxPeakFinder 

### LINKED ISSUE #890: TrackpyLocalMaxPeakFinder creates intenisty table with pixel values as floats instead of ints.
this is the cause of #825

shanaxel42:
fixes #890 and #825 unblocking Brian


ttung:
Let me know if you have thoughts on either of my above comments. :)

shanaxel42:
> Let me know if you have thoughts on either of my above comments. :)

My plan was to add "transfer physical coords tests" that would capture this when I added the function detect spots.

 > Also, if we have certain requirements regarding the types of the attributes, maybe we should assert on them in SpotAttributes

This is a good idea.

PR #902, opened 2018-12-19
RFC: Support crop-on-load for TileSet -> ImageStack construction

ttung:
Create new classes (CroppingTileCollectionData, CroppingTileData) for representing a cropped view of a (TileCollectionData, TileData).  Because cropping happens at the time of ImageStack construction, the memory requirements are significantly lowered to O(number-of-cropped-tiles * size-of-one-cropped-tile + size-of-one-original-tile).

Test plan: new imagestack construction tests that verify that the cropping occurs as expected.

Depends on #828, #903

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -46,6 +46,7 @@ from starfish.experiment.builder import build_image, TileFetcher
 from starfish.experiment.builder.defaultproviders import OnesTile, tile_fetcher_factory
 from starfish.imagestack import indexing_utils, physical_coordinate_calculator
 from starfish.imagestack.parser import TileCollectionData, TileKey
+from starfish.imagestack.parser.crop import CropParameters, CroppingTileCollectionData
 from starfish.imagestack.parser.numpy import NumpyData
 from starfish.imagestack.parser.tileset import parse_tileset
 from starfish.intensity_table.intensity_table import IntensityTable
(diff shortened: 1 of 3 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -224,22 +225,31 @@ def __repr__(self):
...
         Returns:
         --------
         ImageStack :
             An ImageStack representing encapsulating the data from the TileSet.
         """
-        parsed = parse_tileset(tileset)
-        return cls(*parsed)
+        tile_shape, tile_data = parse_tileset(tileset)
+        if crop_parameters is not None:
+            tile_shape = crop_parameters.crop_shape(tile_shape)
```
shanaxel42: Do you need to do the cropping of the shape independently of the image cropping? Couldn't you just crop the TileCollection then derive the shape from the result?

starfish/imagestack/parser/crop.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            start = min(size, crop.start)
+
+        stop: int
+        if crop is None or crop.stop is None:
+            stop = size
+        elif crop.stop is not None and crop.stop < 0:
+            stop = max(0, size + crop.stop)
+        else:
+            stop = min(size, crop.stop)
```
shanaxel42: If someone provides crop parameters outside the range of the axis should we through an error instead of setting the start and stop to 0 and the size?

starfish/imagestack/imagestack.py
ttung: Yes, one could do that, but `CroppingTileCollectionData` doesn't know its shape.  One would need to fetch a tile and grab its shape to know (kind of duplicating the logic in `starfish.imagestack.parser.tileset._parser.parse_tileset`

starfish/imagestack/parser/crop.py
ttung: I don't have a strong opinion here.  I was obeying the semantics of python slicing, but I'm happy to change it up if you think it's better, @shanaxel42.

PR #905, opened 2018-12-20
[easy] Fix documentation for ImageStack.sel

ttung:
screen shot 2018-12-19 at 5 39 19 pm

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -335,19 +335,19 @@ class ImageStack:
         Examples
         --------
 
-        Create an Imagestack using the ``synthetic_stack`` method::
-        >>> from starfish import ImageStack
-        >>> from starfish.types import Indices
-        >>> stack = ImageStack.synthetic_stack(5, 5, 15, 200, 200)
-        >>> stack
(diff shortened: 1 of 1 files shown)


ttung:
Without the indentation, it looks like:

screen shot 2018-12-19 at 5 40 52 pm

starfish/imagestack/imagestack.py
```diff
@@ -335,19 +335,19 @@ def sel(self, indexers: Mapping[Indices, Union[int]]):
         Examples
         --------
 
-        Create an Imagestack using the ``synthetic_stack`` method::
```
joshmoore: Looks like we have a mix of ending these Example introductions with `::`, :`:, and no colon as here.

ttung: Can someone explain what the differences in all these might be?  Hard to google this sort of thing. :(

joshmoore: Yeah, I don't know. Yours certainly looked ok. But `::` did as well. I didn't look up what `:` came out looking like. Probably just best if we pick one we like and all agree on it.

PR #914, opened 2018-12-21
local max peak finder should work in 3d images

ttung:
There is still a z dimension, even though it's size=1.

Test plan: notebooks/py/osmFISH.py seems to run.


ttung:
Also, @ambrosejcarr / @dganguli, I suspect that this code does not conform to the ideal of axis order independence.  I suspect a lot more of the code has this problem.  Post-MVP, we may wish to explore whether we want to adhere to that ideal or scrap it.

PR #922, opened 2019-01-01
Glossary update.

ttung:
Really, just read glossary.md.  I'll complete the refactorings when the rest are conceptually approved. :)

Built on top of #902


docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
...
-### Indices (Tile):
-Indices indicate which round, channel, and z-plane a given (x, y) Tile corresponds to.
+### Axes (Tile):
+Each tile is located along three axes, which are round, channel, and z-plane.  Each (x, y) Tile is assigned values for each of these three axes.
+
+### Labels (Tile):
+Each tile is located along three axes, which are round, channel, and z-plane.  Each (x, y) Tile is assigned values for each of these three axes.  Because it is possible to crop by round, channel, or z-plane, not all values are valid.  The possible set of values for a given axis is the labels for that axis.
+
+### Index (Tile):
+An index indicates the assigned value for a given axis for a (x, y) Tile.
```
ambrosejcarr: Are these always numeric? If so I'd state that explicitly.

docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
...
 
 ### Coordinates (Tile):
 Coordinates refer to the physical location of a Tile with respect to some independent reference.
 
-### Indices (Tile):
-Indices indicate which round, channel, and z-plane a given (x, y) Tile corresponds to.
+### Axes (Tile):
+Each tile is located along three axes, which are round, channel, and z-plane.  Each (x, y) Tile is assigned values for each of these three axes.
+
+### Labels (Tile):
```
ambrosejcarr: I'd reorganize the last sentence: 

```diff
- The possible set of values for a given axis is the labels for that axis.
+ Labels are the possible set of values for each axis.
```

docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
...
+Each tile is located along three axes, which are round, channel, and z-plane.  Each (x, y) Tile is assigned values for each of these three axes.  Because it is possible to crop by round, channel, or z-plane, not all values are valid.  The possible set of values for a given axis is the labels for that axis.
+
+### Index (Tile):
+An index indicates the assigned value for a given axis for a (x, y) Tile.
+
+### Selectors (Tile):
+The full set of indices (round, channel, and z-plane) for a given (x, y) Tile.
```
ambrosejcarr: I think this would be clearer to state this as follows: 

```diff
- The full set of indices (round, channel, and z-plane) for a given (x, y) Tile.
+ The (round, channel, and z-plane) Indices that uniquely identify an (x, y) Tile. 
```

If these are always expressed as `Dict[Indices, Int]`, I would state that here as well, since that makes it impossible to misinterpret what this is.

docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
 A collection of Image Tiles corresponding to a specific volume or plane of the sample, under which the signal for all channels and all imaging rounds were acquired. All tiles within this FOV are the same size, but the manifest allows for different spatial coordinates for different imaging rounds or channels (to accommodate slight movement between rounds, for example).
 In microscopy, a field of view corresponds to the area of the camera sensor, and many such fields of view are expected to be taken per tissue slice.
```
kevinyamauchi: I think it's more accurate to say something like "a field of view corresponds to the area of the camera sensor mapped to the sample plane" (i.e., the area of the field of view in physical coordinates is not the area of the camera sensor, but the area of the camera sensor scaled by the magnification of the microscope.)

docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
...
 
-### Indices (Field of View):
-Indices are used to navigate within a field of view, and comprise the physical dimensions of pixel space (X, Y), the z-plane, channel (C) and imaging round (R) of a field of view.
-
 ### Image Tile
 A single plane, single channel 2D image. In the manifest, each tile has information about it’s (X,Y,Z) coordinates in space, and information about which imaging round (R) and/or fluorescence channel (C) it was acquired under.
```
kevinyamauchi: For completeness it's probably good to add "single round": " A single plane, single channel, single round 2D image

docs/source/glossary/glossary.md
```diff
@@ -41,17 +41,23 @@ The data manifest is a file that includes the locations of all fields of view fo
...
 
 ### Coordinates (Tile):
 Coordinates refer to the physical location of a Tile with respect to some independent reference.
```
kevinyamauchi: Are the coordinates for a specific location within a tile (e.g., upper left corner, centroid)? Unless it's somewhere else in the documentation, I think this should be specified because it will be critical for registering results in multimodal experiments.

docs/source/glossary/glossary.md
joshmoore: I'd also not repeat the Axis definition under Labels (I had to compare the sentences word by word to see if there was new info) Not sure if "Because it is possible to crop by round, channel, or z-plane, not all values are valid." best belongs under Axes or Labels, though.

docs/source/glossary/glossary.md
ttung: Fixed.

docs/source/glossary/glossary.md
ttung: Done, and done.

docs/source/glossary/glossary.md
ttung: done.

ttung:
> It now raises questions for me about our output formats. The Intensity table shares the same three entities of Axes, Indices, and Selectors. However, to form a selector for a Codebook or IntensityTable you need to draw from Axes and Features, which feels wrong to me.

> Does your model here still seem optimal in light of our other objects? If yes, I think this is good to merge.

Um.. I wish I understood the question here. :(

neuromusic:
Can we add "pipeline recipe" to the glossary?

ambrosejcarr:
> Um.. I wish I understood the question here. :(

Do your definitions of axes, indices, and selectors still work given that some of them are used by IntensityTable and Codebook? 

Codebook has `round, channel`

IntensityTable has `round, channel` as matrix dimensions, and uses `x, y, z` and `xc, yc, zc` as feature metadata.

ttung:
> Do your definitions of axes, indices, and selectors still work given that some of them are re-used by IntensityTable and Codebook?

> Codebook has round, channel

> IntensityTable has round, channel as matrix dimensions, and uses x, y, z and xc, yc, zc as feature metadata.

Ummmm neither Codebook nor IntensityTable refer to axes/indices/selectors...?

ambrosejcarr:
We're going to address my question in a follow-up issue. No further concerns!

PR #923, opened 2019-01-02
[easy] Fix sphinx error due to ws

ttung:
The current docblock yields this warning (which is treated as an error):

`/home/travis/build/spacetx/starfish/starfish/imagestack/imagestack.py:docstring of starfish.imagestack.imagestack.ImageStack.set_slice:38:Unexpected indentation.`

Blame: #818

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -427,6 +427,7 @@ class ImageStack:
         Examples
         --------
         Setting a slice indicated by scalars.
+
             >>> import numpy as np
             >>> from starfish import ImageStack
             >>> from starfish.types import Indices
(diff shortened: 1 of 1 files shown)


neuromusic:
If we want to make sure the docs can build before merging into master, readthedocs has docker containers that we can use to build the docs & add them to the checks on PRs. Here's how I've done it on circleci: https://github.com/AllenInstitute/neuroglia/blob/master/.circleci/config.yml#L143

ttung:
> If we want to make sure the docs can build before merging into master, readthedocs has docker containers that we can use to build the docs & add them to the checks on PRs. Here's how I've done it on circleci: https://github.com/AllenInstitute/neuroglia/blob/master/.circleci/config.yml#L143

We do have integration in our CI.  I was impatient about landing.  Whoops!

neuromusic:
ohhhhhh. tsk, tsk :P

ttung:
ok one is green.  going to land.  our tests take too long. :(

joshmoore:
> We do have integration in our CI. I was impatient about landing.

I've had at least one situation where (long) tests pass but readthedocs is still unhappy. I'd vote for copying @neuromusic's config assuming it doesn't increase the total time.

PR #926, opened 2019-01-03
Save log to tileset_data.extras on Imagestack export

shanaxel42:


starfish/imagestack/imagestack.py
```diff
@@ -122,7 +126,12 @@ def __init__(
         }
         self._tile_shape = tile_shape
         self._tile_data = tile_data
-        self._log: List[dict] = list()
+
+        # check for existing log info
+        if STARFISH in tile_data.extras:
+            self._log = loads(tile_data.extras[STARFISH])[LOG]
```
ttung: To make it 100% crystal clear that this is a starfish-ism, let's add something to the ImageStack parsers such that `TileCollectionData` explicitly provides controlled vocabulary metadata.

That might look like:
```
class StarfishMetadata:
    def log(self) -> List[dict]:
        ....

class TileCollectionData:
    def get_starfish_metadata(self) -> StarfishMetadata:
        raise NotImplementedError

class TileSetData:
    def get_starfish_metadata(self) -> StarfishMetadata:
        return json.loads(self.extras[STARFISH])
```

starfish/intensity_table/intensity_table.py
```diff
@@ -167,7 +167,7 @@ def from_spot_data(
     def get_log(self):
         """Deserialize and return a list of pipeline components that have been applied
          throughout a starfish session to create this Intensity Table"""
-        return loads(self.attrs[LOG])
+        return loads(self.attrs[STARFISH])[LOG]
```
ttung: What happens if we load an IntensityTable with no log?  Or an intensity table without a starfish section in the attrs field?

shanaxel42: would throwing an error that says "no log info available" be alright?

starfish/imagestack/imagestack.py
ttung: This should check if the extras.starfish contains a log entry.

starfish/types/_constants.py
```diff
@@ -26,6 +26,10 @@ class Coordinates(AugmentedEnum):
 This is the xarray dimension name for the physical coordinates of the tiles.
 """
 
+STARFISH = 'starfish'
```
ttung: Consider renaming this STARFISH_EXTRAS or STARFISH_EXTRAS_KEY.

PR #928, opened 2019-01-04
starfish.display.stack

ambrosejcarr:
- [x] features complete
- [x] testing complete
- [x] code style
- [x] complete code documentation
- [ ] add to sphinx/readthedocs

Features:
- Displays ImageStacks using Napari
- If provided with an IntensityTable generated from the provided ImageStack, will display spots
- will max project over selected indices
- will hide spots (per round and channel) to display spots above a threshold in individual tiles. 
- will hide any pre-existing entries in the IntensityTable that are `np.NaN` (enables pre-masking). This also solves the smFISH use case where there is only one entry per intensitytable feature and the rest are `np.NaN`. 
- all combinations of the above are functional (e.g. masking over max projections)

@kevinyamauchi it should have everything you need for showing spots. Let me know if anything is confusing. This should get you started: 
cc @berl in case there's interest. 

install: 
```
$ pip install -e .[napari]
```

use:
```python
import starfish.display
help(starfish.display.stack)
```

### DIFF
--- a/starfish/display/stack.py
+++ b/starfish/display/stack.py
@@ -1,0 +1,223 @@
+import warnings
+from collections import OrderedDict
+from typing import Iterable, List, Optional, Set, Tuple, Union
+
+import numpy as np
+
+from starfish.imagestack.imagestack import ImageStack
+from starfish.intensity_table.intensity_table import IntensityTable
+from starfish.types import Axes, Features
+
+
+def _normalize_axes(axes: Iterable[Union[Axes, str]]) -> List[str]:
+    """convert all axes to strings"""
+
+    def _normalize(axes):
+        if isinstance(axes, str):
+            return axes
+        elif isinstance(axes, Axes):
+            return axes.value
+        else:
+            raise TypeError(f"passed axes must be str or Axes objects, not {type(axes)}")
+
+    return list(_normalize(i) for i in axes)
+
+
+def _max_intensity_table_maintain_dims(
+    intensity_table: IntensityTable, dimensions: Set[Axes],
+) -> IntensityTable:
+    """
+    Maximum project an IntensityTable over dimensions, retaining singletons in place of dimensions
+    reduced by the max operation.
+
+    For example, xarray.max(Axes.CH.value) would usually produce a 2-d (features, rounds) output.
+    This function ensures that the output is instead (features, 1, rounds), by inserting singleton
+    dimensions in place of any projected axes.
+
+    Parameters
+    ----------
+    intensity_table : IntensityTable
+        Input intensities
+    dimensions : Set[Axes]
+        Dimensions to project
+
+    Returns
+    -------
+    IntensityTable :
+        full-dimensionality (3-d) IntensityTable
+
+    """
+    str_dimensions = _normalize_axes(dimensions)
+    initial_dimensions = OrderedDict(intensity_table.sizes)
+    projected_intensities = intensity_table.max(str_dimensions)
+    expanded_intensities = projected_intensities.expand_dims(str_dimensions)
+    return expanded_intensities.transpose(*tuple(initial_dimensions.keys()))
+
+
+def _mask_low_intensity_spots(
+    intensity_table: IntensityTable,
+    intensity_threshold: float
+) -> np.ndarray:
+    """
+    returns a flattened boolean mask (c order flattening) that is True where
+    (feature, round, channel) combinations whose intensities are above or equal to intensity
+    threshold and where values are not nan
+    """
+    with warnings.catch_warnings():
+        # we expect to have invalid values (np.nan) in IntensityTable, hide the warnings from users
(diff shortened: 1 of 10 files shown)


ambrosejcarr:
closes #822

ambrosejcarr:
Going to leave the sphinx docs for @neuromusic to do later.

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+
+import numpy as np
+
+from starfish.imagestack.imagestack import ImageStack
+from starfish.intensity_table.intensity_table import IntensityTable
+from starfish.types import Features, Indices
+
+
+def _normalize_indices(indices: Iterable[Union[Indices, str]]) -> List[str]:
+    """convert all indices to strings"""
```
ttung: When you rebase, take care to update the documentation to the new glossary.

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _normalize(indices):
+        if isinstance(indices, str):
+            return indices
+        elif isinstance(indices, Indices):
+            return indices.value
+        else:
+            raise TypeError(f'passed indices must be str or Indices objects not {type(indices)}')
```
ttung:
```suggestion
            raise TypeError(f"passed indices must be str or Indices objects, not {type(indices)}")
```

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+
+    return list(_normalize(i) for i in indices)
+
+
+def _max_intensity_table_maintain_dims(
+    intensity_table: IntensityTable, dimensions: Set[Indices],
+) -> IntensityTable:
+    """
+    Maximum project an IntensityTable over dimensions, retaining singletons in place of dimensions
```
ttung: please exercise care when you rebase this.

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+) -> np.ndarray:
+    """
+    returns a flattened boolean mask (c order flattening) that is True where
+    (feature, round, channel) combinations whose intensities are above or equal to intensity
+    threshold and where values are not nan
+    """
+    with warnings.catch_warnings():
+        # we expect to have invalid values (np.nan) in IntensityTable, hide the warnings from users
+        warnings.simplefilter('ignore', RuntimeWarning)
+        return np.ravel(intensity_table.values >= intensity_threshold)
```
ttung: so this is basically a reshape operation?  i personally find that less confusing to think about.

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+    expanded_intensities = projected_intensities.expand_dims(str_dimensions)
+    return expanded_intensities.transpose(*tuple(initial_dimensions.keys()))
+
+
+def _mask_low_intensity_spots(
+    intensity_table: IntensityTable,
+    intensity_threshold: float
+) -> np.ndarray:
+    """
+    returns a flattened boolean mask (c order flattening) that is True where
```
ttung: this flattening concept is foreign to me.  I thought it was some reduction operation whereas it is a reshaping operation.

starfish/display/stack.py
ambrosejcarr: Yes, reshaping into a 1-d vector. I'll try to clarify.

starfish/display/stack.py
```diff
@@ -0,0 +1,222 @@
...
+    coords[:, 2] = np.tile(np.tile(range(n_rounds), n_channels), n_features)
+    coords[:, 3] = np.tile(np.repeat(range(n_channels), n_rounds), n_features)
+    coords[:, 4] = np.repeat(intensity_table[Features.AXIS][Indices.Z.value].values, code_length)
+
+    sizes = np.repeat(intensity_table.radius.values, code_length)
+
+    return coords, sizes
+
+
+def stack(
```
shanaxel42: maybe we should explicitly call this napari_stack, I think we use stack in a lot of other places synonymous with imagestack so this was confusing to me at first

ambrosejcarr: Well, we _are_ showing an ImageStack here? I can call this `display_stack` if you'd like. Are you OK with the import pattern of `starfish.display.stack`? Are there alternatives you'd prefer? I'm not wed to this. :-)

ttung: When I first read this, I thought a super-cool way to do this would be a starfish.display(..) method that takes a variety of objects and just knows how to display them.  But the other display methods seem to require additional parameters, so it's less clear how that strategy would work.

shanaxel42: Ah yea starfish.display.dispay_stack would be weird. I think I misunderstood the organization, we're you thinking starfish.display.stack displays an imagestack and starfish.display.whatever would display something else? If so I'm on board with that

PR #936, opened 2019-01-05
Clarify docs and warning messages for set slice

ambrosejcarr:
I had to rewrite some code relating to these methods. The below changes make things clearer to me. 

Given that not everyone follows (Y, X) ordering, I am tempted to suggest that users should actually fully specify the axes to avoid weird transpositions where they have square tiles and submit (X, Y) accidentally (make X and Y `axes` required).


starfish/imagestack/imagestack.py
```diff
@@ -434,8 +434,8 @@ def set_slice(
         data : np.ndarray
             The source data for the operation
         axes : Optional[Sequence[Indices]]
-            The order of the axes for the source data, excluding (Y, X).  If not provided, it is
-            assumed that the data is a 2D tile.
+            The order of the axes for the source data, excluding (Y, X). Optional ONLY if data
```
ttung: the wording of this seems to presume that X, Y tiles are supported.  I don't believe that is the case.  can this be further clarified that it requires Y, X tiles?

ambrosejcarr: Updated & re-requested.

PR #944, opened 2019-01-09
Increase parallelization for travis runs.

ttung:
Introduce a slow test marker.  This breaks out check-notebooks and slow-test into a parallel travis run.  If travis has sufficient compute, this allows us to complete a test for a PR in ~9 minutes instead of 15 (see https://travis-ci.com/spacetx/starfish/builds/96790705).

Small change I snuck in: all makefile targets are hyphenated instead of underscored.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,12 @@ stages:
 - test
 jobs:
   include:
-    - name: Syntax and Tests
-      script: make check_notebooks install-src fast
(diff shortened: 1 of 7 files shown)


ttung:
> only addition I can think of if you have a definition of "slow" in mind is to make use of timeout for the fast run

we can revisit if it becomes slow again. :)

PR #949, opened 2019-01-11
Add StarMAP tilefetcher

ambrosejcarr:
Created a TileFetcher for StarMAP data.

Later PRs will add this to `starfish.data` and create a vignette.


ambrosejcarr:
Well... the only thing that's a bit unreasonable is that it produces a 3700 x 17000 pixel stitched experiment comprised of only one FoV. I'm hoping that's fixed somewhere on the back-end at some point, and in the mean time I (1) won't add it to `starfish.data` (seems mean to make people download a 20Gb experiment) and `2` will play with it using crop-on-load.

PR #955, opened 2019-01-11
Update and rename Allen notebook -> 3d smFISH

ambrosejcarr:
- Renames notebook to with a more general name that better reflects its utility as an example. 
- Updates the notebook to contain the parameters that @berl and @dganguli put together.


ambrosejcarr:
Fixes broken master issue.

PR #956, opened 2019-01-12
Adding file for travis configuration (and upping cache to 10G)

### LINKED ISSUE #943: investigate if we need to upsize our cache for travis

shanaxel42:
Fixes #943


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
 env:
   global:
-    - STARFISH_VALIDATION_STRICT=true
+    - STARFISH_CONFIG=@.travis-config.json
```
ttung: I'm concerned that this is a relative path.  I would prefer that we copy it to an absolute path, and reference an absolute path.

shanaxel42:
> Nice! I'd be interested to hear any feedback regarding the configuration itself (and/or the docs).

The docs were great! The only thing I found slightly confusing is that you can configure, for example starfish validation with either STARFISH_VALIDATION_STRICT or in the separate json config file. And then what takes precedent if both are set? But otherwise it was pretty easy to set up!

PR #959, opened 2019-01-15
[3/3] fix no-spots-detected errors

ambrosejcarr:
This PR fixes errors in each SpotFinder that occur when no spots are detected, and catches warnings that are otherwise repeatedly emitted for every application of a spot-finding call (many times per `ImageStack`). 

It also adds tests that verify that each way of calling each spot detectors works in the case where the `ImageStack` contains no detectable spots, and fixes the two failing tests from the previous PR that resulted from related problems (where _some_ but not _all_ slices of the ImageStack contained no spots).

### DIFF
--- a/starfish/spots/_detector/blob.py
+++ b/starfish/spots/_detector/blob.py
@@ -108,6 +108,9 @@ class BlobDetector(SpotFinderAlgorithmBase):
             self.overlap
         )
 
+        if fitted_blobs_array.shape[0] == 0:
+            return SpotAttributes.empty(extra_fields=['intensity', 'spot_id'])
+
         # create the SpotAttributes Table
         columns = [Axes.ZPLANE.value, Axes.Y.value, Axes.X.value, Features.SPOT_RADIUS]
         fitted_blobs = pd.DataFrame(data=fitted_blobs_array, columns=columns)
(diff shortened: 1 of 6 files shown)


starfish/test/spots/detector/test_spot_detection.py
```diff
@@ -74,19 +78,30 @@ def test_spot_detection_with_reference_image(
...
+
+        return detect_spots(
+            data_stack=stack,
+            spot_finding_method=spot_detector.image_to_spots,
+            reference_image=reference_image_mp_numpy,
+            measurement_function=np.max,
+            radius_is_gyration=radius_is_gyration
+        )
+
+    intensity_table = call_detect_spots(data_stack)
```
shanaxel42: why is this refactor necessary? def call_detect_spots

ambrosejcarr: It's being called both with `data_stack` and `EMPTY_IMAGESTACK`. This route introduces a function that defines the parameters only once.

starfish/spots/_detector/blob.py
```diff
@@ -108,6 +108,9 @@ def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> SpotAtt
             self.overlap
         )
 
+        if fitted_blobs_array.shape[0] == 0:
+            return SpotAttributes.empty(extra_fields=['intensity', 'spot_id'])
```
ttung: So pandas can't create a dataframe of size 0?

ambrosejcarr: I wasn't able to find a way to do it with a non-zero number of columns.

PR #960, opened 2019-01-15
[2/3] Refactor spot finding tests, add new examples

ambrosejcarr:
This PR refactors testing data that was found in `pixel_spot_detector.py`, moving it to `test_utils.py`. In addition, it makes the produced testing data more robust by returning, in addition to an `ImageStack`, a `Codebook` that could have generated the data (enables use with `PixelSpotDecoder`) and the `max_intensity` value for the `ImageStack` (removes magic numbers from the tests). 

The data that was originally found in `pixel_spot_detector.py` was specific to in-situ sequencing datasets, which left some testing holes in our input data types. This PR replicates the above testing construct of `Codebook, ImageStack, max_intensity` for sparse codebooks (MERFISH) and codebooks with informative blanks (DARTFISH). This covers the data types we currently have access to, and passes these data types to the spot detectors. 

Not all tests pass. These are fixed in [3/3]

### DIFF
--- a/starfish/test/spots/detector/test_spot_detection.py
+++ b/starfish/test/spots/detector/test_spot_detection.py
@@ -1,6 +1,5 @@
 import numpy as np
 import pytest
-from scipy.ndimage.filters import gaussian_filter
 
 from starfish.imagestack.imagestack import ImageStack
 from starfish.spots._detector._base import SpotFinderAlgorithmBase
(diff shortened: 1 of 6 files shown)


starfish/test/test_utils.py
```diff
@@ -70,3 +72,171 @@ def codebook_array_factory() -> Codebook:
...
+        imagestack_data[:, :, z, y, x] = codebook[f].transpose(Axes.ROUND.value, Axes.CH.value)
+
+    # blur with a small non-isotropic kernel TODO make kernel smaller.
+    imagestack_data = gaussian_filter(imagestack_data, sigma=(0, 0, 0.7, 1.5, 1.5))
+    return ImageStack.from_numpy_array(imagestack_data)
+
+
+def two_spot_one_hot_coded_data_factory() -> Tuple[Codebook, ImageStack, float]:
+    """
+    Produce a 2-channel 2-round Codebook with two codes and an ImageStack containing one spot from
```
shanaxel42: what does the word hot refer to here?

starfish/test/spots/detector/test_spot_detection.py
```diff
@@ -41,69 +49,31 @@ def simple_local_max_spot_detector() -> LocalMaxPeakFinder:
...
-@pytest.mark.parametrize('data_stack, spot_detector, radius_is_gyration', [
-    (data_stack, gaussian_spot_detector, False),
-    (data_stack, trackpy_local_max_spot_detector, True),
-    (data_stack, local_max_spot_detector, False)
-])
+# test parameterization
+test_parameters = (
+    'data_stack, spot_detector, radius_is_gyration, max_intensity',
+    [
+        (ONE_HOT_IMAGESTACK, gaussian_spot_detector, False, ONE_HOT_MAX_INTENSITY),
```
shanaxel42: It looks like the radius_is_gyration parameter only gets used in the `reference_image is not None` case, so testing every scenario for each case seems like overkill?

starfish/test/test_utils.py
ambrosejcarr: Sorry for missing this comment. One-hot is a type of codebook encoding. 

https://en.wikipedia.org/wiki/One-hot

starfish/test/spots/detector/test_spot_detection.py
ambrosejcarr: I'm going to be seriously revisiting spot detection soon (issues to come today AM). Trackpy will likely be removed, and this (good) point will be mooted.

PR #961, opened 2019-01-15
[Easy] clean up gitignore

ambrosejcarr:


ambrosejcarr:
That's a good point. I'm fine to reverse this and go that way. 👍

ambrosejcarr:
@ttung @shanaxel42 with this change you'll need to include `.idea` in your private ignore files.

PR #962, opened 2019-01-15
[Easy] Rename examples -> data_formatting_examples

ambrosejcarr:
`examples` is ambiguous. This PR: 

1. renames the directory to `data_formatting_examples`
2. renames the prefix of any data formatting example that does not fetch data from a remote store from `get_` to `format_` (e.g. `get_iss_data.py -> format_iss_data.py`).


ambrosejcarr:
@ttung I snuck in one more change after you approved, which is to move the `get_cli_test_data` into the `test` folder. Please let me know if you have any concerns.

Issue #941, opened 2019-01-05
Factor multiprocessing to enable it to be replaced by map when n_processes = 1


ambrosejcarr:
For debugging, it would be useful to be able to completely turn off multiprocessing. If we wrapped our use of the multiprocessing module, we should be able to do this across our package.

PR #964, opened 2019-01-15
Multiprocessing module for debugging

shanaxel42:
This PR:
-Creates a multiprocessing.pool wrapper class called Spool (starfish+pool, but definitely open for suggestions) that, when given n_processes=1 overrides the map function to use a non-mulitprocessed map. 
-Adds n_processes as a parameter in detect.py and pixel_spot_detector.py
-refactors imagestack.transform to use the spool class in a n_processes=1 scenario
-Modified a couple filter/detect tests to verify everything still works with spool. 

With these changes we should be able to attach the debugger to any Filter/detect step when n_processes=1.
 

fixes: #941

### DIFF
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -49,7 +49,7 @@ from starfish.imagestack.parser.crop import CropParameters, CroppedTileCollectio
 from starfish.imagestack.parser.numpy import NumpyData
 from starfish.imagestack.parser.tileset import parse_tileset
 from starfish.intensity_table.intensity_table import IntensityTable
-from starfish.multiprocessing.pool import StarfishPool
+from starfish.multiprocessing.pool import Spool
 from starfish.multiprocessing.shmem import SharedMemory
 from starfish.types import (
     Axes,
@@ -847,7 +847,7 @@ class ImageStack:
         if verbose and StarfishConfig().verbose:
             selectors_and_slice_lists = tqdm(selectors_and_slice_lists)
 
-        with StarfishPool(
+        with Spool(
                 processes=n_processes,
                 initializer=SharedMemory.initializer,
                 initargs=((self._data._backing_mp_array,
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -853,32 +853,13 @@ class ImageStack:
                 initargs=((self._data._backing_mp_array,
                            self._data._data.shape,
                            self._data._data.dtype),)) as pool:
-            if n_processes == 1:
-                sp_applyfunc: Callable = partial(
-                    self._singleprocessing_workflow,
-                    partial(func, **kwargs),
-                    self._data.data.values,
-                )
-                results = pool.map(sp_applyfunc, selectors_and_slice_lists)
-            else:
-                mp_applyfunc: Callable = partial(
-                    self._multiprocessing_workflow, partial(func, **kwargs))
-                results = pool.imap(mp_applyfunc, selectors_and_slice_lists)
+            mp_applyfunc: Callable = partial(
+                self._processing_workflow, partial(func, **kwargs))
+            results = pool.imap(mp_applyfunc, selectors_and_slice_lists)
             return list(zip(results, selectors))
 
     @staticmethod
-    def _singleprocessing_workflow(
-            worker_callable: Callable[[np.ndarray], Any],
-            numpy_array: np.ndarray,
-            selector_and_slice_list: Tuple[Mapping[Axes, int],
-                                           Tuple[Union[int, slice], ...]],
-    ):
-        sliced = numpy_array[selector_and_slice_list[1]]
-
-        return worker_callable(sliced)
-
-    @staticmethod
-    def _multiprocessing_workflow(
+    def _processing_workflow(
             worker_callable: Callable[[np.ndarray], Any],
             selector_and_slice_list: Tuple[Mapping[Axes, int],
                                            Tuple[Union[int, slice], ...]],
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -1,5 +1,4 @@
 import collections
-import multiprocessing
 import os
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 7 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -847,30 +847,24 @@ def transform(
...
-                               self._data._data.dtype),)) as pool:
-                mp_results = pool.imap(mp_applyfunc, selectors_and_slice_lists)
-                return list(zip(mp_results, selectors))
+        with Spool(
+                processes=n_processes,
+                initializer=SharedMemory.initializer,
+                initargs=((self._data._backing_mp_array,
+                           self._data._data.shape,
+                           self._data._data.dtype),)) as pool:
+            if n_processes == 1:
```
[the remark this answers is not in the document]
joshmoore: I haven't grok'd the underlying issue yet, but having run across [toolz](https://toolz.readthedocs.io/en/latest/parallelism.html following a [Rocklin blog post link](http://matthewrocklin.com/blog/work/2015/02/17/Towards-OOC-Bag), would an abstraction layer higher (one that potentially others are trying out with dask) help to determine when to use `multiprocessing`'s map versus regular `map` (versus perhaps `ipyparallel`'s map in the future)? From the config POV, it'd also be good to constrain the `num_processes==1` check to as few locations as possible so we can add an override, though if necessary that could be done on _reading_ `num_processes`.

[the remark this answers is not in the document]
ttung: `_multiprocessing_workflow` depends on `SharedMemory.initializer` being called.  In the `n_processes` > 1 case, that's handled by `mulitprocessing.Pool`.

One possible strategy is to have `Spool` mimic the logic of `multiprocessing.Pool()`s `initalizer` method when `n_processes` == 1.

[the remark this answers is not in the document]
shanaxel42: > One possible strategy is to have `Spool` mimic the logic of `multiprocessing.Pool()`s `initalizer` method when `n_processes` == 1.

This works!

starfish/imagestack/imagestack.py
```diff
@@ -50,6 +49,7 @@
 from starfish.imagestack.parser.numpy import NumpyData
 from starfish.imagestack.parser.tileset import parse_tileset
 from starfish.intensity_table.intensity_table import IntensityTable
+from starfish.multiprocessing.pool import Spool
```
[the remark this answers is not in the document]
shanaxel42:
```starfish/multiprocessing/pool.py:5: error: Name 'Pool' already defined (possibly by an import)
starfish/multiprocessing/pool.py:5: error: Cycle in inheritance hierarchy```

We'd be breaking some inheritance logic, but we could change to the delegation strategy Tony mentioned if you wanted Pool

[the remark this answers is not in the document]
Comment (ambrosejcarr) [reply starfish/imagestack/imagestack.py]: Interesting. Thanks for exploring this. I suggest digging into why @ttung suggested that to start. It might be that the right pattern is actually governed by how we import these things, and we should name the class or function properly. The thing I _really_ care about is that the API is the same as the `stdlib` `multiprocessing.pool.Pool`. 

```python
# file: starfish/multiprocessing/pool.py

class OptionalPool:
    """Multiprocessing pool that executes on the main process when n_processes=1"""
    # code ...
```

then

```python
# file: starfish/image/filter/some_filter.py

from starfish.multiprocessing.pool import OptionalPool as Pool
# code ...
```

starfish/imagestack/imagestack.py
```diff
@@ -847,44 +847,19 @@ def transform(
...
-            worker_callable: Callable[[np.ndarray], Any],
-            numpy_array: np.ndarray,
-            selector_and_slice_list: Tuple[Mapping[Axes, int],
-                                           Tuple[Union[int, slice], ...]],
-    ):
-        sliced = numpy_array[selector_and_slice_list[1]]
-
-        return worker_callable(sliced)
+                self._processing_workflow, partial(func, **kwargs))
+            results = pool.imap(mp_applyfunc, selectors_and_slice_lists)
```
ambrosejcarr: I notice we're using imap here, but we only wrap map in the multiprocessing module. Is this a problem?

starfish/multiprocessing/pool.py
```diff
@@ -0,0 +1,20 @@
...
+        mp.Pool.__init__(self, **kwargs)
+        self.n_processes = processes
+
+    def map(self, func, iterable, chunksize=None):
+        if self.n_processes == 1:
+            self._initializer(*self._initargs)
+            return map(func, iterable)
+        return mp.Pool.map(self, func, iterable, chunksize)
+
+    def __reduce__(self):
```
ambrosejcarr: Why is this needed?

starfish/imagestack/imagestack.py
shanaxel42: ah good catch! It didn't matter with the two processing methods but now it does.

starfish/multiprocessing/pool.py
shanaxel42: it's one of the abstract methods required for inheriting Pool

ttung: I'm confused.  It looks to be [implemented to prevent pool objects from being pickled](https://github.com/python/cpython/blob/master/Lib/multiprocessing/pool.py#L560).  When are we pickling?

starfish/multiprocessing/pool.py
```diff
@@ -0,0 +1,26 @@
+import multiprocessing.pool as mp
+from typing import Optional
+
+
+class Pool(mp.Pool):
+    """Wrapper class for multiprocessing pool. If n_processes=1 just use map on main
+    thread for debugging purposes """
+
+    def __init__(self, processes: Optional[int]=None, **kwargs):
```
ttung: To make this work exactly as multiprocessing.Pool, you would probably need:

`def __init__(self, processes: Optional[int]=None, *args, **kwargs):`

and then pass *args into mp.Pool's constructor.

starfish/multiprocessing/pool.py
```diff
@@ -0,0 +1,26 @@
+import multiprocessing.pool as mp
+from typing import Optional
+
+
+class Pool(mp.Pool):
+    """Wrapper class for multiprocessing pool. If n_processes=1 just use map on main
+    thread for debugging purposes """
```
ttung: main process.

it may also be worth documenting a couple of non-trivial differences in behavior between mp.Pool and this for n_processes==1:

1. Depending on the forking model, global state may not exist in mp.Pool whereas they do exist if we just map on main process..
2. initializer gets called on every invocation of mp.Pool.{imap,map}.

PR #965, opened 2019-01-17
Transfer physical Coords to Expression Matrix

### LINKED ISSUE #846: IntensityTable transfers coordinates to ExpressionMatrix

shanaxel42:
Fixes #846


starfish/intensity_table/intensity_table.py
```diff
@@ -428,7 +432,11 @@ def to_expression_matrix(self, regions: Optional[regional.many]=None) -> Express
                 "z": ("cells", np.zeros(counts.shape[0]))
```
shanaxel42: @ambrosejcarr I was a little confused about this portion, when would we supply the region info?

ambrosejcarr: The `regional` objects record cell centroids specifically. If we _don't_ know the centroid, we can guess at it by taking the extreme values of called points -- but if those don't lie at the cell boundaries, the (x, y, z) location of cells won't be properly specified. 

There's another hiccup though: this is out of date. Now we don't use regional, but store segmentation masks.

shanaxel42: But when do we get the segmentation mask info? Is it a part of SpotAttributes? Or is it something a user already has pre-pipeline?

starfish/intensity_table/intensity_table.py
```diff
@@ -428,7 +432,11 @@ def to_expression_matrix(self, regions: Optional[regional.many]=None) -> Express
...
             }
         else:
-            grouped = self.to_features_dataframe().groupby(['cell_id'])[['x', 'y', 'z']]
+            if self.has_physical_coords:
+                grouped = self.to_features_dataframe().groupby(['features'])[['x', 'y', 'z',
```
ttung: Consider not using string constants when enums for these exist.

starfish/intensity_table/intensity_table.py
```diff
@@ -173,6 +173,10 @@ def get_log(self):
         else:
             raise RuntimeError('No log info found.')
 
+    @property
+    def has_physical_coords(self):
+        return Coordinates.X and Coordinates.Y in self.coords
```
ttung: I don't think this does what you want.

```
>>> 1 and 2 in [2]
True
```

starfish/intensity_table/intensity_table.py
```diff
@@ -416,7 +420,7 @@ def to_expression_matrix(self, regions: Optional[regional.many]=None) -> Express
         """
 
         # create the 2-d counts matrix
-        grouped = self.to_features_dataframe().groupby(['cell_id', 'target'])
+        grouped = self.to_features_dataframe().groupby(['features', 'target'])
```
ttung: string constants alert.  also, it's pretty likely this wouldn't have broken if we had used the enums. :)

starfish/intensity_table/intensity_table.py
```diff
@@ -416,7 +420,7 @@ def to_expression_matrix(self, regions: Optional[regional.many]=None) -> Express
...
         counts = grouped.count().iloc[:, 0].unstack().fillna(0)
 
         if regions:
```
ttung: are there any pipelines where we know the cell_ids of the spots but don't have region (or segmentation masks) data?

starfish/test/intensity_table/test_intensity_table_coords.py
```diff
@@ -89,3 +89,39 @@ def test_tranfering_physical_coords_to_intensity_table():
...
+    intensities = intensity_table_coordinates. \
+        transfer_physical_coords_from_imagestack_to_intensity_table(stack, intensities)
+
+    expression_matrix = intensities.to_expression_matrix()
+
+    # Assert that coords were transferred
+    xc = expression_matrix.coords[Coordinates.X]
+    yc = expression_matrix.coords[Coordinates.Y]
+    zc = expression_matrix.coords[Coordinates.Z]
+    assert xc.size == NUMBER_SPOTS
```
ttung: this is puzzling to me.  this implies we have a cell for every spot?  what's the point of such a target assignment (basically none?)

shanaxel42: I thought it just implied that every spot had an xc,yc,zc coordinate?

starfish/intensity_table/intensity_table.py
ttung: ping

starfish/test/intensity_table/test_intensity_table_coords.py
ttung: expression matrix is supposed to be a cells x gene plot.

starfish/intensity_table/intensity_table.py
ambrosejcarr: I've updated the ISS vignette to include the finding of cell ids for some of the spots. That data can be used as an example to create testing data, or saved and used directly here.

starfish/intensity_table/intensity_table.py
ambrosejcarr: Apologies. we get that from the `Image.Segmentation.Watershed` pipeline component. See #1078 for an updated vignette that includes a usage example for this function.

ttung:
My guess is that this needs a rebase and maybe a tweak to the expected size of the expression matrix table, and it's ready to land.

starfish/intensity_table/intensity_table.py
```diff
@@ -419,49 +430,46 @@ def to_decoded_spots(self) -> DecodedSpots:
...
             cell x gene expression table
         """
-
-        # create the 2-d counts matrix
-        grouped = self.to_features_dataframe().groupby(['cell_id', 'target'])
+        try:
+            grouped = self.to_features_dataframe().groupby([Features.CELL_ID, Features.TARGET])
+        except KeyError as e:
+            if Features.CELL_ID in str(e):
+                raise RuntimeError(
```
ttung: `KeyError`?  or `ValueError`?

starfish/intensity_table/intensity_table.py
```diff
@@ -419,49 +430,46 @@ def to_decoded_spots(self) -> DecodedSpots:
...
         -------
         ExpressionMatrix :
             cell x gene expression table
         """
-
-        # create the 2-d counts matrix
-        grouped = self.to_features_dataframe().groupby(['cell_id', 'target'])
+        try:
+            grouped = self.to_features_dataframe().groupby([Features.CELL_ID, Features.TARGET])
+        except KeyError as e:
```
ttung: Rather than do this, why not just do:

```
if Features.CELL_ID not in self.coords:
    raise KeyError("IntensityTable must have 'cell_id' assignments for each cell before...")
```

starfish/intensity_table/intensity_table.py
ambrosejcarr: `ValueError` could work.

starfish/test/intensity_table/test_intensity_table_coords.py
```diff
@@ -68,3 +68,50 @@ def test_tranfering_physical_coords_to_intensity_table():
         z_plane = spot[Axes.ZPLANE.value].data
         physical_z = stack.xarray[Coordinates.Z.value][z_plane]
         assert np.isclose(spot[Coordinates.Z.value], physical_z)
+
+
+def test_tranfering_physical_coords_to_expression_matrix():
+    stack_shape = OrderedDict([(Axes.ROUND, 3), (Axes.CH, 2),
+                               (Axes.ZPLANE, 1), (Axes.Y, 50), (Axes.X, 40)])
+
+    physical_coords = OrderedDict([(PhysicalCoordinateTypes.X_MIN, 1),
```
ambrosejcarr: Per our conversation last week, I'd encourage testing this with negative coordinates.

starfish/test/intensity_table/test_intensity_table_coords.py
```diff
@@ -68,3 +68,50 @@ def test_tranfering_physical_coords_to_intensity_table():
...
+        num_z=stack_shape[Axes.ZPLANE],
+        height=stack_shape[Axes.Y],
+        width=stack_shape[Axes.X],
+        n_spots=NUMBER_SPOTS
+    )
+
+    intensities = intensity_table_coordinates. \
+        transfer_physical_coords_from_imagestack_to_intensity_table(stack, intensities)
+
+    # CHeck that error is thrown before target assignment
```
ambrosejcarr:
```suggestion
    # Check that error is thrown before target assignment
```

starfish/test/intensity_table/test_intensity_table_coords.py
```diff
@@ -68,3 +68,50 @@ def test_tranfering_physical_coords_to_intensity_table():
...
+    # mock out come cell_ids
+    cell_ids = random.sample(range(1, 20), NUMBER_SPOTS)
+    intensities[Features.CELL_ID] = (Features.AXIS, cell_ids)
+
+    expression_matrix = intensities.to_expression_matrix()
+    # Assert that coords were transferred
+    xc = expression_matrix.coords[Coordinates.X]
+    yc = expression_matrix.coords[Coordinates.Y]
+    zc = expression_matrix.coords[Coordinates.Z]
+    assert xc.size == len(set(cell_ids))
```
ambrosejcarr: Any chance this could test anything about the values of the coordinates?

starfish/intensity_table/intensity_table.py
```diff
@@ -419,49 +430,46 @@ def to_decoded_spots(self) -> DecodedSpots:
...
-            coordinate_df = min_ + (max_ - min_) / 2
-            metadata = {name: ("cells", data.values) for name, data in coordinate_df.items()}
-            metadata['area'] = ("cells", np.full(counts.shape[0], fill_value=np.nan))
+            grouped = self.to_features_dataframe().groupby([Features.CELL_ID])[[
+                Axes.X, Axes.Y, Axes.ZPLANE]]
+        min_ = grouped.min()
+        max_ = grouped.max()
+        coordinate_df = min_ + (max_ - min_) / 2
+        metadata = {name: (Features.CELLS, data.values) for name, data in coordinate_df.items()}
+        metadata[Features.AREA] = (Features.CELLS, np.full(counts.shape[0], fill_value=np.nan))
```
ambrosejcarr: I know you're propagating this, but it seems like it needs a `#TODO fixme` ;-)

@kne42 is this `AREA = np.nan` getting fixed in your work on segmentation?

kne42: no – will this be a problem?

Issue #761, opened 2018-11-01
`make release` target


joshmoore:
see: https://github.com/spacetx/starfish/pull/709#discussion_r226819374

Potential sequence:
 * [ ] run extended check_requirements to sync conda
 * [x] bump version number
 * [x] build `sdist`, install, and test with the installation:
   - [ ] run tests with STARFISH_STRICT_LOADING=true (see gh-764)
   - [ ] ~run `get_git_commit_hash` tests etc. (see gh-968)~ (no longer complicated)
 * [x] commit, tag, push tag
 * [x] build package and push to pypi
 * [x] build docker and push to hub
 * [ ] ~generate and push commit to bioconda-recipes (and potentially open a PR)~ (happens automatically now on tagging)
 * [ ] ~add .dev to version number, commit, and push to master~ (not needed with versioneer)

Similar checklist for comparison:
 * https://gist.github.com/audreyr/5990987

Other things to consider:
 * moving away from setuptools per https://packaging.python.org/tutorials/packaging-projects/
   - https://github.com/sdispater/poetry
   - https://github.com/ofek/hatch

ambrosejcarr:
Hey @joshmoore we've received some feedback that it has been hard for external developers to come back to starfish after a few weeks have elapsed and figure out what's happened. We're thinking about initiating to a release cycle (maybe bi-weekly?) to help mitigate this. 

Would you be interested in creating this make target? If you build it, we'll commit to using it every <TBD> weeks, and having clear changelogs :-)

joshmoore:
Investigating. I'll post some thoughts ASAP.

joshmoore:
Ok. I have something that's mostly working based on `pbr`:

 * https://docs.openstack.org/pbr/latest/user/packagers.html
 * http://www.dougalmatthews.com/2016/Sep/01/automate-publishing-to-pypi-with-pbr-and-travis/

If there are any preliminary objections to that, now would be a good time to express them. The general release process would be:

```
git tag -s -m "..." 0.0.35 HEAD  # No need to edit a file with the version
make release-prep   # Fails if no tag is present; runs quickly & installs starfish
make release-verify # Runs slowly, needs to be manually reviewed
make release-upload # Prints commands that all need to run successfully, e.g.

  # Please execute the following steps
  git push origin
  docker push spacetx/starfish:latest
  docker push spacetx/starfish:0.0.35
  twine upload dist/starfish-0.0.35.tar.gz
```

[...]

PR #968, opened 2019-01-23
Add to provenance log  

shanaxel42:
Added fields for:
-core dependencies 
-git hash
-os information

Also refactored the LogEncoder to be in the new utils.logging file.


starfish/util/logging.py
```diff
@@ -0,0 +1,40 @@
...
+    for dependency in CORE_DEPENDENCIES:
+        ps = subprocess.Popen(('pip', 'show', dependency), stdout=subprocess.PIPE)
+        version = subprocess.check_output(('grep', 'Version'), stdin=ps.stdout).strip()
+        dependency_info[dependency] = version
+    return dependency_info
+
+
+@lru_cache(maxsize=1)
+def get_git_commit_hash():
+    return subprocess.check_output(["git", "describe", "--always"]).strip()
```
joshmoore: This will fail if either or both of `starfish was installed via pip` or `user is not int the starfish directory` hold true

shanaxel42: That is an excellent point

starfish/util/logging.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+from starfish.types import CORE_DEPENDENCIES
+
+
+@lru_cache(maxsize=1)
+def get_core_dependency_info():
+    dependency_info = dict()
+    for dependency in CORE_DEPENDENCIES:
+        ps = Popen(('pip', 'show', dependency), stdout=PIPE)
```
ttung: it is highly likely that [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html) is a more portable way of determining this.

starfish/util/logging.py
```diff
@@ -0,0 +1,50 @@
...
+        version = check_output(('grep', 'Version'), stdin=ps.stdout).strip()
+        dependency_info[dependency] = version
+    return dependency_info
+
+
+@lru_cache(maxsize=1)
+def get_git_commit_hash():
+    # First check for git repo
+    if call(["git", "branch"], stderr=STDOUT, stdout=open(os.devnull, 'w')) != 0:
```
ttung: So this breaks down if someone has a different git repo that pip installs starfish.  Then we would be picking up the hash of the repo that depends on starfish.

I think what we want is to detect that starfish was installed via `pip install -e .` (and its conda equivalent), and only in that case do we log the commit hash.

starfish/util/logging.py
```diff
@@ -0,0 +1,50 @@
...
+        return check_output(["git", "describe", "--always"]).strip()
+    else:
+        return "No git repo initialized"
+
+
+@lru_cache(maxsize=1)
+def get_os_info():
+    return {"Platform": platform.system(), "Version:": platform.version()}
```
ttung: python_version is not os_info, but also worth logging.

starfish/util/logging.py
shanaxel42: wait you mean only if it wasn't installed with pip or conda then we check for the git hash right?

shanaxel42: Ah wait I think I understand what you mean. What do you think is the best way to check for that? Checking is the starfish package location doesn't include site-packages?

ttung: So here's my guess, with basically zero time put into proving its feasibility:

1. get the path of some module in starfish.  resolve it to its `.py` file (https://stackoverflow.com/questions/7162366/get-location-of-the-py-source-file might give you an idea of how to do this, but I would test it out) as opposed to the `.pyc` / `.pyo` file.
2. run `git status --porcelain -- 123 $file`.  this output will be different, depending on whether the file is in a git repo or not.  You may need to `cd` into the directory first though.

ttung: Also, I would investigate `pkg_resources` to see if you can determine how a package was installed.

shanaxel42: Having some issues with `git status --porcelain` but alternate simpler idea; what if we just check if a starfish.py is being tracked by git with `git ls-files --error-unmatch starfish.py` and if so do a git descibe

ttung: > Having some issues with git status --porcelain but alternate simpler idea; what if we just check if a starfish.py is being tracked by git with git ls-files --error-unmatch starfish.py and if so do a git descibe

What issues are you having with `git status --porcelain`?

starfish/util/logging.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+import pkg_resources
+
+import starfish
+from starfish.types import CORE_DEPENDENCIES
+
+
+@lru_cache(maxsize=1)
+def get_core_dependency_info():
```
ttung:
```suggestion
def get_core_dependency_info() -> Mapping[str, str]:
```

starfish/util/logging.py
```diff
@@ -0,0 +1,51 @@
...
+    dependency_info = dict()
+    for dependency in CORE_DEPENDENCIES:
+        version = pkg_resources.get_distribution(dependency).version
+        dependency_info[dependency] = version
+    return dependency_info
+
+
+@lru_cache(maxsize=1)
+def get_git_commit_hash():
```
ttung:
```suggestion
def get_git_commit_hash() -> str:
```

starfish/util/logging.py
```diff
@@ -0,0 +1,51 @@
...
+    # First check if in starfish repo
+    try:
+        check_output(["git", "ls-files", "--error-unmatch", starfish.__file__])
+        return check_output(["git", "describe", "--always"]).strip()
```
ttung: This should not be in the try-block.  It would seem that a failure here is indicative of a bug, rather than starfish is not under git tracking.

starfish/util/logging.py
```diff
@@ -0,0 +1,51 @@
...
+    except CalledProcessError:
+        return "Starfish project not under git tracking"
+
+
+@lru_cache(maxsize=1)
+def get_os_info():
```
ttung:
```suggestion
def get_os_info() -> Mapping[str, str]:
```

starfish/imagestack/imagestack.py
```diff
@@ -984,7 +984,12 @@ def update_log(self, class_instance) -> None:
         ----------
         class_instance: The instance of a class being applied to the imagestack
         """
-        entry = {"method": class_instance.__class__.__name__, "arguments": class_instance.__dict__}
+        entry = {"method": class_instance.__class__.__name__,
```
ttung: this is missing the starfish version.

starfish/util/logging.py
ttung: this method, as written, is subject to strange working directory effects.  i suspect you should run the commands in the parent path of `starfish.__file__`.

starfish/util/logging.py
shanaxel42: The --porcelain flag doesn't produce any output. Git status does

starfish/util/logging.py
shanaxel42: Are you talking about the `git describe` or the `git ls-files --error-unmatch`? Because the second is just list the files under tracking and error if starfish.__file__ is not one of them, isn't that the behavior we want?

starfish/util/logging.py
shanaxel42: starfish.__file__ outputs the file directory path though so does it matter?

starfish/util/logging.py
ttung: I am pretty sure that all you need to do is to execute `git ls-files os.path.basename(starfish.__file__)`, after you chdir to `os.path.dirname(starfish.__file__)`.

git ls-files of a file tracked by git, but not while you are in the directory fails:
```
[tt]:~> git ls-files microscopy/starfish-git/starfish/__init__.py
fatal: not a git repository (or any of the parent directories): .git
```

but chdir and ls-files and it works:

```
[tt]:~> cd microscopy/starfish-git/starfish/
[tt]:~/microscopy/starfish-git/starfish:tonytung-pr-684> git ls-files __init__.py
__init__.py
[tt]:~/microscopy/starfish-git/starfish:tonytung-pr-684> 
```

ttung:
I also strongly encourage you to add some tests.

starfish/util/logging.py
```diff
@@ -0,0 +1,58 @@
...
+
+@lru_cache(maxsize=1)
+def get_git_commit_hash() -> str:
+    # First check if in starfish repo
+    os.chdir(os.path.dirname(starfish.__file__))
+    try:
+        check_output(["git", "ls-files", starfish.__file__])
+    except CalledProcessError:
+        return "Starfish project not under git tracking"
+    return check_output(["git", "describe", "--always"]).strip()
```
joshmoore: @shanaxel42 : why did this get moved outside the try/except block? Won't this also throw if `ls-files` throws?

starfish/util/logging.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_dependency_version(dependency: str) -> str:
+    return pkg_resources.get_distribution(dependency).version
+
+
+@lru_cache(maxsize=1)
+def get_git_commit_hash() -> str:
+    # First check if in starfish repo
+    os.chdir(os.path.dirname(starfish.__file__))
```
joshmoore: You'll likely want a `finally` along with this `chdir` returning the user to `getcwd` otherwise file look ups will fail.

starfish/util/logging.py
shanaxel42: well if `ls-files` throws then we just return the "Starfish project not under git tracking" string. So this should only be called if `ls-files` was valid

joshmoore: 🤦‍♂️ don't mind me.

starfish/util/logging.py
ttung: +1

shanaxel42:
> > Thinking about the testing strategy here, in either docker or just via virtualenvs there should likely be three-ish different install strategies each which call `get_git_commit_hash`:
> 
> > * `python -m starfish ...`
> > * `pip install -e . && starfish ..`
> > * `python setup.py sdist && pip install dist/build/starfish... && starfish ...`
> 
> I would expand the matrix a bit further to state that logging has to work regardless of cwd.
> 
> Would recommend shoving this into new travis stages. Can run on a lower frequency (only on commits to master? or if @joshmoore builds a release process, we can have tests that only run when we attempt to cut a release?) if it takes too much time.

I don't understand the third example of install?

joshmoore:
> I wanna push back a little and say "is anyone actually asking for this git hash info to be included in the log"?

I think this matches my concern: basically getting (and keeping) this right is going to be high cost for low value.

> For most users (that installed through pip) it's going to result in a line that says "Starfish project not under git tracking" for every command, which might confuse them.

_nods_ The reverse strategy be to encode the git during build and otherwise `UNKNOWN`.

ttung:
> I wanna push back a little and say "is anyone actually asking for this git hash info to be included in the log"? For most users (that installed through pip) it's going to result in a line that says "Starfish project not under git tracking" for every command, which might confuse them.

Yes, I think @berl asked for it...?  My guess is that you are right: people should not be processing data in a permanent way with unreleased versions, and if that happens, we should just assume that all bets are off.  Recording the hash is pretty meaningless when people can be mucking around with the source files.

Proposal: record that it's being run from source and not a released version.  Probably the best way to do that is to have the release build set a magic variable.

berl:
I mentioned the git hash as a possibility for more fine-grained tracking of the codebase to augment the version number if applicable.  @ttung proposal sounds good.  Of course no one will need to ever muck about with source files in starfish anyway!

starfish/__init__.py
```diff
@@ -22,6 +22,9 @@
 # NOTE: if we move to python 3.7, we can produce this value at call time via __getattr__
 __version__ = pkg_resources.require("starfish")[0].version
 
+# Variable to be set by release process
```
ttung: I would make it super explicit, e.g., `is_released_version` or something like that.

also, cc: #761

shanaxel42:
> out of curiosity, can you dump what the log looks like now? might be good to have the community (i.e., #starfish-dev) look it over to see if it covers all the bases.

Example from ISS pipeline: 

```<class 'list'>: [
{'method': 'WhiteTophat', 
'arguments': 
    {'masking_radius': 15, 'is_volume': False}, 
    'os': {'Platform': 'Darwin', 'Version:': 'Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64', 
    'Python Version': '3.6.5'}, 
    'dependencies': {'scikit-learn': '0.20.2', 'pandas': '0.24.1', 'scikit-image': '0.14.2', 'scipy': '1.2.1', 'numpy': '1.16.1', 'sympy': '1.3', 'xarray': '0.11.3'}, 
    'release tag': 'Running starfish from source', 'starfish version': '0.0.33'},
{'method': 'FourierShiftRegistration', 
    'arguments': 
        {'upsampling': 1000, 'reference_stack': '"<starfish.ImageStack (r: 1, c: 1, z: 1, y: 140, x: 200)>"'}, 
        'os': {'Platform': 'Darwin', 'Version:': 'Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64', 'Python Version': '3.6.5'}, 
        'dependencies': {'scikit-learn': '0.20.2', 'pandas': '0.24.1', 'scikit-image': '0.14.2', 'scipy': '1.2.1', 'numpy': '1.16.1', 'sympy': '1.3', 'xarray': '0.11.3'}, 
        'release tag': 'Running starfish from source', 'starfish version': '0.0.33'}, 
{'method': 'BlobDetector', 
    'arguments': 
        {'min_sigma': 1, 
        'max_sigma': 10, 
        'num_sigma': 30, 
        'threshold': 0.01, 
        'overlap': 0.5, 
        'is_volume': True, 
        'measurement_function': '"<function mean at 0x10adbabf8>"', 
        'detector_method': '"<function blob_log at 0x12af9a048>"'}, 
        'os': {'Platform': 'Darwin', 'Version:': 'Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64', 'Python Version': '3.6.5'}, 
        'dependencies': {'scikit-learn': '0.20.2', 'pandas': '0.24.1', 'scikit-image': '0.14.2', 'scipy': '1.2.1', 'numpy': '1.16.1', 'sympy': '1.3', 'xarray': '0.11.3'}, 
        'release tag': 'Running starfish from source', 'starfish version': '0.0.33'}
        ]```

PR #972, opened 2019-01-29
Fixing build

shanaxel42:


joshmoore:
Glad this worked, @shanaxel42. Is the commenting out a result of not having napari in the conda environment? Is that the next step here?

Issue #932, opened 2019-01-04
ease pinned requirements to facilitate installation for users


neuromusic:
Currently, _starfish_ specifies it's dependencies like so:

1. we specify requirements with loose dependencies in `REQUIREMENTS.txt.in` 
2. a dev uses `make REQUIREMENTS.txt` to dump the `pip freeze` to `REQUIREMENTS.txt`, which gets committed
3. `REQUIREMENTS.txt` with hard-pinned dependencies gets scooped up into `setup.py`

This results in two things that are not ideal for users.

First, users are forced to install packages which are needed for development, but not for use of the package, such as sphinx and pytest.

Second, by including hard pinned versions in setup.py, pip will aggressively force their environment to match the pinning. This may result in users unable to install _starfish_ due to conflicts with other packages, or (if users are using conda environments) multiple versions of dependent packages installed.

To alleviate these issues, I have two proposals:

First, limit the primary "REQUIREMENTS.txt" file to those requirements needed to *use* starfish, then split off requirements for development (e.g., pytest for local testing and CI) and building documentation (sphinx & viz packages) into a separate requirements files for dev and docs: `REQUIREMENTS-dev.txt` and `REQUIREMENTS-docs.txt`. We can point travis and readthedocs at these files for building the environments for tests and docs. We can use pip's `-r` in these files to inherit the core dependencies from `REQUIREMENTS.txt`.

Second, we should carefully consider the exact packages that need strict pinning for users and limit the hard pinning to the smallest set of packages we can. For example, @ttung emphasized that numpy & scikit-image need hard pins. We should also consider keeping the list of dependencies in `REQUIREMENTS.txt` separate from those in `setup.py`, using `setup.py` to (a) specify hard pins and (b) any packages critical for starfish to work. This distinction may allow us to be more strict with the pinning in `REQUIREMENTS.txt`, as it becomes "this is a description of the environment that we test against" rather than "this is the environment that users must use".

Would love to hear thoughts on this proposal.

neuromusic:
"how do we get the docs building again in the short term" -> https://github.com/spacetx/starfish/pull/733#discussion_r229084779

neuromusic:
I think one of the key differences here as compared with Nov 9 is a separate `REQUIREMENTS-DOCS.txt` & implementing inheritance with `-r` to prevent repetition in specifying dependencies.

ambrosejcarr:
[...]

PR #981, opened 2019-02-04
ease dependencies in setup.py

neuromusic:
This PR allows starfish to be installed via pip with a looser set of dependencies, while leaving the pinned requirements.txt and other dev tooling for careful dependency control.

Note: I was able to use this method to cleanly installed napari with no version collisions.

cc @sofroniewn

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setuptools.setup(
         'semantic-version',
         'trackpy',
         'showit',
-        'slicedimage',
+        'slicedimage==1.0.3',
         'tqdm',
         'sympy',
         'jsonschema',
--- a/setup.py
+++ b/setup.py
@@ -14,10 +14,6 @@ CLASSIFIERS = [
     "Topic :: Scientific/Engineering :: Bio-Informatics",
 ]
 
-install_requires = [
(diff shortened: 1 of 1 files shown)


neuromusic:
this addresses the second part of the proposal in #932

neuromusic:
> What's the rationale for not propagating this list straight into REQUIREMENTS.txt?

Part of the goal here is to separate out the needs of users of the library (specifically: install _starfish_ in an *existing* working environment on any platform) from those of the developers (namely, a stable environment for testing and development)

berl:
@neuromusic  good idea.  Also, I feel that even for people who are strictly users, setting up an environment for starfish is not too much to ask.

neuromusic:
> setting up an environment for starfish is not too much to ask

While I agree that isolated environments for different projects is good best-practice, I wholeheartedly disagree re: expecting users to isolate starfish. To be fair, I'm coming from the assumption that starfish is a library that people use beyond the SpaceTx benchmarking effort and that they have other libraries they also want to use as well, in the same project/environment.

ttung:
Let's argue about this in #997

ambrosejcarr:
> Let's argue about this in #997

@ttung should this PR be closed then?

ttung:
closing.

Issue #945, opened 2019-01-10
Pipeline components for preprocessing of images


kevinyamauchi:
# The issue
Most microscopy images will require some amount of correction before processing to account for systematic errors (e.g.,  drift, illumination non-uniformity, channel crosstalk). Currently, the starfish expects the ingested images to already be corrected and thus the built in logging does not capture the corrects that were performed. Given that these types of corrections can have a significant effect on the downstream processing, I think it is important that these steps are performed in starfish and thus are entered into the logs.

## Example corrections
Some common corrections that I think would be useful to support include:
* Illumination non-uniformity: the illumination intensity will vary across the field of view for most imaging systems. The image must thus be corrected to take this into account so that the observed variations in intensity are due to the sample and not the illumination.
* Channel-to-channel differences in intensity: due to differences in illumination power, the properties of the fluorescent dyes, and tissue autofluorescence, the different fluorescence channels will have different intensities. It is common to normalize image intensities such that they occupy a similar range.
* Channel crosstalk: due to overlapping spectra, signal may bleed over between channels. Sometimes this is intentional to increase multiplexing. Other times this is unavoidable due to the properties of the dyes (e.g., Illumina Miseq sequencing by synthesis chemistry).
* Chromatic shift: a given point in the sample may map to different positions on the camera sensor for different wavelengths (may be due to optical properties or assembly of the imaging system). This is covered by the "Image Registration" pipeline component described in the H1 2019 Roadmap (#895 )
* Stage drift: in multiround imaging, the microscope stage may not return to the exact same spot for a given X,Y coordinate. This is covered by the "Image Registration" pipeline component described in the H1 2019 Roadmap (#895 ).

# Proposed solutions
## Proposed solution 1 (my preferred approach)
We implement two pipeline components that cover the most common corrections:
* Intensity normalization: This would cover both channel-to-channel differences in intensity and illumination non-uniformity. The component would perform element-wise multiplication of a supplied matrix (same size as a `Tile`) with specified `Indices`. For illumination correction, the user would supply the full matrix. For the intensity normalization case, the user could supply a float and the component would create the matrix: `norm_factor * np.ones((tile_shape))`.
* Channel crosstalk: this component performs linear unmixing of channels. See the BaristaSeq notebook.

## Proposed solution 2
We create general pipeline components that perform common matrix math operations to the image tensor and users can compose their own corrections. This seems like it would be less useful to the typical user and the logging may be difficult to interpret.

# Future work
[...]

PR #983, opened 2019-02-05
[RFC] Add ElementWiseMult

kevinyamauchi:
# Objective
This PR introduces a pipeline component for field flatness correction and image normalization via a user supplied correction image. 

# Overview
Per our discussion in #945, I did the following:

> For illumination correction, the user would supply the full matrix. This method uses the Filter PipelineComponent. The user supplied matrix is a 5-d tensor with any singleton dimension being broadcast. For example, to correct illumination uniformly across (x, y) user would pass (1, 1, 1, x, y). To uniformly scale channels, user would pass (1, c, 1, 1, 1) where c- is a vector of length equal to the number of channels

# Usage
```python
import numpy as np

import starfish

# Make the ImageStack
im = np.ones((1, 3, 5, 2, 2))
stack = ImageStack.from_numpy_array(im)

# Make the correction matrix
corr_mat = np.ones((1, 1, 1, 1, 3))
corr_mat[0, 0, 0, 0, 1] = 0.5
corr_mat_xr = xr.DataArray(corr_mat, dims=('r', 'x', 'z', 'y', 'c'))

# Correct...
filter_mult = starfish.image.Filter.ElementWiseMult(mult_mat=corr_mat_xr)
stack2 = filter_mult.run(stack, in_place=False, verbose=True)
```

# To do
- [x] Add tests (we can probably use variants of the usage example above)

# Questions
* I am not familiar with click. Can I supply an ndarray as a default parameter?
* For the `_DEFAULT_TESTING_PARAMETERS = {"corr_mat": 0}`: I assume these are passed as kwargs into some test. Where are those tests?
* Currently, I use `group_by` to chunk by ROUND, CH, Z (where appropriate) and numpy broadcasting for X, Y. I think this is preferred, as it is compatible with the multiprocessing. We can also just directly multiply the `mult_mat` with the `image` thanks to the magic of numpy, but I don't think this will use the multiprocessing, as implemented. Thoughts?
* Is there a better way to get the `group_by` axes (see the `run()` method)?

### DIFF
--- a/starfish/image/_filter/element_wise_mult.py
+++ b/starfish/image/_filter/element_wise_mult.py
@@ -26,8 +26,10 @@ class ElementWiseMult(FilterAlgorithmBase):
         self.mult_mat = mult_mat
 
     _DEFAULT_TESTING_PARAMETERS = {
-        "mult_mat": xr.DataArray(np.array([[[[[1]]], [[[0.5]]]]]),
-            dims=('r', 'c', 'z', 'y', 'x'))
+        "mult_mat": xr.DataArray(
+            np.array([[[[[1]]], [[[0.5]]]]]),
+            dims=('r', 'c', 'z', 'y', 'x')
+        )
     }
 
     @staticmethod
--- a/starfish/image/_filter/element_wise_mult.py
+++ b/starfish/image/_filter/element_wise_mult.py
@@ -1,19 +1,17 @@
-from functools import partial
-from typing import Optional
+from copy import deepcopy
 
 import numpy as np
 import xarray as xr
 
 from starfish.imagestack.imagestack import ImageStack
-from starfish.types import Axes
 from starfish.util import click
 from ._base import FilterAlgorithmBase
 from .util import preserve_float_range
 
 
-class ElementWiseMult(FilterAlgorithmBase):
+class ElementWiseMultiply(FilterAlgorithmBase):
 
-    def __init__(self, mult_mat: xr.core.dataarray.DataArray) -> None:
+    def __init__(self, mult_array: xr.core.dataarray.DataArray) -> None:
         """Perform elementwise multiplication on the image tensor. This is useful for
         performing operations such as image normalization or field flatness correction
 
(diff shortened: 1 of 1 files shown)


kevinyamauchi:
Ah, yes it does. I did not realize that wasn't guaranteed. Is there a way to get the order of the axes in `image` passed to `_mult()`?

ambrosejcarr:
This is great @kevinyamauchi  

> Ah, yes it does. I did not realize that wasn't guaranteed. Is there a way to get the order of the axes in `image` passed to `_mult()`?

Here's my best attempt. It gets stuck on something in xarray which I've reached out to them to ask about. 

```
import starfish
import xarray as xr
import numpy as np

# 3d array of ones
x = np.ones([10, 10, 10])
# 3d [0, 9] range over 2nd axis, singletons in 1st and 3rd axes
y = np.arange(10)[None, :, None]

# create DataArrays
xx = xr.DataArray(x, dims=('x', 'y', 'z'))
# note reversal of dimensions in y
yy = xr.DataArray(y, dims=('y', 'x', 'z'))

# if xx is our ImageStack, this provides the aligned-axes guarantee
yy.transpose(*xx.dims)

# I wish this worked, see pydata/xarray#2171
yy * xx
```

kevinyamauchi:
I was trying to use the `ImageStack.apply()` method to leverage the multiprocessing support that comes with it. As I understand it, that passes an ndarray of the selected tiles to the filter function. Is there a way to force it to pass an xarray instead?

kevinyamauchi:
@ambrosejcarr , I made the changes we discussed in our call. Would you mind taking a look? I updated the usage snippet in the PR description.

ambrosejcarr:
> I was trying to use the `ImageStack.apply()` method to leverage the multiprocessing support that comes with it. As I understand it, `ImageStack.apply()` uses ` _processing_workflow()`, which passes an ndarray of the selected tiles to the filter function. Is there a way to force it to pass an xarray instead?

#1035

ambrosejcarr:
@kevinyamauchi I ran some tests at scale to try to understand whether we need the multiprocessing.

```python
In [0]: import starfish
   ...: import os
   ...: import xarray as xr
   ...: import numpy as np
   ...: from starfish.imagestack.parser.crop import CropParameters  # noqa 
   ...: experiment = starfish.Experiment.from_json(os.path.expanduser('~/scratch/seqfish/experiment.json')) 
   ...: fov = experiment['fov_000'] 
   ...: crop_params = CropParameters(x_slice=slice(0, 1024), y_slice=slice(1024, 2048)) 
   ...: image = fov.get_image('primary', crop_params)

In [1]: mult_array = xr.DataArray(np.arange(12)[None, :, None, None, None], dims=('r', 'c', 'z', 'y', 'x'))

In [2]: ewm = starfish.image.Filter.ElementWiseMultiply(mult_array)

In [3]: %timeit -n1 -r1 ewm.run(image, n_processes=8)
38.7 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [4]: %timeit -n1 -r1 image.xarray * mult_array.values
24.2 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [5]: res_mp = ewm.run(image)

In [6]: res_direct = image.xarray * mult_array.values
```

While it does eventually get faster if you add enough processes (20), I suspect it's better to just leverage numpy's vectorization. Do you mind if I simplify the method to use a single process?

kevinyamauchi:
My intuition was that for our current tile size this would be the case. I am less sure how this works if people bring really large tiles (e.g., stitched image of a piece of tissue). What do you think about that? Is that in-scope for near-term starfish goals? I'm fine with either approach!

ambrosejcarr:
> My intuition was that for our current tile size this would be the case. I am less sure how this works if people bring really large tiles (e.g., stitched image of a piece of tissue). What do you think about that? Is that in-scope for near-term starfish goals? I'm fine with either approach!

For larger tiles, @ttung is working on a workflow runner that will break them up such that they fit in memory and send them off to individual compute nodes. I think the single-process case here makes more sense. 

I omitted the size from the test above, but it is (4, 12, 27, 1024, 1024), so a good size test for a single FoV.

Issue #985, opened 2019-02-05
FOV Refactor (aka Spiritual Bag of Images Refactor)


shanaxel42:
Description of steps for the FOV refactor. 

1.) In Experiment.from_json() when we create a FOV, FOV initialization creates set of coordinate groups
Refactor FOV object to store coordinate groups = dict of cords and the r/ch tiles associated with them 
	-Delete Imagestack from FOV constructor 
	-store coordinate groups for aux images 
	-store as Dict{tilesetName: List[CropParams]}
	-Create method to describe aligned groups to users 
	-refactor get_image() to take an aligned group number
2.) Refactor Imagestack to only accept aligned tiles
3.) Refactor Imagestack coordinates to be stored on imagestack.xarray.coords
4.) In Notebooks and cli tests, refactor pipeline to iterate through each AlignedImagestack 
5.) Create method that combines resulting IntensityTables
6.) CLI
	-new command “inspect” that shows crop param options
	-need to provide crop params from cli for Imagestack.from_path_or_url()

PR #986, opened 2019-02-05
Spiritual Bag of Images Refactor: Part 1

shanaxel42:
This PR corresponds to part 1 of the FOV refactor epic #985. Specifically it refactors the FOV object to store aligned coordinate groups for each tileset passed to the constructor. 

The only files that have substantial changes in them are:
starfish/experiment/experiment.py
starfish/imagestack/parser/crop.py
starfish/test/experiment/test_experiment.py

The PR also:
-removes the ability to create a FOV from an Imagestack (decided on in convo with @ttung)
-parses coordinate groups for the primary image tileset and auxiliary image tilesets (decided on after convo with @ambrosejcarr)
-removes __getitem__ from fov, users must now call fov.get_image() with a coordinate group defined (the default group is 0), this is the cause of most of the file changes, sorry!
-adds a method to the fov class called show_aligned_image_groups() that describes the list of coordinate groups in each tilesets in the format {tilesetname: [CordinateGroupNumber: <Imagestack shape>]}
-adds new tests to test_experiement that test initializing a FOV with aligned and unaligned tilesets


ttung:
Haven't started reviewing the code, but I ❤ your commit msg. :)

starfish/experiment/experiment.py
```diff
@@ -81,6 +79,26 @@ def __repr__(self):
...
+            x_y_coords = tuple(i for i in [
+                tile.coordinates[Coordinates.X][0], tile.coordinates[Coordinates.X][1],
+                tile.coordinates[Coordinates.Y][0], tile.coordinates[Coordinates.Y][1],
+            ])
+            if x_y_coords in coord_groups:
+                crop_params = coord_groups[x_y_coords]
+                crop_params.add_permitted_axes(Axes.CH, tile.indices[Axes.CH])
+                crop_params.add_permitted_axes(Axes.ROUND, tile.indices[Axes.ROUND])
+                if Axes.ZPLANE in tile.indices:
+                    crop_params.add_permitted_axes(Axes.ZPLANE, tile.indices[Axes.ZPLANE])
```
ambrosejcarr: Somewhere in our documentation needs to be explicitly listed that we are ONLY offering registration on `(x, y)` coordinates. @neuromusic should verify this assumption with our users -- that *FISH data does not have misalignment over the z-axis. 

I think this is correct, but we should make sure to communicate it. :-)

starfish/experiment/experiment.py
```diff
@@ -89,11 +107,69 @@ def name(self) -> str:
...
-        return ImageStack.from_tileset(self._images[item], crop_parameters)
+    def show_aligned_image_groups(self):
+        """
+        Describe the aligned subgroups for each Tileset in this FOV
+
+        ex.
+            {'primary': 'Group 0:  <starfish.ImageStack (8, 2, 1, 205, 405))>',
+             'nuclei': 'Group 0:  <starfish.ImageStack (1, 1, 1, 205, 405))>'}
+
+        Means there are two tilesets in this FOV, (primary and nuclei) each tileset only
```
ambrosejcarr: Here I would state the reason why there is only one TileSet group to make this a bit clearer: 

`Means there are two tilesets in this FOV (primary and nuclei), and because all images have the same (x, y) coordinates, each tileset has a single aligned subgroup.`

Then, below:

`Means there is one tileset in this FOV (primary), but the 4 rounds of Group 0 have different (x, y) coordinates from the rounds of Group 1, and the tiles are therefore partitioned into different coordinate groups.`

Also, with the `repr` the way it is, it's not possible to disambiguate in this example whether the cause of the misalignment is over round (n=8) or channel (n=4). Could it make sense to print out the channels and rounds associated with each Group? e.g: 

```
            {'primary': 'Group 0:  <starfish.ImageStack round=(0, 1, 2, 3), ch=(0, 1), (x, y, z)=(1, 205, 405)>
                         Group 1:  <starfish.ImageStack round=(4, 5, 6, 7), ch=(0, 1), (x, y, z)=(1, 205, 405)>'}
```

starfish/experiment/experiment.py
```diff
@@ -89,11 +107,69 @@ def name(self) -> str:
...
         return set(self._images.keys())
 
-    def get_image(self, item: str, crop_parameters: Optional[CropParameters]=None) -> ImageStack:
-        return ImageStack.from_tileset(self._images[item], crop_parameters)
+    def show_aligned_image_groups(self):
+        """
+        Describe the aligned subgroups for each Tileset in this FOV
+
+        ex.
+            {'primary': 'Group 0:  <starfish.ImageStack (8, 2, 1, 205, 405))>',
```
ambrosejcarr: There is an extra closed brace on this and all such lines. 
```suggestion
            {'primary': 'Group 0:  <starfish.ImageStack (8, 2, 1, 205, 405)>',
```

starfish/experiment/experiment.py
```diff
@@ -89,11 +107,69 @@ def name(self) -> str:
...
+
+    def iterate_image_type(self, image_type: str):
+        for aligned_group, _ in enumerate(self.aligned_coordinate_groups[image_type]):
+            yield self.get_image(item=image_type, aligned_group=aligned_group)
+
+    def get_image(self, item, aligned_group: int = 0,
+                  x_slice: Optional[Union[int, slice]] = None,
+                  y_slice: Optional[Union[int, slice]] = None,
+                  ) -> ImageStack:
+        """
```
ambrosejcarr: I'm confused about this. I thought that crop_params were previously used to crop in pixel space, but my understanding here is that the `aligned_coordinate_groups` are in coordinate space. 

In any case, would like very robust description here with examples -- this function will be called by users.

starfish/imagestack/parser/crop.py
```diff
@@ -43,6 +43,14 @@ def __init__(
         self._x_slice = x_slice
         self._y_slice = y_slice
 
+    def add_permitted_axes(self, axis_type: Axes, permitted_axis: int) -> None:
```
ambrosejcarr: Am I misunderstanding this function? It seems like this would silently fail if any axis such as `Axes.CH` evaluates to `None`.

I wonder if this could be simplified by adding an assert statement somewhere upstream in the code about an inconsistent state, similar to what Tony's done elsewhere, and make this fail explicitly.

starfish/test/experiment/test_experiment.py
```diff
@@ -1,17 +1,90 @@
...
+                        Axes.CH: ch,
+                        Axes.ZPLANE: z,
+                    }
+                )
+                tile.numpy_array = np.zeros((100, 100))
+                alignedTileset.add_tile(tile)
+    return alignedTileset
+
+
+def get_un_aligned_tilset():
```
ambrosejcarr:
```suggestion
def get_un_aligned_tileset():
```

starfish/test/experiment/test_experiment.py
```diff
@@ -1,17 +1,90 @@
...
+WIDTH = 10
+
+
+def get_aligned_tilset():
+    alignedTileset = TileSet(
+        [Axes.X, Axes.Y, Axes.CH, Axes.ZPLANE, Axes.ROUND],
+        {Axes.CH: NUM_CH, Axes.ROUND: NUM_ROUND, Axes.ZPLANE: NUM_Z},
+        (HEIGHT, WIDTH))
+
+    for r in range(NUM_ROUND):
```
ambrosejcarr: 😆 It would take me 5 minutes to figure out the way that this isn't aligned by following what `round_to_z` is doing. Could you add a comment? 

Also I thought, based on above, that we're only testing alignment on (x, y) -- but here it looks like `(x, y)` are all the same and we're testing alignment over z? I'm very confused!

shanaxel42: we're testing that z is set to the midpoint of each zrange, like discussed

starfish/experiment/experiment.py
```diff
@@ -56,17 +57,14 @@ def __init__(
...
-            self._images = images
-            if image_tilesets is not None:
-                raise ValueError(
-                    "Only one of (images, image_tilesets) should be set.")
-        elif image_tilesets is not None:
+        self.aligned_coordinate_groups: Dict[str, List[CropParameters]] = dict()
+        if image_tilesets:
+            for name, tileset in image_tilesets.items():
+                self.aligned_coordinate_groups[name] = self.parse_coordinate_groups(
+                    image_tilesets[name])
```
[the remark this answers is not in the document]
shanaxel42: good catch

starfish/experiment/experiment.py
shanaxel42: aligned_coordinate_groups are just crop params that describe what (permitted) r/ch/z are in each aligned group. If a user wants to grab an aligend group and also crop on x/y pixel space, they'd pass in these slices.

starfish/imagestack/parser/crop.py
shanaxel42: You actually have to do the check here of the linter gets mad with `error: Item "None" of "Optional[Set[int]]" has no attribute "add"`

starfish/test/experiment/test_experiment.py
shanaxel42: They should all vary because they all get multiplied based on the round value. But I added a comment!

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
...
+                for k, v in enumerate(groups)
+            )
+            all_groups[name] = f'{info}'
+        return all_groups
+
+    def iterate_image_type(self, image_type: str):
+        for aligned_group, _ in enumerate(self.aligned_coordinate_groups[image_type]):
+            yield self.get_image(item=image_type, aligned_group=aligned_group)
+
+    def get_image(self, item, aligned_group: int = 0,
```
ttung:
```suggestion
    def get_image(self, image_type: str, aligned_group: int = 0,
```

starfish/experiment/experiment.py
```diff
@@ -41,10 +46,8 @@ class FieldOfView:
     PRIMARY_IMAGES = 'primary'
 
     def __init__(
-            self,
-            name: str,
-            images: Optional[MutableMapping[str, ImageStack]]=None,
-            image_tilesets: Optional[MutableMapping[str, TileSet]]=None,
+            self, name: str,
+            image_tilesets: Optional[MutableMapping[str, TileSet]] = None
```
ttung: Since we no longer accept ImageStacks _or_ TileSets, the `image_tilesets` parameter should be required.

```suggestion
            image_tilesets: MutableMapping[str, TileSet],
```

starfish/experiment/experiment.py
```diff
@@ -56,17 +59,13 @@ def __init__(
...
         self._name = name
-        self._images: MutableMapping[str, TileSet]
-        if images is not None:
-            self._images = images
-            if image_tilesets is not None:
-                raise ValueError(
-                    "Only one of (images, image_tilesets) should be set.")
-        elif image_tilesets is not None:
+        self.aligned_coordinate_groups: Dict[str, List[CropParameters]] = dict()
+        if image_tilesets:
```
ttung: `image_tilesets` must be set.  we should remove this if-statement.

starfish/experiment/experiment.py
```diff
@@ -81,6 +79,26 @@ def __repr__(self):
             f"    {images}"
         )
 
+    def parse_coordinate_groups(self, tileset: TileSet) -> List[CropParameters]:
+        coord_groups: Dict[tuple, CropParameters] = dict()
+        for tile in tileset._tiles:
```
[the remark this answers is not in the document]
ttung:
```suggestion
        for tile in tileset.tiles():
```

starfish/experiment/experiment.py
```diff
@@ -81,6 +80,36 @@ def __repr__(self):
...
+         create aligned coordinate groups (groups of tiles that have the same physical coordinates)
+
+         Returns
+         -------
+         A list of CropParameters. Each entry describes the r/ch/z values of tiles that are aligned
+         (have matching coordinates)
+         """
+        coord_groups: Dict[tuple, CropParameters] = dict()
+        for tile in tileset._tiles:
+            x_y_coords = tuple(i for i in [
```
ttung: I personally think it's more clear just to say:

```
            x_y_coords = (
                tile.coordinates[Coordinates.X][0], tile.coordinates[Coordinates.X][1],
                tile.coordinates[Coordinates.Y][0], tile.coordinates[Coordinates.Y][1],
            )
```

starfish/imagestack/parser/crop.py
ttung: I am kind of on the fence whether this method should flag an error if the `_permitted_{axis}` is `None` to begin with.  That almost suggests user error.

I would also consider making `add_permitted_axes` a "private" method (a.k.a., `_add_permitted_axes`).

starfish/experiment/experiment.py
```diff
@@ -81,6 +80,36 @@ def __repr__(self):
...
+                crop_params.add_permitted_axes(Axes.ROUND, tile.indices[Axes.ROUND])
+                if Axes.ZPLANE in tile.indices:
+                    crop_params.add_permitted_axes(Axes.ZPLANE, tile.indices[Axes.ZPLANE])
+            else:
+                coord_groups[x_y_coords] = CropParameters(
+                    permitted_chs=[tile.indices[Axes.CH]],
+                    permitted_rounds=[tile.indices[Axes.ROUND]],
+                    permitted_zplanes=[tile.indices[Axes.ZPLANE]] if Axes.ZPLANE in tile.indices
+                    else None)
+        return list(coord_groups.values())
```
ttung: In python 3.6, dictionary iteration yielding the same order as insertion is an implementation detail, and not mandatory.  Therefore, multiple runs of this code can theoretically yield different orders.  I would consider sorting the keys, and using that to return a list of CropParameters.

In python 3.7, that has become a language level edict, so it can be relied on.  Unfortunately, we are on python 3.6.  But you can drop a comment in here saying that the complexity can be removed when we move to 3.7. :)

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
     def image_types(self) -> Set[str]:
         return set(self._images.keys())
 
-    def get_image(self, item: str, crop_parameters: Optional[CropParameters]=None) -> ImageStack:
-        return ImageStack.from_tileset(self._images[item], crop_parameters)
+    def show_aligned_image_groups(self):
```
ttung: Would not call this `show_*` as it doesn't actually show anything.

also:
```suggestion
    def show_aligned_image_groups(self) -> dict:
```

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
...
-    def __getitem__(self, item: str) -> ImageStack:
-        return self.get_image(item)
+       Means there are two tilesets in this FOV (primary and nuclei), and because all images have
+       the same (x, y) coordinates, each tileset has a single aligned subgroup.
+
+        ex.
+            {'primary': 'Group 0:  <starfish.ImageStack (4, 2, 1, 205, 405)>
+                         Group 1:  <starfish.ImageStack (4, 2, 1, 205, 405)>'}
+
+        Means there is one tileset in this FOV, (primary), it has two different aligned
```
ttung: I hypothesize that users may be interested in the physical coordinates of each group.

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
...
+
+        """
+        all_groups = dict()
+        for name, groups in self.aligned_coordinate_groups.items():
+            y_size = self._images[name].default_tile_shape[0]
+            x_size = self._images[name].default_tile_shape[1]
+            info = '\n'.join(
+                f" Group {k}: "
+                f" <starfish.ImageStack "
+                f"r={v._permitted_rounds if v._permitted_rounds else 1}, "
```
ttung: This does not look like the example text above.

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
...
+                f"ch={v._permitted_chs if v._permitted_chs else 1}, "
+                f"z={v._permitted_zplanes if v._permitted_zplanes else 1}, "
+                f"(y, x)={y_size, x_size}>"
+                for k, v in enumerate(groups)
+            )
+            all_groups[name] = f'{info}'
+        return all_groups
+
+    def iterate_image_type(self, image_type: str):
```
ttung:
```suggestion
    def iterate_image_type(self, image_type: str) -> Iterator[ImageStack]:
```

ttung: This has potential memory usage concerns.  I haven't read the subsequent diffs, but it is potentially problematic.

starfish/experiment/experiment.py
```diff
@@ -89,11 +118,70 @@ def name(self) -> str:
...
+        x_slice: int or slice
+            The cropping parameters for the x axis
+        y_slice:
+            The cropping parameters for the y axis
+
+        Returns
+        -------
+        The instantiated ImageStack
+        """
+        crop_params = self.aligned_coordinate_groups[item][aligned_group]
```
ttung: You'll want to make a copy of the CropParameters before you start modifying it.

starfish/test/experiment/test_experiment.py
```diff
@@ -1,17 +1,93 @@
...
+                        Axes.CH: ch,
+                        Axes.ZPLANE: z,
+                    }
+                )
+                tile.numpy_array = np.zeros((100, 100))
+                alignedTileset.add_tile(tile)
+    return alignedTileset
+
+
+def get_un_aligned_tileset():
```
ttung: do we want to test the case where we have multiple tiles that are aligned, yet not all tiles are?

starfish/experiment/experiment.py
shanaxel42: Changed to use a OrderedDict

starfish/test/experiment/test_experiment.py
shanaxel42: added!

starfish/experiment/experiment.py
```diff
@@ -89,11 +121,63 @@ def name(self) -> str:
...
         return set(self._images.keys())
 
-    def get_image(self, item: str, crop_parameters: Optional[CropParameters]=None) -> ImageStack:
-        return ImageStack.from_tileset(self._images[item], crop_parameters)
+    def show_aligned_image_groups(self) -> None:
+        """
+        Describe the aligned subgroups for each Tileset in this FOV
+
+        ex.
+        {'nuclei': ' Group 0:  <starfish.ImageStack r={0}, ch={0}, z={0}, (y, x)=(190,270)>',
```
ttung: Did we decide on no coordinates?  I can't remember. :(

starfish/experiment/experiment.py
ttung: There's probably a compelling reason to remove the whole crop-on-load when we move to dask.

starfish/spots/_detector/blob.py
```diff
@@ -181,7 +181,7 @@ def run(
     )
     @click.pass_context
     def _cli(ctx, min_sigma, max_sigma, num_sigma, threshold, overlap, show, detector_method):
-            instance = BlobDetector(min_sigma, max_sigma, num_sigma, threshold, overlap,
-                                    detector_method=detector_method)
-            #  FIXME: measurement_type, is_volume missing as options; show missing as ctor args
-            ctx.obj["component"]._cli_run(ctx, instance)
+        instance = BlobDetector(min_sigma, max_sigma, num_sigma, threshold, overlap,
+                                detector_method=detector_method)
+        #  FIXME: measurement_type, is_volume missing as options; show missing as ctor args
```
ttung: I don't get this comment.

shanaxel42: I think this is an old one. I didn't write it

starfish/experiment/experiment.py
shanaxel42: yea no coords since we're moving towards making the cli less granular

shanaxel42:
> I'm not going to hold up this PR over this, but I have reservations about exposing the `get_image` API as is. It doesn't appear that it changes by #1038. I suspect we should be returning a list of AlignedImageStacks. Feel free to open up an issue to track it.

I modeled the way get_image() works based off the way experiement.fov() works where, if you provide no params, it just returns the first fov. I think that's the least intrusive way to change things, but definitely down to refactor more if we decide that's the right move

starfish/experiment/experiment.py
ambrosejcarr: Is this really `get_aligned_group` now, and not `get_image`?

PR #987, opened 2019-02-06
[RFC] Add linear unmixing

kevinyamauchi:
# Objective
This PR introduces a component to perform linear unmixing across channels to compensate for crosstalk between filters.

# Overview
Per our discussion in #945, I implemented the following:

> 2\. Channel cross-talk: takes a square channel x channel coefficient matrix. This uses the Filter PipelineComponent.

# Usage
```python
import numpy as np

import starfish

# Create the ImageStack
im = np.ones((1, 3, 5, 2, 2))
stack = ImageStack.from_numpy_array(im)

# Create the coefficient matrix
coeff_mat = np.array([[1, 0, 0], [-0.25, 1, -0.25], [0, 0, 1]])

# Unmix
filter_unmix = starfish.image.Filter.LinearUnmixing(coeff_mat=coeff_mat)
stack2 = filter_unmix.run(stack, in_place=False, verbose=True)
```

# Questions
* I think it would be nice to expose a couple of rescaling options. Perhaps normalizing to the max and/or mean?
* This assumes the unmixing will be performed across channels. Is there value in making it general? At the moment, I can't think of any use cases, so I'm leaning towards keeping it specific to channels.

### DIFF
--- a/starfish/image/_filter/linear_unmixing.py
+++ b/starfish/image/_filter/linear_unmixing.py
@@ -29,7 +29,7 @@ class LinearUnmixing(FilterAlgorithmBase):
 
     @staticmethod
     def _unmix(image: np.ndarray, coeff_mat: np.ndarray) -> np.ndarray:
-        """Clip values of img below and above percentiles p_min and p_max
+        """Perform linear unmixing of channels
 
         Parameters
         ----------
--- a/starfish/image/_filter/linear_unmixing.py
+++ b/starfish/image/_filter/linear_unmixing.py
@@ -39,18 +39,18 @@ class LinearUnmixing(FilterAlgorithmBase):
         coeff_mat : np.ndarray
             matrix of the linear unmixing coefficients. Should take the form:
             B = AX, where B are the unmixed values, A is coeff_mat and X are
-            the observed values.
+            the observed values. coeff_mat has shape (n_ch, n_ch).
 
         Returns
         -------
         np.ndarray :
-          Numpy array of same shape as img
+          Numpy array of same shape as image
 
         """
         # Preallocate the new image in the same shape
         unmixed_image = np.zeros(image.shape)
 
-        # Due to the grouping, assumes image is shape [chan, x, y]
+        # Due to the grouping in run(), image is shape [chan, x, y]
         n_channels = image.shape[0]
 
         for c in range(n_channels):
--- a/starfish/image/_filter/linear_unmixing.py
+++ b/starfish/image/_filter/linear_unmixing.py
@@ -1,0 +1,109 @@
+from functools import partial
+from typing import Optional
+
+import numpy as np
+
+from starfish.imagestack.imagestack import ImageStack
+from starfish.util import click
+from starfish.types import Axes
+from ._base import FilterAlgorithmBase
+from .util import preserve_float_range
+
+
+class LinearUnmixing(FilterAlgorithmBase):
+
+    def __init__(self, coeff_mat: np.ndarray) -> None:
+        """Image scaling filter
+
+        Parameters
+        ----------
+        coeff_mat : np.ndarray
+            matrix of the linear unmixing coefficients. Should take the form:
+            B = AX, where B are the unmixed values, A is coeff_mat and X are
+            the observed values.
+
+        """
+        self.coeff_mat = coeff_mat
+
+    _DEFAULT_TESTING_PARAMETERS = {"coeff_mat": np.array([[1, -0.25], [-0.25, 1]])}
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


kevinyamauchi:
In terms of checking the functionality, I think the following test could be enough to get us started. I haven't used pytest. Is using an `assert` statements sufficient or are there some pytest-specific functions we should use? Other thoughts?

```python
import numpy as np

import starfish

# Create image
im = np.ones((2, 3, 5, 2, 2))
stack = ImageStack.from_numpy_array(im)

# Create reference result
ref_result = np.ones((2, 3, 5, 2, 2))
ref_result[:, 1, ...] = 0.5 * np.ones((5, 2, 2))

# Unmix
coeff_mat = np.array([[1, 0, 0], [-0.25, 1, -0.25], [0, 0, 1]])
filter_unmix = starfish.image.Filter.LinearUnmixing(coeff_mat=coeff_mat)
stack2 = filter_unmix.run(stack, in_place=False, verbose=False)

# Check result
assert np.all(ref_result == stack2.xarray.values)
```

starfish/image/_filter/linear_unmixing.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Parameters
+        ----------
+        image : np.ndarray
+            image to be scaled
+
+        coeff_mat : np.ndarray
+            matrix of the linear unmixing coefficients. Should take the form:
+            B = AX, where B are the unmixed values, A is coeff_mat and X are
+            the observed values.
```
ambrosejcarr: This is the case, right? 
```suggestion
            the observed values. coeff_mat has shape (n_ch, n_ch).
```

starfish/image/_filter/linear_unmixing.py
```diff
@@ -0,0 +1,107 @@
...
+
+    _DEFAULT_TESTING_PARAMETERS = {"coeff_mat": np.array([[1, -0.25], [-0.25, 1]])}
+
+    @staticmethod
+    def _unmix(image: np.ndarray, coeff_mat: np.ndarray) -> np.ndarray:
+        """Clip values of img below and above percentiles p_min and p_max
+
+        Parameters
+        ----------
+        image : np.ndarray
```
ambrosejcarr: I think this would be great if it took an `xarray` with labeled axes. Then later you could make this change:

```diff
-        n_channels = image.shape[0]
-        for c in range(n_channels):
+        for c in image.coords[Axes.CH]:
```

starfish/image/_filter/linear_unmixing.py
```diff
@@ -0,0 +1,107 @@
...
+          Numpy array of same shape as img
+
+        """
+        # Preallocate the new image in the same shape
+        unmixed_image = np.zeros(image.shape)
+
+        # Due to the grouping, assumes image is shape [chan, x, y]
+        n_channels = image.shape[0]
+
+        for c in range(n_channels):
```
ambrosejcarr: The framing of `B = AX` makes me think there's a one-liner that captures the logic in this for loop. Does that seem crazy? I'd be happy to talk through this on a call. 👍

ambrosejcarr:
> In terms of checking the functionality, I think the following test could be enough to get us started. I haven't used pytest. Is using an `assert` statements sufficient or are there some pytest-specific functions we should use? Other thoughts?

`pytest` just uses assert. You're doing the right thing 👍

starfish/image/_filter/linear_unmixing.py
kevinyamauchi: I agree that it would be nice to use the axis labels for safety, but the `apply()` passes a numpy array. I propose that we leave it as is for now and make and issue proposing that we change the `apply()` method to pass an xarray. Thoughts?

starfish/image/_filter/linear_unmixing.py
kevinyamauchi: My intuition was the same. However, I couldn't figure out how to do it without the for loop. Within the for loop (i.e., for a given channel), we could do something like:
```python
np.dot(np.transpose(image[c, :, :]),  coeff_mat[c])
```
but originally, I thought my previous implementation was more clear. I'm definitely not a broadcasting wizard, so if you can figure out how to reshape the matrices to do it all in one line, I'd love to learn.

kevinyamauchi:
@ambrosejcarr I've added a test and addressed your comments. Would you mind taking a look?

starfish/image/_filter/linear_unmixing.py
ambrosejcarr: Fixed in #1035

ambrosejcarr:
Closed by #1045

PR #988, opened 2019-02-06
Config modifications

### LINKED ISSUE #909: Warn on unknown config keys
If a config file is well-formed json but the key is misspelled, a warning should be issued.

joshmoore:
* Primarily fix #909 - warn if config is unknown
 * Add underlying support to make that easier
 * And update a few doc strings (cc: @shanaxel42)

Test plan:
 * put some unknown key in your starfish config
 * any other action should print the warning

```
$ env STARFISH_CONFIG='{"garbage":1}' starfish validate experiment demo/experiment.json
...snip art...
/scratch/repos/starfish/starfish/config/__init__.py:132: UserWarning: unknown configuration: {'garbage': 1}
  warnings.warn(f"unknown configuration: {self._config_obj.data}")
```


joshmoore:
31% _less_ coverage. 😱

ambrosejcarr:
> 31% _less_ coverage?! 😱

This has occurred on a few PRs. Kevin's had this too, and he only modified one file. Something is fishy. I would not worry about it.

ttung:
> This has occurred on a few PRs. Kevin's had this too, and he only modified one file. Something is fishy. I would not worry about it.

Is it because we use parallel travis stages?  If it's not useful, then we should just dump it.

ttung:
> Starfish currently uses the slicedimage library as a backend for storing large image sets.
Configuration values starting with ``SLICEDIMAGE_`` (or optionally, ``STARFISH_SLICEDIMAGE_``)
will be passed to the backend without modification.

Curious how this came to be.  Why not just mandate one?

joshmoore:
> Is it because we use parallel travis stages? If it's not useful, then we should just dump it.

I'm starting to think it's just due to a simple lint-style failure so that no tests are run. Not sure if there's something we can due to tell codecov to do nothing in that case.

> Curious how this came to be. Why not just mandate one?

`STARFISH_SLICEDIMAGE_` just seem so long that I added the simpler version. Happy to drop it.

PR #990, opened 2019-02-06
[RFC] Add affine transform

kevinyamauchi:
# Objective
This PR introduces a pipeline component to perform a precomputed affine transformation on an `ImageStack`

# Overview
Per our discussion in #945, I implemented the a registration component that wraps the `skimage.transform.AffineTransform`. The user can input a list of round, channel, zplane indices to select which tiles are transformed.

# Usage
```python

from skimage import data, io, img_as_float32
import numpy as np

import starfish

# Create an image
tile = img_as_float32(data.checkerboard())

im = np.zeros((2, 3, 5, 200, 200), dtype=np.float32)

for r in range(im.shape[0]):
    for c in range(im.shape[1]):
        for z in range(im.shape[2]):
            im[r, c, z, :, :] = tile
            
stack = starfish.imagestack.imagestack.ImageStack.from_numpy_array(im)

# Transform
translate = starfish.image.Registration.AffineTransform(channels=[0, 2], translation=(100, 100))
stack2 = translate.run(stack, in_place=False)
```

# Questions
* The skimage affine transform allows the user to either input the transformation matrix directly or input the individual parameters (e.g., translation, rotation) independently. I think this is a bit confusing because you can't input both the matrix and individual parameters because they may conflict. Thus, I think we should expose only one option. I opted for the individual parameters, because I thought that would be more intuitive for the general audience. What do you think?

### DIFF
--- a/starfish/image/_registration/affine_transform.py
+++ b/starfish/image/_registration/affine_transform.py
@@ -1,0 +1,123 @@
+from copy import deepcopy
+from typing import Optional, Tuple, Union
+
+from itertools import product
+import numpy as np
+from skimage.transform import AffineTransform, warp
+
+from starfish.image._filter.util import preserve_float_range
+from starfish.imagestack.imagestack import ImageStack
+from starfish.types import Axes
+from starfish.util import click
+from ._base import RegistrationAlgorithmBase
+
+
+class AffineTransform(RegistrationAlgorithmBase):
+    """
+    Performs an affine transform on the specified indices of an ImageStack
+
+    """
+    def __init__(
+        self, scale: Union[list, tuple]=None, rotation: float=None,
+        shear: float=None, translation: Union[list, tuple]=None,
+        rounds=None, channels=None, zplanes=None
+            ) -> None:
+
+        """Performs an affine transform on the specified indices of an ImageStack
+        Wraps skimage's AffineTransform
+        http://scikit-image.org/docs/dev/api/skimage.transform.html
+
+        Parameters
+        ----------
+        scale : list, tuple
+            images are registered to within 1 / upsample_factor of a pixel
+        rotation : float
+            Rotation angle in radians, counter-clockwise direction is positive
+        shear : float
+            Shear angle in radians, counter-clockwise direction is positive
+        translation : list, tuple
+            Number of pixels to translate the image (t_x, t_y)
+        rounds : np.ndarray
+            list of rounds to apply the transformation to. If omitted, transform
+            will be applied to all rounds.
+        channels : np.ndarray
+            list of channels to apply the transformation to. If omitted, transform
+            will be applied to all channels.
+        zplanes : np.ndarray
+            list of zplanes to apply the transformation to. If omitted, transform
+            will be applied to all zplanes.
+        """
+        self.scale = scale
+        self.rotation = rotation
+        self.shear = shear
+        self.translation = translation
+        self.rounds = rounds
+        self.channels = channels
+        self.zplanes = zplanes
+
+    def run(self, image: ImageStack, in_place: bool=False) -> Optional[ImageStack]:
+        """Register an ImageStack against a reference image.
+
+        Parameters
+        ----------
+        image : ImageStack
+            The stack to be registered
+        in_place : bool
+            If false, return a new registered stack. Else, register in-place (default False)
+
+        Returns
+        -------
+
+
+        """
+
+        if not in_place:
+            image = deepcopy(image)
+
+        # Get a list of all indices for an axes without provided indices
+        if self.rounds is None:
+            self.rounds = np.arange(image.num_rounds)
+        if self.channels is None:
+            self.channels = np.arange(image.num_chs)
+        if self.zplanes is None:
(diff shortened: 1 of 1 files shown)


starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
...
+        for r, c, z in slice_indices:
+            selector = {Axes.ROUND: r, Axes.CH: c, Axes.ZPLANE: z}
+
+            tile = image.get_slice(selector)[0]
+            transformed = warp(tile, transformation)
+            image.set_slice(selector, transformed.astype(np.float32))
+
+        if not in_place:
+            return image
+        return None
```
ambrosejcarr: This line isn't needed. 
```suggestion
```

starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
...
+        -------
+
+
+        """
+
+        if not in_place:
+            image = deepcopy(image)
+
+        # Get a list of all indices for an axes without provided indices
+        if self.rounds is None:
```
ambrosejcarr: We can't guarantee that channels, rounds, and zplanes are contiguous and zero-indexed. To account for this, use: 
```python3
if self.rounds is None: 
    self.rounds = image.coords[Axes.ROUND]
...
```

starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.channels is None:
+            self.channels = np.arange(image.num_chs)
+        if self.zplanes is None:
+            self.zplanes = np.arange(image.num_zplanes)
+
+        # Make the list of indices to iterate over
+        slice_indices = product(self.rounds, self.channels, self.zplanes)
+
+        # Get the transform
+        transformation = AT(scale=self.scale, rotation=self.rotation,
```
ambrosejcarr: I would prefer we import the full name here, to keep things explicit: 
```python
        transformation = AffineTransform(
            scale=self.scale, rotation=self.rotation, shear=self.shear,
            translation=self.translation
        )
```

starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
...
+                            shear=self.shear, translation=self.translation)
+
+        # Iterate through the selected tiles and perform the transform
+        for r, c, z in slice_indices:
+            selector = {Axes.ROUND: r, Axes.CH: c, Axes.ZPLANE: z}
+
+            tile = image.get_slice(selector)[0]
```
ambrosejcarr:
```suggestion
            tile, _ = image.get_slice(selector)
```
I prefer this pattern, as it makes it clear that get_slice produces two objects and we're not using the second. ("_" is a formalism for "unused arg")

starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
+from copy import deepcopy
+from itertools import product
+from typing import Optional, Union
+
+import numpy as np
+from skimage.transform import AffineTransform as AT
```
ambrosejcarr:
```suggestion
from skimage.transform import AffineTransform, warp

File: starfish/image/_registration/affine_transform.py [hunk 6]
```diff
@@ -0,0 +1,123 @@
...
+from skimage.transform import warp
```
ambrosejcarr:
```suggestion
```

starfish/image/_registration/affine_transform.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # Make the list of indices to iterate over
+        slice_indices = product(self.rounds, self.channels, self.zplanes)
+
+        # Get the transform
+        transformation = AT(scale=self.scale, rotation=self.rotation,
+                            shear=self.shear, translation=self.translation)
+
+        # Iterate through the selected tiles and perform the transform
+        for r, c, z in slice_indices:
```
ambrosejcarr: We could potentially refactor this to use apply. How fast are these methods? Would it be worth the speed up for the out-of-place transform?

ambrosejcarr:
Follow up concern about use patterns: I would guess that this method, as described, might require a very large number of calls. 

Would it be worth my thinking about a data structure to hold all the transforms for the individual tiles of a field of view, and restructuring this to take that object and register the `ImageStack` in a single use? This would have the advantage of interacting nicely with Shannon's work on the bag of images refactor.

ambrosejcarr:
Capturing an offline discussion with @kevinyamauchi : 

Registration is a two stage process: 
1. (not in this PR) learn a set of transformations
2. apply the transformations

A way to package up the output of (1) and input to (2) would be: 
d[selector] = affine_transform_matrix

For (2), _starfish_ could iterate over the keys of d to obtain the relevant data to be shifted. 

Open question about whether: 
1. imagestack can change shape during this operation (does this screw up anything downstream?)
2. if so, do we want to crop down to the region that contains data in all tiles after the transformation? 
3. if not, do we need to replace any data with zeros or nans to avoid unwanted transformation artifacts? (this could be misleading during decoding)

PR #991, opened 2019-02-06
Fix ImageStack param docstring

kevinyamauchi:
The param section of the ImageStack docstring still had `numpy_array`, so I updated it.


joshmoore:
Just ran into this one myself. Merging.

PR #992, opened 2019-02-07
FOV Refactor - Part 2

shanaxel42:
This PR corresponds with the second item in #985, specifically it refactors the Imagestack class to only accept aligned tiles, and to raise a value error if not.


starfish/test/image/test_imagestack_coordinates.py
```diff
@@ -15,12 +15,12 @@
 WIDTH = 10
 
 
-def round_to_x(r: int) -> Tuple[float, float]:
-    return (r + 1) * 1000, (r + 1) * 100
+def round_to_x() -> Tuple[float, float]:
```
ambrosejcarr: I still have no idea why these are called what they are. 😆 

Could we just hard-code (1000, 100) and (10, 0.1)?

Finally, potentially not related to this, but you're our physical space expert, so: why are the coordinates (high, low)? I would have expected the opposite.

starfish/test/image/test_imagestack_labeled_indices.py
```diff
@@ -31,18 +31,18 @@ def fill_value(round_: int, ch: int, z: int) -> float:
     return float((((round_idx * NUM_CH) + ch_idx) * NUM_Z) + z_idx) / (NUM_ROUND * NUM_CH * NUM_Z)
 
 
-def x_coordinates(round_: int, ch: int) -> Tuple[float, float]:
+def x_coordinates() -> Tuple[float, float]:
     """Return the expected physical x coordinate value for a given round/ch tuple.  Note that in
```
ambrosejcarr: I think we could foresee small variations across channel values in the cases where the filters aren't aligned -- BaristaSeq has this problem.

starfish/test/image/test_imagestack_coordinates.py
shanaxel42: yea we can, that's just how they were formatted before my changes and I was trying to change as little as possible

starfish/test/image/test_imagestack_labeled_indices.py
shanaxel42: ?? variations that make the imagestack unaligned?

ttung: Assuming the filter you are talking about is the filter in front of the sensor to read the different color channels: why would it matter?  Wouldn't the x-y coordinates be just read off of the microscope stage position?  Surely it's not taking the stage position and adding the relative offset from the edge of the sensor to the first pixel that can read a particular channel...?

ttung: cc: @kevinyamauchi

PR #997, opened 2019-02-08
loosen dependent package requirements

ttung:
1. New makefile directives:
install-dev: pip install from the current directory with pinned requirements and tooling for CI)
install-src: pip install from the current directory)
install-released-notebooks: pip install tooling to run notebooks against the released version of starfish)

2. New requirements files:
REQUIREMENTS-CI.txt: stuff needed just for CI.
REQUIREMENTS-STRICT.txt: hard pins.
REQUIREMENTS.txt: minimum requirements.

3. starfish/spots/_detector/blob.py
changed because I upgraded flake8.  Whoops.


ttung:
> LGTM. Maybe update `environment.yml` too?

What should this reflect?  Strict or loose?

kne42:
Seeing as how the strict pip requirements don't work with conda, it would probably be most useful to use the strict versions for the environment.

setup.py
```diff
@@ -28,7 +31,8 @@
     packages=setuptools.find_packages(),
     install_requires=install_requires,
     extras_require={
-        'napari': ['napari-gui==0.0.5.1', 'matplotlib==2.1.2']
+        'napari': ['napari-gui==0.0.5.1', 'matplotlib==2.1.2'],
```
neuromusic: with the loose dependencies, matlab does not need to be pinned here

ttung: Is that true?  I thought `napari` requires this specific version of matplotlib.  @kne42?

ttung:
> Seeing as how the strict pip requirements don't work with conda, it would probably be most useful to use the strict versions for the environment.

Wait, if strict versions doesn't work with conda, wouldn't we want to use loose versions...?

setup.py
kne42: We don't require any versions of matplotlib. I think it was some vispy thing which @shanaxel42 and @ambrosejcarr discovered. So I can't tell you for sure.

kne42:
Strict requirements _can_ work with conda. They just don't rn cuz they're all in pip which conflicts with how conda installs some things. So if you generated a conda environment with all of the right pinned dependencies from the right channels, then yeah, it should work. This only really impacts developers or those who choose to install from source tho since the only way to have strict requirements like this from a distributed form would be if we made it conda-installable.

setup.py
ambrosejcarr: @neuromusic napari 0.0.5.1 requires matplotlib<2.2.2 via their vispy dependency. To my knowledge, this relationship is required until that is fixed.

starfish/test/test_config.py
```diff
@@ -82,7 +82,7 @@ def test_starfish_config_value_default_key(monkeypatch):
 
 @mark.parametrize("name,config", (
     ("enabled", {
-        "expected": (2841272, 4e6),
+        "expected": (2658848, 4e6),
```
ambrosejcarr: What's going on here?

setup.py
ambrosejcarr: Checking my understanding: `pip3 install -e .[napari]` will install with loose requirements?

setup.py
```diff
@@ -28,7 +31,8 @@
     packages=setuptools.find_packages(),
     install_requires=install_requires,
```
ambrosejcarr: I think it's worth adding a note here about precedence between `install_requires` and `extras_require`, since in the strict case, `setuptools.setup` receives both the inputs from `REQUIREMENTS.txt` and `REQUIREMENTS-STRICT.txt`.

setup.py
neuromusic: I cannot replicate the incompatibility I have napari installed with a more recent matplotlib, no problem

```
>>> import napari_gui
>>> napari_gui.__version__
'0.0.5.1'
>>> import matplotlib
>>> matplotlib.__version__
'3.0.2'
>>> 
```

ttung:
Ha, a significant wrinkle in this: strict doesn't work.  Reason: pip is not a dependency resolver.  I keep saying this, but I forgot myself. :)

starfish/test/test_config.py
ttung: This number was empirically determined, but incorrectly includes the data from sqlite cache files.  2658848 is the actual number of bytes for tile data.

setup.py
ttung: This in fact is the reason `.[strict]` does not work. :(

setup.py
ttung: Correct.

joshmoore:
@ttung : let me know if you want me to resurrect the environment.yml changes here. Otherwise, I'll address them in a follow up PR.

ttung:
I revived what appears to be your commit as https://github.com/spacetx/starfish/pull/1001.

PR #1002, opened 2019-02-13
RTD: use new requirements files

joshmoore:
The sphinx_bootstrap_theme requirement is now in a second file.
RTD v1 only supports a single requirements file (which was
previously auto-detected). With v2 both `CI` and `STRICT` can
be referenced.

see:
 - https://github.com/spacetx/starfish/pull/997
 - https://readthedocs.org/projects/spacetx-starfish/builds/8570377/
 - https://docs.readthedocs.io/en/stable/config-file/v2.html#migrating-from-v1

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,8 +1,9 @@
+version: 2
 build:
     image: latest
 
 python:
     version: 3.6
-
-requirements_file: 
-    null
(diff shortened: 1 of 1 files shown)


joshmoore:
Closing and re-opening to pull in gh-1004

PR #1004, opened 2019-02-13
Fix exit code of `starfish validate xarray`

ttung:
ctx.exit(0) works by raising an exception, which is caught and reraised as a ctx.exit(1).

This moves the ctx.exit(0) into an else clause.


joshmoore:
Green. Getting this onto `master` so other PRs don't start failing.

PR #1006, opened 2019-02-13
[easy] Reduce the reinstalling of packages on travis runs

ttung:
We just ran `--force-reinstall --upgrade`.  We shouldn't need to again.


ttung:
> My only question is... why didn't we do this sooner? XD

probably because i introduced this bug in #997 :)

PR #1007, opened 2019-02-13
[easy] fix documentation about how to install for developers.

ttung:
install-dev covers installation of strict dependencies, `pip install -e`, and the installation of CI tools.

### DIFF
--- a/docs/source/installation/installation.rst
+++ b/docs/source/installation/installation.rst
@@ -38,8 +38,10 @@ Finally, then install starfish:
 Installation for developers
 ---------------------------
 
-Check out the code for starfish and set up a virtualenv_.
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
I think we probably need some kind of instructions about how to install make on osx if we go this route. (xcode tools).

ttung:
Do we not expect developers to be able to install make...?

ambrosejcarr:
> Do we not expect developers to be able to install make...?

Sorry, I was looking at the changes out of context. You make a good point, but maybe better safe than sorry? Could just be a link out to other instructions.

ttung:
Well, WSL users are going to be totally lost then. :)

ttung:
Updated with a link to https://developer.apple.com/library/archive/technotes/tn2339/_index.html

PR #1008, opened 2019-02-13
travis_wait for osmFISH notebook

ttung:
Test plan: when https://travis-ci.com/spacetx/starfish/builds/100845186 passes.


ttung:
Wrong fix.  #1013 is the right fix.

PR #1027, opened 2019-02-15
add histogram matching normalization filter

ambrosejcarr:
- Add histogram matching filter.

### DIFF
--- a/starfish/compat.py
+++ b/starfish/compat.py
@@ -1,7 +1,7 @@
 import skimage
 from packaging import version
 
-if version.parse(skimage.version) > version.parse("0.14.2"):
+if version.parse(skimage.__version__) > version.parse("0.14.2"):
     from skimage.transform import match_histograms
 else:
 
--- a/starfish/compat.py
+++ b/starfish/compat.py
@@ -1,0 +1,100 @@
+import skimage
+from packaging import version
+
+if version.parse(skimage.version) > version.parse("0.14.2"):
+    from skimage.transform import match_histograms
+else:
+
+    """
+    Unless otherwise specified by LICENSE.txt files in individual
+    directories, all code is
+
+    Copyright (C) 2011, the scikit-image team
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are
+    met:
+
+    1. Redistributions of source code must retain the above copyright
+        notice, this list of conditions and the following disclaimer.
+    2. Redistributions in binary form must reproduce the above copyright
+        notice, this list of conditions and the following disclaimer in
+        the documentation and/or other materials provided with the
+        distribution.
+    3. Neither the name of skimage nor the names of its contributors may be
+        used to endorse or promote products derived from this software without
+        specific prior written permission.
+
+    THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+    IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+    WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+    DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT,
+    INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+    (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+    SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+    HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+    STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+    IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    """
+
+    import numpy as np
+
+    def _match_cumulative_cdf(source, template):
+        """
+        Return modified source array so that the cumulative density function of
+        its values matches the cumulative density function of the template.
+        """
+        src_values, src_unique_indices, src_counts = np.unique(
+            source.ravel(), return_inverse=True, return_counts=True)
+        tmpl_values, tmpl_counts = np.unique(template.ravel(), return_counts=True)
+
+        # calculate normalized quantiles for each array
+        src_quantiles = np.cumsum(src_counts) / source.size
+        tmpl_quantiles = np.cumsum(tmpl_counts) / template.size
+
+        interp_a_values = np.interp(src_quantiles, tmpl_quantiles, tmpl_values)
+        return interp_a_values[src_unique_indices].reshape(source.shape)
+
+    def match_histograms(image, reference, multichannel=False):
+        """Adjust an image so that its cumulative histogram matches that of another.
+        The adjustment is applied separately for each channel.
+        Parameters
+        ----------
+        image : ndarray
+            Input image. Can be gray-scale or in color.
+        reference : ndarray
+            Image to match histogram of. Must have the same number of channels as
+            image.
+        multichannel : bool, optional
+            Apply the matching separately for each channel.
+        Returns
+        -------
+        matched : ndarray
+            Transformed input image.
+        Raises
+        ------
+        ValueError
+            Thrown when the number of channels in the input image and the reference
+            differ.
+        References
+        ----------
+        .. [1] http://paulbourke.net/miscellaneous/equalisation/
+        """
+        if image.ndim != reference.ndim:
+            raise ValueError('Image and reference must have the same number of channels.')
+
+        if multichannel:
+            if image.shape[-1] != reference.shape[-1]:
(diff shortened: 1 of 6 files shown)


ambrosejcarr:
Ugh. @joshmoore @ttung could either of you provide some advice on this travis error? My stab at installation works locally via `pip3 install -e .` but not on travis. 

[Context](https://spacetx.slack.com/archives/CAR91U354/p1550195454181300)

[Build failure log](https://travis-ci.com/spacetx/starfish/jobs/177994624)

[Relevant Docs, I think](https://python-packaging.readthedocs.io/en/latest/dependencies.html#packages-not-on-pypi)

joshmoore:
With `$ pip install --process-dependency-links  .[napari]` (note requirement for the flag from [stackoverflow](https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py)) and the following diff:

```
diff --git a/REQUIREMENTS.txt b/REQUIREMENTS.txt
index 3b060b1..1fa0057 100644
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -5,7 +5,8 @@ numpy != 1.13.0
 pandas >= 0.23.4
 regional
 semantic_version
-git+https://github.com/scikit-image/scikit-image@e9cf8b850c4c2800cc221be6f1dfff6a2a32
+Cython >= 0.23.4
+scikit-image==0.15.dev0
 scikit-learn
 scipy
 showit >= 1.1.4
diff --git a/setup.py b/setup.py
index be13137..1249f26 100644
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setuptools.setup(
     extras_require={
         'napari': ['napari-gui==0.0.5.1'],
     },
+    dependency_links=['git+https://github.com/scikit-image/scikit-image@e9cf8b850c4c2800cc221be6f1dfff6a2a32a4eb#egg=scikit-image-0.15.dev0'],
     entry_points={
         'console_scripts': [
             "starfish=starfish.starfish:starfish",
```

I can get some progress, but it may require more of a build chain on travis.

neuromusic:
This is one of the problems with populating `setup.py`'s `install_requires` from `REQUIREMENTS.txt`: setuptools doesn't support everything that `pip install -r` does (hence @joshmoore's rec).

My suggestion to unblock you: 

- don't point anything in REQUIREMENTS.txt to github, only change REQUIREMENTS-STRICT.txt

Travis will build the environment (incl your specific git tag of skimage) from REQUIREMENTS-STRICT.txt, then when setuptools comes in, the skimage req is already satisfied in that environment.

This will, however, prevent us from making a working release on pypi until the skimage feature is released, but it strikes me a bad idea to make releases with dependencies on pre-release code anyway.

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Returns
+        -------
+        ImageStack :
+            If in-place is False, return the results of filter as a new stack.  Otherwise return the
+            original stack.
+
+        """
+        group_by = determine_axes_to_group_by(is_volume=stack.shape[Axes.ZPLANE] > 1)
+        reference_image: np.ndarray = stack.xarray.sel(self.reference_selector).values
```
kevinyamauchi: Can you use `ImageStack.getSlice()` or `ImageStack.sel()`?

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,97 @@
...
+        result = stack.apply(
+            apply_function,
+            group_by=group_by, verbose=verbose, in_place=in_place, n_processes=n_processes
+        )
+        return result
+
+    @staticmethod
+    @click.command("MatchHistograms")
+    @click.option(
+        "--reference-selector", type=dict,
```
kevinyamauchi: Should this be required?

starfish/image/_filter/match_histograms.py
ambrosejcarr: It could use these, and the latter could be seamlessly changed. However, this would be the change:
```diff
-        reference_image: np.ndarray = stack.xarray.sel(self.reference_selector).values
+        reference_image: np.ndarray = stack.sel(self.reference_selector).xarray.values
```
I think this is a sign we've done something right, or really pointless. 😆 

Since I expect `xarray` to have a more stable API than `ImageStack`, I'd like to keep it the way I've written it.

ambrosejcarr:
Note to self: This needs to be updated to match all slides to the average intensity across the channels. See e.g.: 

```python
def quantile_normalize(xarray):
    stacked = xarray.stack(pixels=(Axes.X.value, Axes.Y.value, Axes.ZPLANE.value))
    inds = stacked.groupby(Axes.CH.value).apply(np.argsort)
    pos = inds.groupby(Axes.CH.value).apply(np.argsort)

    sorted_pixels = deepcopy(stacked)
    for v in sorted_pixels.coords[Axes.CH.value]:
        sorted_pixels[v, :] = sorted_pixels[v, inds[v].values].values

    rank = sorted_pixels.mean(Axes.CH.value)

    output = deepcopy(stacked)
    for v in output.coords[Axes.CH.value]:
        output[v] = rank[pos[v].values].values

    return output.unstack("pixels")

quantile_normalized = primary_image.xarray.groupby(
    Axes.ROUND.value
).apply(quantile_normalize)
quantile_normalized = starfish.ImageStack.from_numpy_array(quantile_normalized.values)
```

starfish/compat.py
```diff
@@ -0,0 +1,100 @@
+import skimage
+from packaging import version
+
+if version.parse(skimage.__version__) > version.parse("0.14.2"):
```
ttung: Yuck.  Can we just require >= 0.14.2?

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,97 @@
...
+        ----------
+        image, reference : numpy.ndarray[np.float32]
+            2-d or 3-d image data
+
+        Returns
+        -------
+        np.ndarray :
+            image, with intensities matched to reference
+        """
+        if isinstance(image, xr.DataArray):
```
ttung: `np.asarray`?

starfish/compat.py
ambrosejcarr: No, unfortunately we just missed an `skimage` release, and this is the most recent version.

ttung: oh this is the code you added to skimage.

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        self.group_by = group_by
+
+    _DEFAULT_TESTING_PARAMETERS = {"group_by": {Axes.CH, Axes.ROUND}}
+
+    @staticmethod
+    def _harmonize_enum(iterable):
+        return list(v if isinstance(v, str) else v.value for v in iterable)
+
+    def _compute_reference_distribution(self, data: ImageStack):
```
ttung:
```suggestion
    def _compute_reference_distribution(self, data: ImageStack) -> xr.DataArray:
```

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+        stacked = stacked.stack(sort_key=sort_key)
+
+        sorted_stacked = stacked.groupby("sort_key").apply(np.sort)
+        reference = sorted_stacked.mean("sort_key")
+        reference = reference.unstack("chunk_key")
+        return reference
+
+    @staticmethod
+    def _match_histograms(
+        image: Union[xr.DataArray, np.ndarray], reference: np.ndarray
```
ttung: From my reading of the code, reference should be typed `xr.DataArray`?

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+    ) -> np.ndarray:
+        """
+        matches the intensity distribution of image to reference
+
+        Parameters
+        ----------
+        image, reference : numpy.ndarray[np.float32]
```
ttung: this does not match the params above.

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+        """Perform filtering of an image stack
+
+        Parameters
+        ----------
+        stack : ImageStack
+            Stack to be filtered.
+        in_place : bool
+            if True, process ImageStack in-place, otherwise return a new stack
+        verbose : bool
+            if True, report on filtering progress (default = False)
```
ttung: this does not match the above.

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+        Parameters
+        ----------
+        group_by : Set[Axes]
+        """
+        self.group_by = group_by
+
+    _DEFAULT_TESTING_PARAMETERS = {"group_by": {Axes.CH, Axes.ROUND}}
+
+    @staticmethod
+    def _harmonize_enum(iterable):
```
ttung: general comment: this is plausibly a useful method for other places.

```suggestion
    def _harmonize_enum(iterable: Iterable[Axes]) -> Iterable[str]:
```

starfish/image/_filter/match_histograms.py
```diff
@@ -0,0 +1,124 @@
...
+        return list(v if isinstance(v, str) else v.value for v in iterable)
```
ttung:
```suggestion
        return list(v if isinstance(v, str) else str(v) for v in iterable)
```

ambrosejcarr: Oh, I didn't realize that worked. This can just be `list(str(v) for for in iterable)` then.

PR #1026, opened 2019-02-15
CLI command to use strict dependencies

ttung:
Test plan: created a test env with different versions of dependencies.  then ran `starfish install-strict-dependencies`

PR #1028, opened 2019-02-15
put install command under util

joshmoore:
see: https://github.com/spacetx/starfish/pull/1026

### DIFF
--- a/starfish/starfish.py
+++ b/starfish/starfish.py
@@ -72,8 +72,18 @@ def version():
 version.no_art = True  # type: ignore
 
 
-@starfish.command()
+@starfish.group()
+def util():
+    """
+    house-keeping commands for the starfish library
+    """
+    pass
+
+@util.command()
 def install_strict_dependencies():
+    """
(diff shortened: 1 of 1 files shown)


starfish/starfish.py
```diff
@@ -72,8 +72,18 @@ def version():
...
+def util():
+    """
+    house-keeping commands for the starfish library
+    """
+    pass
+
+@util.command()
 def install_strict_dependencies():
+    """
+    warning! updates different packages in your local installation
```
ttung: does this need to be in the help text?

joshmoore: The “warning”? Nah. But there aren’t many packages that install stuff themselves.

ttung: Or printed out to the console? Actually the help text for this command is pretty unhelpful.

joshmoore: Ha. Fair enough. :smile: But I'm misusing your sign-off to get this into your PR where we can discuss a help message better than `None` and "unhelpful"

PR #1030, opened 2019-02-15
add 3d spot display and ability to visually mask failing spots

ambrosejcarr:
Ignore installation strategy for now, which fails on Travis and is blocked by solution to #932 

- Moves napari to a core dependency
- Add n-dimensional markers which enables visualization of spots across Z. Dramatically improves the visual salience of spots, as before the spots would be visible across ~5 z-planes, but the marker would only appear at the point of highest intensity. 
- Refactor marker plotting into `_plot_markers` to support the below functionality
- Enable additional sets of spots to be passed beyond the first array. These spots can be contrast colored to enable A/B comparison of spots that successfully decode and spots that do not, or to identify autofluorescence that is decoded separately. If the zero-color for the image colormap is passed as the color for these markers, the spots will _mask_ the failing spots, making it very clear which spots have been decoded, and which are not explained under current spot-calling strategies, leading to fast iteration on choosing the correct spots.

### DIFF
--- a/REQUIREMENTS-STRICT.txt
+++ b/REQUIREMENTS-STRICT.txt
@@ -26,6 +26,7 @@ MarkupSafe==1.1.0
 matplotlib==2.1.2
 mistune==0.8.4
 mpmath==1.1.0
+git+https://github.com/napari/napari-gui.git@master
 nbconvert==5.4.1
 nbformat==4.4.0
 networkx==2.2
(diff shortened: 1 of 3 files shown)


ambrosejcarr:
@kne42 I think this PR relates to what you're going to work on. This is the most recent tip of my development on starfish visualization using napari. Given our discussions about depending on pre-release features from napari, I am going to close this PR, but I suggest you work from here on any refactors of starfish.display.stack.

PR #1031, opened 2019-02-15
add codebooks to mask autofluorescence

ambrosejcarr:
Adds functions to produce two codebooks: 

1. A codebook that will decode any spot that is present in _all rounds and channels_
2. A codebook that will decode any spot that is present in _all rounds of each specific channel_ (separately)

These codebooks identify autofluorescent spots, and can be used with #1030 to either mask these spots or visualize them with contrasting colors.

### DIFF
--- a/starfish/codebook/codebook.py
+++ b/starfish/codebook/codebook.py
@@ -735,7 +735,7 @@ class Codebook(xr.DataArray):
         data = np.ones((1, self.sizes[Axes.CH.value], self.sizes[Axes.ROUND.value]))
         dims = (Features.TARGET, Axes.CH.value, Axes.ROUND.value)
         coords = {
-            Features.TARGET: (Features.TARGET, ("autofluorescence", )),
+            Features.TARGET: (Features.TARGET, ["autofluorescence"]),
             Axes.CH.value: (Axes.CH.value, self.coords[Axes.CH.value]),
             Axes.ROUND.value: (Axes.ROUND.value, self.coords[Axes.ROUND.value])
         }
--- a/starfish/codebook/codebook.py
+++ b/starfish/codebook/codebook.py
@@ -1,6 +1,5 @@
 import json
 import uuid
-from itertools import product
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


kevinyamauchi:
Good idea! I haven't had a chance to look at the code yet, but one high level question: many autofluorescent things have fairly broad fluorescence and thus will be visible in multiple channels. Do you think there's a reasonable way to generalize without remaking `from_code_array()`?

ambrosejcarr:
> Good idea! I haven't had a chance to look at the code yet, but one high level question: many autofluorescent things have fairly broad fluorescence and thus will be visible in multiple channels. Do you think there's a reasonable way to generalize without remaking `from_code_array()`?

This made me realize there are simpler ways to create these. I pushed a new commit. when you look at the code now, is it clear how you'd create something yourself? 

You could also skip `Codebook._create_codebook` completely, since `Codebook` is a subclass of `xarray`, you could make a codebook directly from a `numpy` array. e.g. `constitutive_fluorescence_mask` could just as easily be constructed doing: 

```python
import numpy as np
from starfish import Codebook

N_CH = 5
N_ROUND = 3

code_names = ["autofluorescence"]
Codebook(
    data=np.ones((1, N_CH, N_ROUND)),
    coords=(
        ("target", ["autofluorescence"]),
        ("c", range(N_CH)),
        ("r", range(N_ROUND))
    )
)
```

ttung:
Can this be implemented as a filter?  Logically, that makes more sense to me.

ambrosejcarr:
Interesting. It'd be a filter on the `IntensityTable`, right? I could do that, and agree that would make sense.

ttung:
> Interesting. It'd be a filter on the IntensityTable, right? I could do that, and agree that would make sense.

I don't know that it has to be one vs the other.  You could definitely do it in image space.

If you _must_ do it in codebook space, I would suggest that you make a method that takes an existing codebook, and adds these code points.  The resulting codebook can be used to decode an image and you get out what you expect (throwing away the wonky points).

ambrosejcarr:
@ttung I adjusted these to be additive to existing codebooks.

ttung:
What happens if someone does this to a codebook that has a code at that position?  IMO, we should fail in that case.

ambrosejcarr:
I haven't seen much interest in using the content of this PR so I suggest closing it for the time being unless someone objects.

PR #1032, opened 2019-02-15
FOV refactor - Part 3

shanaxel42:
This PR corresponds to part 3 of the FOV refactor epic #985. Specifically it refactors physical coordinates to be stored as coords on Imagestack.xarray instead of as a separate xarray. 

It also: 
-refactors imagestack properties to be calculated from it's xarray data instead of storing them upon initialization. (This was to make the imagestack.sel() method cleaner)
-modifies transfer_physical_coords_from_imagestack_to_intensity_table() to read from Imagestack.xarray coords (NOTE: this can be refactored more, but I figured that should be a separate PR)
-Modifies a lot of tests to verify physical coords on the entire imagestack instead of per tile. 
-removes stale coordinate calculating code.


starfish/imagestack/imagestack.py
```diff
@@ -175,35 +159,39 @@ def __init__(
...
 
         all_selectors = list(self._iter_axes({Axes.ROUND, Axes.CH, Axes.ZPLANE}))
         first_selector = all_selectors[0]
         tile = tile_data.get_tile(r=first_selector[Axes.ROUND],
                                   ch=first_selector[Axes.CH],
                                   z=first_selector[Axes.ZPLANE])
-        # only compare X,Y coords
+
+        # Set up coordinates
+        self._data[Coordinates.X.value] = xr.DataArray(
```
ambrosejcarr: Cool. 👍

starfish/imagestack/imagestack.py
```diff
@@ -175,35 +159,39 @@ def __init__(
...
+            np.linspace(tile.coordinates[Coordinates.X][0], tile.coordinates[Coordinates.X][1],
+                        self.xarray.sizes[Axes.X.value]), dims=Axes.X.value)
+        self._data[Coordinates.Y.value] = xr.DataArray(
+            np.linspace(tile.coordinates[Coordinates.Y][0], tile.coordinates[Coordinates.Y][1],
+                        self.xarray.sizes[Axes.Y.value]), dims=Axes.Y.value)
+        if Coordinates.Z in tile.coordinates:
+            # Fill with zeros for now, then replace with calculated midpoints
```
ambrosejcarr: Is there no way to look up this information? would it make sense to record the extrema on the TileSet?

starfish/imagestack/imagestack.py
```diff
@@ -175,35 +159,39 @@ def __init__(
...
+            self._data[Coordinates.Z.value] = xr.DataArray(np.zeros(
+                self.xarray.sizes[Axes.ZPLANE.value]),
+                dims=Axes.ZPLANE.value)
+
+        # Get coords on first tile to verify imagestack is aligned
```
ambrosejcarr: This comment doesn't make sense to me. I imagine that _all_ tiles would need to be parsed to verify this.

starfish/imagestack/imagestack.py
```diff
@@ -212,13 +200,10 @@ def __init__(
...
             if Coordinates.Z in tile.coordinates:
-                coordinates_values.extend([
-                    tile.coordinates[Coordinates.Z][0], tile.coordinates[Coordinates.Z][1],
-                ])
-            else:
-                coordinates_values.extend([np.nan, np.nan])
-
-            self._coordinates.loc[coordinate_selector] = np.array(coordinates_values)
+                z_range = (tile.coordinates[Coordinates.Z][0], tile.coordinates[Coordinates.Z][1])
+                # As discussed just taking the middle of the z range
```
```suggestion
                # Use mid-point of the z range for a tile for the z-coordinate
```

starfish/imagestack/imagestack.py
```diff
@@ -399,19 +384,9 @@ def sel(self, indexers: Mapping[Axes, Union[int, tuple]]):
         """
 
         # convert indexers to Dict[str, (int/slice)] format
+        stack = deepcopy(self)
```
ambrosejcarr: Whoa -- this is going to make a full copy of the ImageStack to extract a subset. Are we sure this is necessary?

starfish/test/image/test_imagestack_cropped_load.py
```diff
@@ -37,22 +39,18 @@ def data(round_: int, ch: int, z: int) -> np.ndarray:
 
 
 def x_coordinates() -> Tuple[float, float]:
-    """Return the expected physical x coordinate value for a given round/ch tuple.  Note that in
-    real life, physical coordinates are not expected to vary for different ch values.  However, for
-    completeness of the tests, we are pretending they will."""
-    return 0.01, 0.01
+    """Return the expected physical x coordinates."""
```
ambrosejcarr: It feels like these could be hard-coded into the tests to make things simpler, or hard coded as constants into the tests, if the desire is to fix the values across the tests.

starfish/imagestack/imagestack.py
shanaxel42: I use the coordinates from the first tile and verify that the rest match

starfish/imagestack/imagestack.py
shanaxel42: yea you don't want to modify the existing imagestack. The old method created a brand new imagestack so I don't this is too crazy of a difference

starfish/imagestack/imagestack.py
```diff
@@ -1107,13 +1057,15 @@ def export(self,
...
+            x_coordinates = (float(min(self.xarray[Coordinates.X.value])),
+                             float(max(self.xarray[Coordinates.X.value])))
+            y_coordinates = (float(min(self.xarray[Coordinates.Y.value])),
+                             float(max(self.xarray[Coordinates.Y.value])))
             coordinates[Coordinates.X] = x_coordinates
             coordinates[Coordinates.Y] = y_coordinates
-            if z_coordinates[0] != np.nan and z_coordinates[1] != np.nan:
+            if Coordinates.Z in self.xarray.coords:
+                z_coordinates = (float(min(self.xarray[Coordinates.Z.value])),
+                                 float(max(self.xarray[Coordinates.Z.value])))
```
shanaxel42: won't work. Need to go through each zplane and set value

starfish/imagestack/imagestack.py
shanaxel42: which info are you talking about?

starfish/imagestack/imagestack.py
shanaxel42: Tried to change this to just a basic copy(self) but ran into an internal pandas error. I think after tony's changes to the image stack class land though we should be able to use xarray.copy(deep=false) and save some memory.

starfish/imagestack/imagestack.py
ambrosejcarr:
```suggestion
        # Get coords on first tile, then verify all subsequent tiles are aligned
```

starfish/imagestack/imagestack.py
ambrosejcarr:
```suggestion
        # TODO shanaxel42 check if this can be changed to xarray.copy(deep=false)
        stack = deepcopy(self)

File: starfish/intensity_table/intensity_table_coordinates.py [hunk 8]
```diff
@@ -19,6 +18,7 @@ def transfer_physical_coords_from_imagestack_to_intensity_table(image_stack: Ima
         - Calculate the physical x,y values
         - Assign those values to the coords arrays for this spot
     """
+    # TODO consider refactoring pixel case where were can just reshape from Imagestack
```
ambrosejcarr:
```suggestion
    # TODO shanaxel42 consider refactoring pixel case where were can just reshape from Imagestack
```

starfish/imagestack/imagestack.py
ttung: @ambrosejcarr

starfish/imagestack/imagestack.py
ttung: What sort of pandas error?

starfish/imagestack/imagestack.py
ambrosejcarr: The Z mid-point information.

starfish/test/image/imagestack_test_utils.py
```diff
@@ -38,20 +38,21 @@ def verify_stack_fill(
...
 def verify_physical_coordinates(
         stack: ImageStack,
-        selectors: Mapping[Axes, int],
         expected_x_coordinates: Tuple[float, float],
         expected_y_coordinates: Tuple[float, float],
         expected_z_coordinates: Tuple[float, float]
 ) -> None:
-    """Given an imagestack and a set of selectors, verify that the physical coordinates for the data
-    referred to by the selectors match the expected physical coordinates.
+    """Given an imagestack and a set coordinate min/max values
```
ttung:
```suggestion
    """Given an imagestack and a set of coordinate min/max values
```

you can also unwrap a bit, as the line length limit is like 100.

starfish/test/image/imagestack_test_utils.py
```diff
@@ -38,20 +38,21 @@ def verify_stack_fill(
 
 def verify_physical_coordinates(
         stack: ImageStack,
-        selectors: Mapping[Axes, int],
         expected_x_coordinates: Tuple[float, float],
         expected_y_coordinates: Tuple[float, float],
         expected_z_coordinates: Tuple[float, float]
```
ttung: I would argue that Z coordinates should be allowed to vary throughout the imagestack, and that we should support testing for that.

starfish/test/image/test_imagestack_coordinates.py
```diff
@@ -17,7 +18,7 @@
 
 X_COORDS = 100, 1000
 Y_COORDS = .1, 10
-
+Z_COORDS = 0.01, 0.001
```
ttung: This is unrealistic.

starfish/test/image/imagestack_test_utils.py
ambrosejcarr: What kind of non-aligned data are you thinking about supporting here? I agree that generality is important, but for the SpaceTx groups this might not be necessary (at least in the short term). The registration that I've seen applied is: 

1. register `(x, y)` tiles across `z` against a single `z` anchor/reference tile
2. register `(x, y, z)` volumes across `(r, c)` against a single `(r, c)` anchor/reference volume

starfish/test/image/test_imagestack_coordinates.py
shanaxel42: why?

PR #1035, opened 2019-02-21
Imagestack.transform executes applied function on xarray objects instead of numpy arrays

ambrosejcarr:
@kevinyamauchi pointed out in #987 and #983 that our apply method hampers our ability to write expressive `PipelineComponent`s because it returns numpy arrays instead of xarray DataArray objects. 

This PR:

1. Modifies `ImageStack.transform` so that methods used by `apply` can take advantage of `xarray` named axes. Because we are rebuilding an array from the `sharedmemory` object, this does not give us access to data stored in the coordinates. Thus, the returned arrays only guarantee that the Axes are in the correct order. If `is_volume` is True, xarrays generated by `ImageStack.transform` have dims = `(z, y, x)` and no coordinates. If `is_volume` is False, xarrays generated by `ImageStack.transform` have dims = `(y, x)` and no coordinates. These are set based on the information in `starfish.imagestack.dataorder.AXES_DATA`

2. Fixes a bug in `preserve_float_range`. That function has obviously never seen an `xarray` before. 😆 

3. Minor updates to `SpotFinder` methods to support receiving `xr.DataArray` objects

4. Updates to docstrings to reflect flexibility to receive `Union[xr.DataArray, np.ndarray]` images.

### DIFF
--- a/starfish/image/_filter/util.py
+++ b/starfish/image/_filter/util.py
@@ -133,17 +133,19 @@ def preserve_float_range(
 
     """
     array = array.copy()
+
     if isinstance(array, xr.DataArray):
         data = array.values
     else:
         data = array
+
     if np.any(data < 0):
-        data[array < 0] = 0
-    if np.any(array > 1):
+        data[data < 0] = 0
+    if np.any(data > 1):
         if rescale:
             data /= data.max()
         else:
-            data[array > 1] = 1
+            data[data > 1] = 1
     return array.astype(np.float32)
 
 
--- a/starfish/imagestack/imagestack.py
+++ b/starfish/imagestack/imagestack.py
@@ -857,12 +857,24 @@ class ImageStack:
             selector_and_slice_list: Tuple[Mapping[Axes, int],
                                            Tuple[Union[int, slice], ...]],
     ):
+        # build the numpy array from the shared memory object
         backing_mp_array, shape, dtype = SharedMemory.get_payload()
         unshaped_numpy_array = np.frombuffer(backing_mp_array.get_obj(), dtype=dtype)
-        numpy_array = unshaped_numpy_array.reshape(shape)
 
-        sliced = numpy_array[selector_and_slice_list[1]]
+        # get the correct dimension order for the imagestack based on AXES_DATA
+        dims = [
+            name.value for (name, data) in
+            sorted(AXES_DATA.items(), key=lambda kv: kv[1].order)
+        ] + [Axes.Y.value, Axes.X.value]
 
+        # build and then slice the xarray to get the piece needed for this worker
+        data_array = xr.DataArray(
+            data=unshaped_numpy_array.reshape(shape),
+            dims=dims,
+        )
+        sliced = data_array.sel(selector_and_slice_list[0])
+
+        # return the result of the function called on the slice
         return worker_callable(sliced)
 
     @property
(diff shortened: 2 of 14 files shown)


kevinyamauchi:
Thanks, @ambrosejcarr! I am not very experienced with multiprocessing, so I think I should leave the approval to @shanaxel42.

I did have a question though: in general, why do the filtering functions in all of the filters retain compatibility for `image` being passed in as an `ndarray`?

ambrosejcarr:
> I did have a question though: in general, why do the filtering functions in all of the filters retain compatibility for image being passed in as an ndarray?

Not _all_ of them did, but this is by design: xarray wraps a bumpy array and implements or delegates most of the bumpy array functionality. Pretty cool right? :-)

kevinyamauchi:
Sorry for being unclear. I was more specifically wondering if the fact that the type hint for `image` in many of the filtering methods (e.g., `Clip._clip()`)  include `ndarray` implies that there are conditions in which `ImageStack._processing_workflow()` would pass an `ndarray` instead of an `xarray`. Or is it just to indicate the compatibility as you described above?

ambrosejcarr:
> is it just to indicate the compatibility as you described above?
Just to indicate compatibility.

kevinyamauchi:
Got it. Thanks!

starfish/imagestack/imagestack.py
```diff
@@ -857,12 +857,24 @@ def _processing_workflow(
...
         backing_mp_array, shape, dtype = SharedMemory.get_payload()
         unshaped_numpy_array = np.frombuffer(backing_mp_array.get_obj(), dtype=dtype)
-        numpy_array = unshaped_numpy_array.reshape(shape)
 
-        sliced = numpy_array[selector_and_slice_list[1]]
+        # get the correct dimension order for the imagestack based on AXES_DATA
+        dims = [
+            name.value for (name, data) in
+            sorted(AXES_DATA.items(), key=lambda kv: kv[1].order)
+        ] + [Axes.Y.value, Axes.X.value]
```
shanaxel42: we're just trying to maintain the current axes order here right? why not just `dims = [dim for dim in self.xarray.dims]`

ambrosejcarr: Yes, but this is a static method so does not have access to self. Happy to simplify this if you can find an easier way!

starfish/image/_filter/util.py
```diff
@@ -133,17 +133,19 @@ def preserve_float_range(
...
+
     if isinstance(array, xr.DataArray):
         data = array.values
     else:
         data = array
+
     if np.any(data < 0):
-        data[array < 0] = 0
-    if np.any(array > 1):
+        data[data < 0] = 0
```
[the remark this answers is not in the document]
ambrosejcarr: 🐛 !

If data is an xarray, we need to convert it to a numpy array to run these comparisons because indexing works differently between the two objects: http://xarray.pydata.org/en/stable/indexing.html#indexing-rules

This function was just never properly tested with an xarray input. np.ndarray[xr.DataArray] does not work.

starfish/imagestack/imagestack.py
shanaxel42: ohhhh I missed that. hmm well you could make dims_order a param you pass into _processing_workflow, then do then above in transform()... might be a little more clear

starfish/imagestack/imagestack.py
```diff
@@ -857,12 +857,24 @@ def _processing_workflow(
...
 
+        # build and then slice the xarray to get the piece needed for this worker
```
ttung: This does not set the xarray coordinates correctly. I imagine that if you use crop-on-load, your labels will all be incorrect.

PR #1041, opened 2019-02-22
Move preserve_float_range

ambrosejcarr:
## Purpose

Clipping and scaling was not happening properly in our filters, leading to some bugs. Specifically, `preserve_float_range` with `rescale=True` scales by the max value of the provided image. This was being called on _each chunk independently_, when some filters need to call it on the _whole imagestack_. This PR adds that flexibility. 

## Changes

- Move `preserve_float_range` calls into `ImageStack.apply` and enable 3 modes of clipping (see below).
- Move `preserve_float_range` from `filter` to `starfish.util.dtype`.
- Each filter that can create values outside [0, 1] now receives a `clip_method` parameter. 
- Add comments to `apply`, `transform` and related methods to clarify process flow. 
- Add tests for clipping types
- Improve correlation for MERFISH notebook by adjusting deconvolution clipping
- Fix miscellaneous documentation errors in the Filters

New `clip_method` parameter:
```
        clip_method : int
            (Default 0) Controls the way that data are scaled to retain skimage dtype
            requirements that float data fall in [0, 1].
            0: data above 1 are set to 1, and below 0 are set to 0
            1: data above 1 are scaled by the maximum value, with the maximum value calculated
               over the entire ImageStack
            2: data above 1 are scaled by the maximum value, with the maximum value calculated
               over each slice, where slice shapes are determined by the group_by parameters
```

## Review strategy

- review `starfish/imagestack/imagestack.py`
- review `starfish/test/image/test_apply.py`
- review `starfish/test/full_pipelines/api/test_merfish.py`
- glance at all the filters and the merfish notebook -- changes there are minimal.

### DIFF
--- a/starfish/image/_filter/bandpass.py
+++ b/starfish/image/_filter/bandpass.py
@@ -9,7 +9,7 @@ from starfish.imagestack.imagestack import ImageStack
 from starfish.types import Number
 from starfish.util import click
 from ._base import FilterAlgorithmBase
-from .util import determine_axes_to_group_by, preserve_float_range
+from .util import determine_axes_to_group_by
 
 
 class Bandpass(FilterAlgorithmBase):
@@ -75,7 +75,7 @@ class Bandpass(FilterAlgorithmBase):
             image, lshort=lshort, llong=llong, threshold=threshold,
             truncate=truncate
         )
-        return preserve_float_range(bandpassed)
+        return bandpassed
 
     def run(
             self, stack: ImageStack, in_place: bool = False, verbose: bool = False,
--- a/starfish/image/_filter/bandpass.py
+++ b/starfish/image/_filter/bandpass.py
@@ -6,7 +6,7 @@ import xarray as xr
 from trackpy import bandpass
 
 from starfish.imagestack.imagestack import ImageStack
-from starfish.types import Number
+from starfish.types import Clip, Number
 from starfish.util import click
 from ._base import FilterAlgorithmBase
 from .util import determine_axes_to_group_by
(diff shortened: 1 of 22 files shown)


starfish/image/_filter/bandpass.py
```diff
@@ -31,6 +33,14 @@ def __init__(
             deviations (default 4)
         is_volume : bool
             If True, 3d (z, y, x) volumes will be filtered. By default, filter 2-d (y, x) planes
+        clip_method : int
```
ttung: This is completely unintuitive. Please use an enum to denote the different clip methods.

starfish/spots/_detector/detect.py
```diff
@@ -226,8 +226,8 @@ def detect_spots(data_stack: ImageStack,
         )
 
     if reference_image_from_max_projection:
-        reference_image = data_stack.max_proj(Axes.CH, Axes.ROUND)
-        reference_image = reference_image._squeezed_numpy(Axes.CH, Axes.ROUND)
+        reference_image = data_stack.max_proj(Axes.CH, Axes.ROUND).xarray.squeeze()
+        # reference_image = reference_image._squeezed_numpy(Axes.CH, Axes.ROUND)
```
ttung: ehh?

starfish/imagestack/imagestack.py
```diff
@@ -857,13 +892,25 @@ def _processing_workflow(
...
-        numpy_array = unshaped_numpy_array.reshape(shape)
 
-        sliced = numpy_array[selector_and_slice_list[1]]
+        # get the correct dimension order for the imagestack based on AXES_DATA
+        dims = [
+            name.value for (name, data) in
+            sorted(AXES_DATA.items(), key=lambda kv: kv[1].order)
+        ] + [Axes.Y.value, Axes.X.value]
+
+        # build and then slice the xarray to get the piece needed for this worker
```
ttung: This drops the coords, right?

ambrosejcarr: From the filtered function, yes. From the resulting object, no.

PR #1043, opened 2019-02-22
validate: add further commands

joshmoore:
* [x] `validate codebook`
 * [x] `validate fov`
 * [x] `validate manifest`

Name change requests welcome.

### DIFF
--- a/sptx_format/cli.py
+++ b/sptx_format/cli.py
@@ -32,6 +32,23 @@ def validate(ctx, experiment_json, fuzz):
 
 
 @validate.command()
+@click.argument("codebook_file", metavar="JSON_FILE_OR_URL")
+@click.option("--fuzz", is_flag=True)
+@click.pass_context
+def codebook(ctx, codebook_file, fuzz):
+    """validate codebook against the jsonschemas"""
+
+    try:
+        valid = validate_sptx.validate_codebook(codebook_file, fuzz)
+        if valid:
+            ctx.exit(0)
+        else:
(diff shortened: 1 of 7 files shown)


sptx_format/cli.py
```diff
@@ -32,14 +32,65 @@ def validate(ctx, experiment_json, fuzz):
...
+
+@validate.command()
+@click.argument("file", metavar="JSON_FILE_OR_URL")
+@click.option("--fuzz", is_flag=True)
+@click.pass_context
+def fov(ctx, file, fuzz):
+    """validate field-of-view against the jsonschemas"""
+
+    try:
+        valid = validate_sptx.validate_file(file, "field_of_view/field_of_view.json", fuzz)
```
ttung: tiny nit: should avoid using python keywords as variable names.

sptx_format/test_experiment.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 from pkg_resources import resource_filename
 
+import starfish  # noqa
```
ttung: what's this for?

joshmoore: There's an import recursion somewhere between `starfish` and `sptx_format` that shows up in the `sptx_format` tests if importing relatively without first importing `starfish`.

sptx_format/cli.py
joshmoore: https://github.com/spacetx/starfish/pull/1062

PR #1044, opened 2019-02-22
iss: add validation command

joshmoore:
On adding a validation command, `test_iss.py` begins to fail due to a confusion between the older "hybridization.json" and the newer "primary_images.json". This PR:

 * [x] adds validation
 * [x] makes `get_cli_test_data.py` configurable
 * [x] updates `test_iss.py` to pass `--primary-name=hybridization.json`

(`test_iss.py` should have failed _previously_ which needs tracking down.)

### DIFF
--- a/starfish/config/__init__.py
+++ b/starfish/config/__init__.py
@@ -171,9 +171,14 @@ class StarfishConfig(object):
             v = v[k]
         v[lookup[-1]] = value
 
-    @staticmethod
-    def flag(name, default_value=""):
-        value = os.environ.get(name, default_value)
+    def flag(self, name, default_value=""):
+
+        if name in os.environ:
+            value = os.environ[name]
+            self._env_keys.remove(name)
+        else:
+            value = default_value
+
         if isinstance(value, str):
             value = value.lower()
             return value in ("true", "1", "yes", "y", "on", "active", "enabled")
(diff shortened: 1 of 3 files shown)


joshmoore:
With the last commit, the test fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpsltx0z3k/formatted/fov_001/hybridization.json'
```

since these are the contents of hybridization.json:

```
{"contents": {"fov_001": "fov_001/hybridization.json"}, "extras": null, "version": "0.0.0"}
```

joshmoore:
💚

ttung:
This is getting kind of wild and hacky.  What do you think is a good long-term solution, @joshmoore?

two ideas:
1. The `Experiment` objects exposes the filenames.
1. We gather up all the constituent `TileSets`, form the collection, and write that out instead.

joshmoore:
I could see making use of the filenames in Experiment, even elsewhere. I'm less sure of what your second suggestion would provide. From my side, it felt like there `get_cli_test_data.py` couldn't really afford to make any assumptions about the contained names and if that's the case, perhaps it becomes a more general `starfish cp` command.

Additionally, I was surprised that the commands in the test were taking json below experiment.json and this is probably at least partially the reason that validation wasn't taking place. One potential fix there would be to accept either, e.g. "registration takes experiment.json **or** primary_images.json", though it's perhaps too late for such changes.

ttung:
Ok, well I think we can table this for now.  It may resurface as we gain more experience with registered ImageStacks...

PR #1046, opened 2019-02-26
Fix how we cast from xarray to numpy

ttung:
There are some places in the codebase where we need a numpy array instead of an xarray.  Switch all instances where we do this to `np.asarray(...)`.  Also, fixed type hints where appropriate.

This should fix the master-breaking bug in #1035, but does not resolve the issue of (xarray) coordinates not being transferred to the constituent xarrays.

### DIFF
--- a/starfish/image/_filter/util.py
+++ b/starfish/image/_filter/util.py
@@ -134,10 +134,7 @@ def preserve_float_range(
     """
     array = array.copy()
 
-    if isinstance(array, xr.DataArray):
-        data = array.values
-    else:
-        data = array
+    data = np.asarray(array)
 
     if np.any(data < 0):
         data[data < 0] = 0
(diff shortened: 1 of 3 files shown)


ttung:
@kne42 What does skimage think of differences in API between xarray and numpy?  Should it hew to the lowest common denominator?

ttung:
> This can be replaced by `img = np.asarray(img)`.

Is there any advantage to that?

kne42:
It's a lot more general if you ever plan to allow any type of array to be passed in and really is just the standard for handling things like this, used by libraries such as `scipy.ndimage` (which calls `np.asarray` first thing in every function).

Specifically answering your earlier question, scikit-image tries to be agnostic to  what is being handed in by assuming that the passed array implements all the same methods as an `ndarray` and avoiding using methods on it in the first place (e.g. `np.dot` instead of `ndarray.dot`), so basically duck typing? That isn't applicable in this case since clearly xarray forgot some things.

PR #1048, opened 2019-02-28
Get the imagestack tile_shape from the TileCollectionData object.

ttung:
This simplifies the constructor a bit.

This is built on top of #1032

### DIFF
--- a/starfish/imagestack/parser/numpy/__init__.py
+++ b/starfish/imagestack/parser/numpy/__init__.py
@@ -102,6 +102,10 @@ class NumpyData(TileCollectionData):
         return keys
 
     @property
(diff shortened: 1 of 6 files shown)


starfish/imagestack/parser/numpy/__init__.py
```diff
@@ -101,6 +101,10 @@ def keys(self) -> Collection[TileKey]:
 
         return keys
 
+    @property
+    def tile_shape(self) -> Tuple[int, int]:
+        return self.data.shape[-2:]
```
shanaxel42: is this just to reverse the order?

ttung: No, it's to take the last two elements.

PR #1050, opened 2019-03-01
Make ImageStack a thinner wrapper around the xarray.

ttung:
1. Constructor takes the xarray and the tile_data.
2. Make a builder that takes a TileCollectionData and return an ImageStack.

This will allow us to avoid using the `from_numpy_array` escape hatch for methods like max_proj.

Depends on #1032, #1048, #1049

### DIFF
--- a/starfish/core/imagestack/imagestack.py
+++ b/starfish/core/imagestack/imagestack.py
@@ -87,24 +87,18 @@ class ImageStack:
            the shape of the image tensor by categorical index (channels, imaging rounds, z-layers)
     """
 
-    def __init__(
-            self,
-            tile_data: TileCollectionData,
-    ) -> None:
+    def __init__(self, data: MPDataArray, tile_data: Optional[TileCollectionData]=None):
+        self._data = data
+        self._tile_data = tile_data
+        self._log: List[dict] = list()
+
+    @classmethod
+    def from_tile_collection_data(cls, tile_data: TileCollectionData) -> "ImageStack":
         axes_sizes = {
             Axes.ROUND: len(set(tilekey.round for tilekey in tile_data.keys())),
             Axes.CH: len(set(tilekey.ch for tilekey in tile_data.keys())),
             Axes.ZPLANE: len(set(tilekey.z for tilekey in tile_data.keys())),
         }
-
-        self._tile_data = tile_data
-        self._log: List[dict]
-
-        # check for existing log info
-        if STARFISH_EXTRAS_KEY in tile_data.extras and LOG in tile_data.extras[STARFISH_EXTRAS_KEY]:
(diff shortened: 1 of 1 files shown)


starfish/imagestack/imagestack.py
```diff
@@ -107,24 +107,19 @@ class ImageStack:
         save the (potentially modified) image tensor to disk
     """
 
-    def __init__(
-            self,
-            tile_data: TileCollectionData,
-    ) -> None:
+    def __init__(self, data: MPDataArray, tile_data: Optional[TileCollectionData]=None):
```
ambrosejcarr: Is it possible for the constructor to take an `xarray` (not an `MPDataArray`) and have the constructor deal with the multiprocessing internally? Alternatively, can I have a `from_xarray` constructor? 

I really want to enable the following use case: 

-----

The user realizes they want to do something unexpected to the ImageStack. This is fine, because they can just drop down to the `xarray`! Take the below case, where the goal is to invert the values of the array so that the user can take a `"min_proj"`. 

```python
In [1]: decoded: ImageStack
In [2]: similarity = 1 - decoded.xarray
```

Now I can use `max_proj` to get the _minimum_ value in the array, because I've inverted it, and after that I could use `starfish.display.stack` to take a look at the result. 

```python
In [3]: label = similarity.max_proj(Axes.CH)                                                                     
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-20b057c9a575> in <module>
----> 1 label = similarity.max_proj(Axes.CH)

~/projects/spacetx/starfish/.venv/lib/python3.6/site-packages/xarray/core/common.py in __getattr__(self, name)
    177                     return source[name]
    178         raise AttributeError("%r object has no attribute %r" %
--> 179                              (type(self).__name__, name))
    180 
    181     def __setattr__(self, name, value):

AttributeError: 'DataArray' object has no attribute 'max_proj'
```
Crap. The user is in an `xarray`. Desired functionality: 

```python
In [4]: similarity_image = starfish.ImageStack(similarity)
In [5]: starfish.display.stack(similarity)
```

ttung: No, this is not possible today.  It's not possible to jam in the shared memory construct without doing an expensive copy.  If we move towards multithreading, we can make it just pass in an xarray.

ambrosejcarr: Ok. One thing that I _can_ do right now is: 

```python
similarity: ImageStack
new_imagestack = starfish.ImageStack.from_numpy_array(similarity.xarray)
```

However, `new_imagestack` won't have any of the coordinates from `similarity`. (they just get dummies, I think?)

If coordinates can be passed over, that's a workable solution for me!

ttung: I think if we de-risk multithreading and deem the perf implications to be acceptable, then this goes away and you get what you want.

Issue #664, opened 2018-10-01
TileFetcher and FetchedTile should be able to create spaceTx-format without copying data


ambrosejcarr:
Users want to be able to generate our metadata without copying image data, which may already exist in a desired location, e.g. an external database, or may be onerously large (`Tb`++). 

It seems like `TileFetcher` and `FetchedTile` could be adapted to reference the location of the images in the directory that is provided instead of writing a new set of Images. 

Better yet would be if a `--no-validate` mode existed in which the images themselves did not need to be read, but rather, just the names would be extracted to build the correct metadata. 

To do: 
- [ ] Can build spaceTx-format using starfish tools without _copying_ image data
- [ ] (Optional) Can build spaceTx-format using starfish tools without _reading or copying_ image data [edit: this is a pro-only feature; this will lead to long term horrible UX, and will not be compatible with the HCA... but would speed up prototyping in some cases]
- [ ] (For Biohub) TileFetcher can parse s3 "directories"

cc @ttung @kevinyamauchi -- this would Benefit the Biohub use case by helping them to support their Image Database. It might also be good for us in the longer term, as it could streamline our updates to spaceTx-format datasets that need to be changed as a result of changes to `starfish` and `slicedimage`.

ttung:
What's would be the path to these not-on-disk images?

ttung:
Summary of findings:

@berl could use a variant that produces the json files that reference existing images.  reading once to calculate the checksums is okay for this round.

joshmoore:
Though the paths wouldn't provide the tile name information, we could likely provide a version which read TIFFs from OMERO which may be useful in the Sanger usage where the Harmony data will first be imported. cc: @olatarkowska

berl:
bumping this because I'm not sure if it's still on the radar. @njmei is working on our analysis pipeline using starfish and we will have limited amount of buffer storage to do processing before we upload to our internal data storage and external archiving.  
cc: @ttung @ambrosejcarr @neuromusic

kevinyamauchi:
Thanks @berl for re-raising the issue. This would also be useful for us at the Biohub!

ttung:
Proposed solution:

`experiment.builder.providers.FetchedTile` should add a second method for providing a path to the original file.
when that method is provided, we should use a custom `tile_opener` in our calls to `Writer.write_to_path` that opens a fake file that accepts write calls but does nothing with them.  everything else will probably just work.

njmei:
[...]

PR #1053, opened 2019-03-02
Add the ability to create SpaceTx format without copying data

njmei:
See commit message.

I decided that it didn't make sense to create a toy dataset to test `write_experiment_json()`, the test would be too specific and brittle...

I'm definitely open to other suggestions though.

### DIFF
--- a/starfish/experiment/builder/__init__.py
+++ b/starfish/experiment/builder/__init__.py
@@ -1,5 +1,7 @@
+import io
 import json
 import os
+import sys
 from typing import (
     Any,
     BinaryIO,
@@ -35,17 +37,46 @@ AUX_IMAGE_NAMES = {
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
 
 
+class DevNull(io.BytesIO):
+    """A class meant to mimic an open(filepath, 'wb') operation but
+    prevents any actual writing, reading, or seeking.
+
+    This class is an ugly hack to prevent the slicedimage
+    Writer.generate_partition_document() function from needlessly creating
+    a copy of image data.
+
+    See: https://docs.python.org/3/library/io.html
+    See also: cpython/Lib/_pyio.py
+    """
+    def __init__(self, filepath, mode='wb', *args, **kwargs):
+        super(DevNull, self).__init__(*args, **kwargs)
+        self.name = filepath
+        self.mode = mode
+
+    def read(self, size=-1):
+        raise NotImplementedError()
+
+    def write(self, b):
+        return sys.getsizeof(b)
+
+    def seek(self, pos, whence=0):
+        raise NotImplementedError()
+
+
 def _tile_opener(toc_path: str, tile: Tile, file_ext: str) -> BinaryIO:
-    tile_basename = os.path.splitext(toc_path)[0]
-    return open(
-        "{}-Z{}-H{}-C{}.{}".format(
-            tile_basename,
-            tile.indices[Axes.ZPLANE],
-            tile.indices[Axes.ROUND],
-            tile.indices[Axes.CH],
-            ImageFormat.TIFF.file_ext,
-        ),
-        "wb")
+    if tile.filepath:
+        return DevNull(tile.filepath, mode='wb')
+    else:
+        tile_basename = os.path.splitext(toc_path)[0]
+        return open(
+            "{}-Z{}-H{}-C{}.{}".format(
+                tile_basename,
+                tile.indices[Axes.ZPLANE],
+                tile.indices[Axes.ROUND],
+                tile.indices[Axes.CH],
+                ImageFormat.TIFF.file_ext,
+            ),
+            "wb")
 
 
 def _fov_path_generator(parent_toc_path: str, toc_name: str) -> str:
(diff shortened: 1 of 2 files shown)


neuromusic:
this is meant to fix #664, right?

njmei:
@neuromusic It would only resolve the first entry in the 3 item checklist in #664

berl:
@ambrosejcarr @kevinyamauchi  it looks like checkboxes 2 and 3  in #664 are stale. Correct?

starfish/experiment/builder/providers.py
```diff
@@ -49,6 +49,20 @@ def extras(self) -> dict:
         """
         return {}
 
+    @property
+    def filepath(self) -> str:
```
ttung:
```suggestion
    def filepath(self) -> Optional[str]:
```

starfish/experiment/builder/providers.py
```diff
@@ -49,6 +49,20 @@ def extras(self) -> dict:
...
+
+        Providing this attribute will disable copying of data done by:
+        starfish.experiment.builder.write_experiment_json()
+
+        Returns
+        -------
+        str
+            A filepath string.
+        """
+        return ""
```
ttung:
```suggestion
        return None
```

starfish/experiment/builder/__init__.py
```diff
@@ -35,17 +37,40 @@
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
 
 
+class ReadOnlyTile(io.BytesIO):
```
ttung: I wouldn't call this `ReadOnlyTile`.  It's more like a `/dev/null/`.

Furthermore, I would stub out the `read` and `seek` methods because both of them won't behave as one might expect.

starfish/experiment/builder/providers.py
njmei: Hmm... I was trying to be consistent with the `extras` property. Should that also be changed to `Optional[dict]` and return `None`?

starfish/experiment/builder/__init__.py
```diff
@@ -137,6 +168,7 @@ def build_image(
                 image.shape,
                 extras=image.extras,
             )
+            tile.filepath = image.filepath
```
shanaxel42: I'm not sure I understand this? Are you setting tile.filepath every-time? If so wouldn't we always hit the DevNull case?

njmei: The `filepath` property in the modified [starfish/experiment/builder/providers.py ](https://github.com/spacetx/starfish/pull/1053/commits/511a5b9dc5ac7b717326b350baa154a8f977f63b#diff-36f200ab95f1d7cb24a4d35429e88ee1R53) defaults to `None`. So the DevNull case will only happen if the `TileFetcher` implementation explicitly sets a value to `filepath` for a  `FetchedTile`.

shanaxel42: ahh ok that makes sense, thanks!

ttung:
is this ready for review again?

njmei:
Hey Tony,
I'd like to hold off on merging until I've run a full current Allen dataset (just to be extra sure). This has taken a bit because we've been reworking our mfish metadata format to convert more nicely starfish.
-Nick

ttung:
If you successfully convert a FOV, that's a pretty good sign.

njmei:
@ttung @berl 

Well, it works in the sense that no files are copied. But the resulting FOV .jsons are 'wrong' in the sense that we'd like to pass in relative pathnames but `slicedimage io` looks to always just take the basename of a filepath. (see: https://github.com/spacetx/slicedimage/blob/2c5b76037709d135d7a285047895184438e36226/slicedimage/io.py#L293)

berl:
@ttung @njmei I would rather keep the spacetx json files all in one place, but just allow the path to the .tif file to be a full path.  Right now, I think it's safe to make them relative paths from the Experiment.json instead of absolute paths.

njmei:
@ttung We decided to make the file structure thing a different issue. Let me know if there are any other changes I should make before merging.

starfish/experiment/builder/__init__.py
```diff
@@ -35,17 +37,46 @@
...
+    See: https://docs.python.org/3/library/io.html
+    See also: cpython/Lib/_pyio.py
+    """
+    def __init__(self, filepath, mode='wb', *args, **kwargs):
+        super(DevNull, self).__init__(*args, **kwargs)
+        self.name = filepath
+        self.mode = mode
+
+    def read(self, size=-1):
+        return bytes(b'')
```
ttung: I think this should raise an error if it's called.  We're certainly not expecting this to ever be called.

starfish/experiment/builder/__init__.py
```diff
@@ -35,17 +37,46 @@
...
+
+    def write(self, b):
+        return sys.getsizeof(b)
+
+    def seek(self, pos, whence=0):
+        return pos
```
ttung: ditto.

starfish/experiment/builder/__init__.py
njmei: Do you have a preference on error type or just: `NotImplementedError`?

njmei:
Tested and things look fine.

njmei:
Closing this pull request in favor of: 6fbe555bf92ea1f0f2617b63cca22c6ed38658dd

Issue #562, opened 2018-09-14
IntensityTable should have a command to combine with other tables


ambrosejcarr:
In the most Naive formulation, where we assume that images have been pre-registered and cut into fields of view for starfish, we can reconstruct the data for the complete experiment by simply concatenating IntensityTables, while transforming the x, y, and z coordinates according to tile offsets calculated from the global coordinates of the ImageStack. 

We may want to store coordinates as an `Attribute` in the `IntensityTable`, once they are available in the `ImageStack` and `TileSet` objects, to facilitate this operation.

PR #1055, opened 2019-03-02
Example of FOV iteration & Beginning of concat method

shanaxel42:
This PR provides an example of the aligned image group iteration workflow on the 3d_smFish test data. It also starts a concat method in IntensityTable that just does a simple xarray.concat, but should be extended to include different overlap strategies, including Brian's (if overlap only include spots from FOV with more spots).


ambrosejcarr:
provides an MVP of the requested feature in #562

Issue #1005, opened 2019-02-13
Unifying Spot Finding


ambrosejcarr:
# SpotFinding in starfish

This issue pulls together a series of investigations into how to support spot-finding and decoding approaches used by SpaceTx labs. I've chronicled some proposed changes and open questions. It is a work in progress that would greatly benefit from user and developer feedback. 

science cc @kevinyamauchi @berl @sofroniewn @gapartel
tech cc @ttung @shanaxel42 @kne42 

Addresses issues brought up in #380 #874 #373 #708, #499, #525, #759 and #908 

## Definitions

**Spot**: In the context of image-based transcriptomics, a spot is an ellipsoid blob that is the result of capturing light generated by a fluorophore that has been intentionally and specifically bound to an RNA molecule to visualize the molecule's spatial location.

**Blob finding**:  blob finding is the process of detecting spots an n-dimensional image. Blob finding is accomplished by integrating the intensity of an image over a gaussian kernel, returning the coordinates of bright blobs on dark backgrounds. The shape and sign of the kernel preferentially detects ellipsoid blobs equal in width to the kernel. 

**Region of interest (ROI) detection**: ROI detection is a more general approach that does not require the region to have a specific shape. In image-based transcriptomics there are two uses of ROI detection: instance-based segmentation of cells and spots. Spot ROI detection is currently accomplished in starfish by connecting adjacent pixels together when (1) the decode to the same gene and (2) they meet a minimum intensity criteria.

## Blob finding

At present, blob finding is carried out by three methods in starfish:
1. `BlobDetector`. This method enables users to employ a Laplacian of Gaussians filter that enhances spots (bright circular objects with dark backgrounds) that match the standard deviation of the gaussian filter, and uses the skimage `local_max_peak_finder` approach on the Laplacian filtered image to identify the resulting spots that are brighter than a specified threshold value. It will attempt to find the standard deviation that best fits a given spot, selecting from a range of values passed by the user. It also enables users to approximate the Laplacian filtered image using a Difference of Gaussians or Determinant of Hessians approximation to save time. 
2. `LocalMaxPeakFinder`: This method applies a Laplacian filter with a _specific_ standard deviation to an image, testing numerous different thresholds. Like `BlobDetector` above, it uses `local_max_peak_finder`: to identify spots. The method selects an appropriate threshold by identifying a stable regime where modifying the threshold does not change the number of spots detected.
[...]

PR #1060, opened 2019-03-03
[WIP] decoding filter and label transform

ambrosejcarr:
## Decoding filter 

This filter calculates the distance of each gene target in the codebook to the tiles across `(round, channel)`. The resulting ImageStack is of shape `(1, n_codes, z, y, x)` where `n_codes` is the number of gene targets in the codebook

[This link](http://www.giphy.com/gifs/jICSOgY01aaqSMV9MJ) contains an example output from the filter. The clim is first set manually, then the similarity of each code is iterated through. You can see that there is highly uneven expression of different genes. These results match what was observed by the original study authors (r^2 ~ 0.98)

This code projects across channels to identify the maximum similarity score with a gene, which can then be visualized. 
```
# filter image
dfilt = starfish.image.Filter.Decode(experiment.codebook)
decoded = dfilt.run(quantile_normalized)

# transform distance to similarity
similarity = starfish.ImageStack.from_numpy_array(1 - decoded.xarray)

# find maximum similarity across codebook
label = similarity.max_proj(Axes.CH)

# look at the max-similarity projection
starfish.display.stack(label)
```

[This link](http://www.giphy.com/gifs/1X7hdHewl6txITF3N7) contains an example of visualizing the output of the above code in napari, and using the clim slider to walk through the range of similarity thresholds. This can be used to visually inspect and select the right similarity for the pixel spot detector on these data. The final frame is a similarity of about 0.45, or a max distance of 0.55. 

Note that this filter calculates the full cross-distance matrix between each code and pixel. This can get very, very large in cases where there are a lot of pixels or a large codebook (e.g. SeqFISH, which has a 10,000 code codebook). 

## Label transform

To enable similar functionality to above on a large experiment or large codebook, this PR also adds a label transform, which carries out an approximate nearest-code matching of a user-specified number `k_targets`. This transform emits: 

1. A `4-d (k_targets, z, y, x)` image whose values are the ranked distances to the k-closest codes. 
2. A `4-d (k_targets, z, y, x)` image whose values are the ranked indices that map each distance in (1) to an integer representing the code. 
3. an `index_to_code` dict that allows the integer code IDs to be mapped to their gene targets. 

This enables similar functionality to the above decode function for cases where the data cannot all fit in memory, or for which a dense `cdist` calculation is not practically computationally feasible.

cc @kevinyamauchi @berl I'd love any comments you might have on this. Next step is to fix pixel spot decoding to work cumulatively on the results of this filter, and to enable both (1) connected-component based roi-finding and (2) spot-finding on the distance maps.

### DIFF
--- a/starfish/image/_filter/decode.py
+++ b/starfish/image/_filter/decode.py
@@ -1,0 +1,116 @@
+import numpy as np
+import xarray as xr
+from scipy.spatial.distance import cdist
+
+from starfish.codebook.codebook import Codebook
+from starfish.imagestack.imagestack import ImageStack
+from starfish.types import Axes, Features
+from starfish.util import click
+from ._base import FilterAlgorithmBase
+
+# TODO ambrosejcarr replace this with constitutive channel fluorescence mask when that PR is merged
+def _generate_test_codebook():
+    n_channel = 2
+    n_round = 2
+    data = np.zeros((n_channel, n_channel, n_round))
+    for c in range(n_channel):
+        data[c, c] = 1
+        code_names = [str(c) for c in range(n_channel)]
+    return Codebook._create_codebook(code_names, n_channel, n_round, data)
+
+
+class Decode(FilterAlgorithmBase):
+
+    def __init__(self, codebook: Codebook, metric: str="euclidean", norm_order: int=2) -> None:
+        """Decode an image in round/channel space to target space
+
+        The resulting ImageStack will be of shape (1, n_targets, z, y, x) where n_targets are the
+        number of targets in the codebook. The values of the decoded images are the distance of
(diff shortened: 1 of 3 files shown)


kevinyamauchi:
Awesome! I'm glad you're working on this!

Rather than doing pairwise distances, what do you think about using a K-D tree to speed up the distance calculation?

kevinyamauchi:
I think I have the same confusion as @ttung in #1005 . What is the value of calculating the full similarity matrix over calculating a similarity score for the k nearest codes?

ambrosejcarr:
@kevinyamauchi I think this PR is superseded by the work you've done on the base calling filter. I suspect I should close this. What do you think?

PR #1066, opened 2019-03-06
docs: update sptx_format/README.md with new file names

joshmoore:
cc: @zperova


ttung:
My general concern with this entire section of the doc is that it implies that files have to have a certain name.  I guess we already have that confusion, but I wonder if it's possible to clarify it.

zperova:
@ttung isn't this the standard output of the starfish pipeline? or the output file names are defined by the user?

joshmoore:
@zperova : you're right that these are the standard names as generated by the library (which is what I updated to) but at the same time, _none_ of the names is required by the spec:

```
e.json
  c.json
  p.json
  n.json
```
etc. would be fine as long as you know to pass e.json as the experiment and all the internal names are correct.

ttung:
> @ttung isn't this the standard output of the starfish pipeline? or the output file names are defined by the user?

Output file names are defined by the user.

joshmoore:
@ttung : is that more like what you were thinking?

PR #1067, opened 2019-03-06
RFC: move sptx_format inside starfish pkg

ttung:
This gets rid of wonky recursive import problems like https://github.com/spacetx/starfish/pull/1043/files#r262371442.


ambrosejcarr:
Can we rename this `spacetx_format`? I am not a fan of the name.

joshmoore:
Happy to leave this up to you all; I don't have any strong objections. (Do note that I was a fan of having the format in a separate repo so you probably don't want to listen to me anyway... :smile:)

An alternative if you're looking for them is forbidding imports of starfish from in `sptx_format`:

```
/opt/hca/starfish $git grep starfish sptx_format | grep format | grep py
sptx_format/cli.py:from starfish.util import click
sptx_format/test_codebook.py:from starfish.codebook._format import DocumentKeys
sptx_format/test_experiment.py:import starfish  # noqa
sptx_format/test_fuzz.py:from starfish.codebook._format import CURRENT_VERSION, DocumentKeys
sptx_format/validate_sptx.py:from starfish.config import StarfishConfig
sptx_format/validate_sptx.py:from starfish.util import click
```

which would mean one or more of:
 * have a util module to contain the shared bits
 * move sptx_format/cli.py into starfish/cli
 * move tests to starfish/tests

ttung:
Renamed to `starfish.spacetx_format`

PR #1072, opened 2019-03-08
Change tile shape to a dict by default

### LINKED ISSUE #528: FieldOfView tile_shape has ambiguous (x, y) ordering
The schema claims that it is ordered (x, y)
https://github.com/spacetx/sptx-format/blob/master/schema/field_of_view/field_of_view.json#L91

Internally, we don't validate on this field yet, we simply slurp in the data and make sure that the sizes fit. However, we might want to provide information on a TileSet (before it's loaded into memory).

We should replace this with an explicit map of {x: size, y: size}, instead of a list to prevent errors.

Ported from spacetx/sptx-format#23

ttung:
1. Where tile shape is being used, we use a dict between axis and the size of the tile along that axis.
2. Where there is a contract for tile data, we indicate that the data should be row-major.

Test plan: `make -j fast`  tests on travis will not complete until https://github.com/spacetx/slicedimage/pull/81 has landed and a new version is released.

Resolves #528

### DIFF
--- a/starfish/imagestack/parser/_tiledata.py
+++ b/starfish/imagestack/parser/_tiledata.py
@@ -11,11 +11,18 @@ class TileData:
     Base class for a parser to implement that provides the data for a single tile.
     """
     @property
-    def tile_shape(self) -> Tuple[int, int]:
+    def tile_shape(self) -> Mapping[Axes, int]:
         raise NotImplementedError()
 
     @property
     def numpy_array(self) -> np.ndarray:
+        """Return the image data representing the tile.  The tile must be row-major.
+
+        Returns
+        -------
+        ndarray :
+            The image data
+        """
         raise NotImplementedError()
 
     @property
@@ -41,7 +48,7 @@ class TileCollectionData:
         raise NotImplementedError()
 
     @property
-    def tile_shape(self) -> Tuple[int, int]:
+    def tile_shape(self) -> Mapping[Axes, int]:
         """Returns the shape of a tile."""
         raise NotImplementedError()
 
--- a/starfish/imagestack/parser/tileset/_parser.py
+++ b/starfish/imagestack/parser/tileset/_parser.py
@@ -22,11 +22,14 @@ class _TileSetConsistencyDetector:
     trigger a :py:class:`starfish.errors.DataFormatWarning`.
     """
     def __init__(self) -> None:
-        self.tile_shape: Optional[Tuple[int, ...]] = None
+        self.tile_shape: Optional[Mapping[Axes, int]] = None
         self.kind = None
         self.dtype_size = None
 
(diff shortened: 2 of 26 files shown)


starfish/imagestack/parser/_tiledata.py
```diff
@@ -11,11 +11,18 @@ class TileData:
     Base class for a parser to implement that provides the data for a single tile.
     """
     @property
-    def tile_shape(self) -> Tuple[int, int]:
+    def tile_shape(self) -> Mapping[Axes, int]:
         raise NotImplementedError()
 
     @property
     def numpy_array(self) -> np.ndarray:
+        """Return the image data representing the tile.  The tile must be row-major.
```
ambrosejcarr: What happens if it isn't? Can we detect these problems? Can we emit a warning or throw an error?

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -22,20 +22,23 @@ class _TileSetConsistencyDetector:
...
         """As each tile is parsed, the tile shape should be reported via this method.  If an
         inconsistency is detected, a ValueError exception will be raised.
 
         Parameters
         ----------
         r, ch, zplane : int
             The indices of the tile, which is used to generate the exception text.
-        tile_shape : Tuple[int, ...]
-            The shape of the tile.
+        tile_shape : Tuple[Axes, int]
```
ambrosejcarr:
```suggestion
        tile_shape : Mapping[Axes, int]
```

starfish/imagestack/parser/tileset/_parser.py
```diff
@@ -99,9 +102,11 @@ def _load(self):
         self._expectations.report_dtype(self._r, self._ch, self._zplane, self._numpy_array.dtype)
 
     @property
-    def tile_shape(self) -> Tuple[int, int]:
+    def tile_shape(self) -> Mapping[Axes, int]:
         self._load()
-        tile_shape = self._numpy_array.shape
+        raw_tile_shape = self._numpy_array.shape
+        assert len(raw_tile_shape) == 2
```
shanaxel42: why are we starting to assert the length here? Is there a case when the raw shape wouldn't be length 2?

starfish/imagestack/parser/_tiledata.py
ttung: Not really.  We can only tell when the data is misshapen.

starfish/imagestack/parser/tileset/_parser.py
ttung: fixed

starfish/imagestack/parser/tileset/_parser.py
ttung: If we start supporting N-dimensional chunks where N>2, then we may need to adjust the code.

Issue #1057, opened 2019-03-02
Codebook needs a null expected value


ambrosejcarr:
Currently, we represent smFISH codebooks as follows: 

```python
In [6]: smfish                                                                                                    
Out[6]: 
<starfish.Experiment (FOVs=1)>
{
fov_001: <starfish.FieldOfView>
  Primary Image: <slicedimage.TileSet (z: 34, r: 2, c: 3, x: 2048, y: 2048)>
  Auxiliary Images:
    nuclei: <slicedimage.TileSet (r: 2, c: 1, x: 2048, y: 2048)>
}

In [7]: smfish.codebook                                                                                           
Out[7]: 
<xarray.Codebook (target: 3, c: 3, r: 2)>
array([[[1, 0],
        [0, 0],
        [0, 0]],

       [[0, 0],
        [1, 0],
        [0, 0]],

       [[0, 0],
        [0, 0],
        [1, 0]]], dtype=uint8)
Coordinates:
  * target   (target) object 'Nmnt' 'Ptprt' 'Nxph2'
  * c        (c) int64 0 1 2
  * r        (r) int64 0 1
```

And we decode by measuring distance from the expected codes. However, this is not strictly correct -- smFISH only measures each gene in each round/channel _separately_, and as @berl has pointed out many times (sorry), there may be spots that overlap from other genes, and these cases should be _discounted_ by the algorithms, rather than result in a worse spot count. 

Proposal: allow smFISH or other codebooks to store JSON `null`, or python `None` or `np.NaN` in codebooks for `(round, ch)` pairs that should not be measured, and update spot finding to respect the presence of these values. In the above example, I envision all zeros becoming np.NaN. 

cc @kevinyamauchi

berl:
@ambrosejcarr  the current format allows for sparse representation of the codebook on disk, with unspecified r,c combinations assumed to be zero. Could we have an import mode or tag that would accomplish your suggestion while retaining the ability to store sparse codebooks?

ttung:
Doesn't the IntensityTable for smFISH have a bunch of `NaN`s for the rounds/channels that shouldn't be measured?  If that's factored into distance measurement, wouldn't that be sufficient?

ambrosejcarr:
> Doesn't the IntensityTable for smFISH have a bunch of NaNs for the rounds/channels that shouldn't be measured? If that's factored into distance measurement, wouldn't that be sufficient?

[...]

PR #1073, opened 2019-03-10
Explain how SpaceTx codebook zero/missing values are interpreted

ambrosejcarr:
This PR clarifies how SpaceTx codebook zeros and missing values should be interpreted. 

Future PRs will implement the logic suggested for implementing libraries, as noted by @berl (see #1057).

### DIFF
--- a/starfish/spacetx_format/README.md
+++ b/starfish/spacetx_format/README.md
@@ -212,3 +212,18 @@ In contrast, a coded experiment may have a more complex codebook:
 
 The above example describes the coding scheme of an experiment with 2 rounds and 2 channels, where each code expects exactly two images out of four to produce signal for a given target.
 In the above example, a spot in the image tensor would decode to `SCUBE2` if the spot was detected in (round=0, channel=0) and (round=0, channel=1).
+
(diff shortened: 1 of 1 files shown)


berl:
thanks @ambrosejcarr  this is exactly what I was thinking of. Look forward to seeing example implementation in decoders.

sptx_format/README.md
```diff
@@ -212,3 +212,18 @@ In contrast, a coded experiment may have a more complex codebook:
 
 The above example describes the coding scheme of an experiment with 2 rounds and 2 channels, where each code expects exactly two images out of four to produce signal for a given target.
 In the above example, a spot in the image tensor would decode to `SCUBE2` if the spot was detected in (round=0, channel=0) and (round=0, channel=1).
+
+Note that the codebook only states opinions about non-zero expected fluorescence values, and is not designed to distinguish between a literal `0` and a missing value, which would imply "any fluorescence level". For example, the following two codewords treated identically and cannot be distinguished at the level of the codebook.
```
ambrosejcarr:
```suggestion
Note that the codebook only states opinions about non-zero expected fluorescence values, and is not designed to distinguish between a literal `0` and a missing value, which would imply "any fluorescence level". For example, the following two codewords are treated identically and cannot be distinguished at the level of the codebook.
```

PR #1074, opened 2019-03-11
Add LocalSearchBlobDetector to support BaristaSeq, SeqFISH, STARmap

ambrosejcarr:
This PR adds a Spot Detector that enables a local search across (round, channel) pairs. It is able to overcome minor registration errors in data for which the average registration error is smaller than the average distance to the closest spot. 

This PR seeds the local search in a single round, which is adequate for BaristaSeq and STARmap, but multi-round consensus seeding will be added in future PRs for SeqFISH. 

The private methods have been separated out with an eye towards the eventual spot finding refactor in #1005.

### DIFF
--- a/notebooks/py/STARmap.py
+++ b/notebooks/py/STARmap.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python
+# coding: utf-8
+#
+# EPY: stripped_notebook: {"metadata": {"kernelspec": {"display_name": "Python 3", "language": "python", "name": "python3"}, "language_info": {"codemirror_mode": {"name": "ipython", "version": 3}, "file_extension": ".py", "mimetype": "text/x-python", "name": "python", "nbconvert_exporter": "python", "pygments_lexer": "ipython3", "version": "3.6.5"}}, "nbformat": 4, "nbformat_minor": 2}
+
+# EPY: START markdown
+### Load Data
+# EPY: END markdown
+
+# EPY: START code
+import os
+from itertools import product
+from functools import partial
+from typing import Tuple
+
+import numpy as np
+
+import starfish
+import starfish.data
+from starfish.types import Axes
+# EPY: END code
+
+# EPY: START code
+# EPY: ESCAPE %gui qt
+import starfish.display
+# EPY: END code
+
+# EPY: START code
+experiment = starfish.data.STARmap(use_test_data=True)
+stack = experiment['fov_000'].get_image('primary')
+# EPY: END code
+
+# EPY: START code
+# look at the channel/round projection
+ch_r_projection = stack.max_proj(Axes.CH, Axes.ROUND)
+# EPY: END code
+
+# EPY: START code
+starfish.display.stack(ch_r_projection)
+# EPY: END code
+
+# EPY: START markdown
+#It actually looks like there is a small shift approximately the size of a spot in the `x = -y` direction for at least one (round, channel) pair (see top left corner for most obvious manifestation).
+#
+#Attempt a translation registration to fix. 
+# EPY: END markdown
+
+# EPY: START code
+# Starmap only requires translation. Verify that things are registered with a quick 
+# similarity registration. 
+
+from skimage.feature import register_translation
+from skimage.transform import warp
+from skimage.transform import SimilarityTransform
+
+def _register_imagestack(target_image, reference_image, upsample_factor=5):
+    target_image = np.squeeze(target_image)
+    reference_image = np.squeeze(reference_image)
+    shift, error, phasediff = register_translation(target_image, reference_image, upsample_factor=1)
+    return SimilarityTransform(translation=shift)
+
+# identify the locations of all the spots by max projecting over z
+projection = stack.max_proj(Axes.CH, Axes.ZPLANE)
+reference_image = projection.sel({Axes.ROUND: 1}).xarray
+
+# learn the transformations for each stack
+register_imagestack = partial(
+    _register_imagestack, reference_image=reference_image, upsample_factor=5
+)
+transforms = projection.transform(register_imagestack, group_by={Axes.ROUND}, n_processes=1)
+# EPY: END code
+
+# EPY: START code
+[t.translation for (t, ind) in transforms]
+# EPY: END code
+
+# EPY: START markdown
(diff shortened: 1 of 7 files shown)


starfish/compat.py
```diff
@@ -98,3 +99,292 @@ def match_histograms(image, reference, multichannel=False):
             matched = _match_cumulative_cdf(image, reference)
 
         return matched
+
+if version.parse(skimage.__version__) > version.parse("0.14.2"):
```
shanaxel42: Can we not enforce that users have skiimage version > 0.14.2?

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # TODO implement consensus seeding (SeqFISH)
+
+        intensity_table = self._build_intensity_table(
+            per_round_spot_results, distances, indices,
+            rounds=data.xarray[Axes.ROUND.value].values, channels=data.xarray[Axes.CH.value].values,
+            search_radius=self.search_radius, anchor_round=self.anchor_round
+        )
+
+        transfer_physical_coords_from_imagestack_to_intensity_table(
```
shanaxel42: If this gets called through the detect_spots() path you shouldn't need to add this here. detect_spots calls it at the end regardless of the detect method.

ambrosejcarr: Unfortunately this does not get called through that pattern; I would need to refactor it but I think I'm more generally going to combine a bunch of the spot detection methods according to #1005

starfish/compat.py
ambrosejcarr: 0.14.2 is the current release, when they next release we'll be able to remove this.

notebooks/py/STARmap.py
```diff
@@ -0,0 +1,128 @@
...
+lsbd = starfish.spots._detector.local_search_blob_detector.LocalSearchBlobDetector(
+    min_sigma=1,
+    max_sigma=8,
+    num_sigma=10,
+    threshold=np.percentile(np.ravel(substack.xarray.values), 95),
+    exclude_border=2,
+    anchor_round=0,
+    search_radius=10,
+)
+intensities = lsbd.run(stack, n_processes=8)
```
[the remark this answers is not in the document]
ambrosejcarr: Oops, I was actually limiting the n_processes when running this on my home machine. I'll remove this so it defaults to n_cpu (probably 8 for most of our laptops.

notebooks/py/STARmap.py
```diff
@@ -0,0 +1,128 @@
...
+
+# identify the locations of all the spots by max projecting over z
+projection = stack.max_proj(Axes.CH, Axes.ZPLANE)
+reference_image = projection.sel({Axes.ROUND: 1}).xarray
+
+# learn the transformations for each stack
+register_imagestack = partial(
+    _register_imagestack, reference_image=reference_image, upsample_factor=5
+)
+transforms = projection.transform(register_imagestack, group_by={Axes.ROUND}, n_processes=1)
```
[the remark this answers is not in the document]
ambrosejcarr: If you're willing to show me how I'd love to make this refactor. In the interest of getting this and some dependent PRs in, I'm going to merge this but am happy to take on the follow-up.

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+}
+
+
+class LocalSearchBlobDetector(SpotFinderAlgorithmBase):
+
+    def __init__(
+            self,
+            min_sigma: Union[Number, Tuple[Number, ...]],
+            max_sigma: Union[Number, Tuple[Number, ...]],
+            num_sigma: int,
```
ttung: This parameter is not threaded into the detector method.

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+            threshold=self.threshold,
+            exclude_border=self.exclude_border,
+            **self.detector_kwargs
+        )
+
+        # if spots were detected
+        if results.shape[0]:
+
+            # measure intensities
+            z_inds = results[:, 0].astype(int)
```
ttung: Does this not work if results is empty?  I didn't look over this code super carefully but it doesn't seem like anything here strictly requires results be non-empty.

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+                    ]
+                )
+            )
+
+        return spot_data
+
+    def _find_spots(
+        self,
+        data_stack: ImageStack,
+        verbose: bool=False,
```
ttung: this is not used anywhere.

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+        help="Name of the type of the skimage blob detection method.")
+    @click.option(
+        "--search-radius", default=3, type=int,
+        help="Number of pixels over which to search for spots in other image tiles.")
+    @click.pass_context
+    def _cli(
+        ctx, min_sigma, max_sigma, threshold, overlap, show, detector_method, search_radius
+    ) -> None:
+        instance = LocalSearchBlobDetector(
+            min_sigma, max_sigma, threshold, overlap,
```
ttung: What's the story with sigma?  I see sometimes that it's allowed to be a tuple and sometimes it has to be a scalar.  In any case, this cli setup does not allow this to be a tuple.

starfish/spots/_detector/local_search_blob_detector.py
```diff
@@ -0,0 +1,423 @@
...
+            Number of processes to devote to spot finding. If None, will use the number of available
+            cpus (Default None).
+
+        Returns
+        -------
+        IntensityTable
+            Contains detected coded spots.
+
+        """
+        per_tile_spot_results = self._find_spots(data, verbose=verbose, n_processes=n_processes)
```
ttung: `_find_spots` returns `Dict[Tuple[int, int], np.ndarray]`
`_merge_spots_by_round` accepts a `Dict[Tuple[int, int], pd.DataFrame]`

one of these is wrong. :)

PR #1077, opened 2019-03-11
Unbreak master

ttung:
Fix how we use `starfish.display`.
remove `assay_comparison.py`


kne42:
tbh not sure if I wanna resurrect `starfish.plot`. I'll leave @ambrosejcarr to decide on this one

ambrosejcarr:
@ttung @kne42 It looks like part of this PR is likely adequate to unbreak master -- just the notebook changes are needed to cover the refactor that was made for `starfish.display`, not any of the subsequent resurrection.

ttung:
The `assay_comparison` notebook refers to `starfish.plot`.

kne42:
I deleted the `.py` and forgot to delete its notebook counterpart and that's why it's still there.

ttung:
Ahh, should I drop the assay_comparison notebook then?

kne42:
Yeah, I would say drop that and remove `starfish.plot` and then it should be good to go

ttung:
Ok, updated.

ttung:
Notebooks aren't run as part of non-master travis runs. :)

kne42:
Ah, that would explain why it passed in my PR.

PR #1078, opened 2019-03-12
Remove unneeded expression matrix code

ambrosejcarr:
- `IntensityTable.to_expression_matrix` takes an optional regions parameter that is produced by code that has been deleted from starfish.


starfish/intensity_table/intensity_table.py
```diff
@@ -426,21 +419,12 @@ def to_expression_matrix(self, regions: Optional[regional.many]=None) -> Express
...
-                "z": ("cells", np.zeros(counts.shape[0]))
-            }
-        else:
-            grouped = self.to_features_dataframe().groupby(['cell_id'])[['x', 'y', 'z']]
-            min_ = grouped.min()
-            max_ = grouped.max()
-            coordinate_df = min_ + (max_ - min_) / 2
-            metadata = {name: ("cells", data.values) for name, data in coordinate_df.items()}
-            metadata['area'] = ("cells", np.full(counts.shape[0], fill_value=np.nan))
+        grouped = self.to_features_dataframe().groupby(['cell_id'])[['x', 'y', 'z']]
```
shanaxel42: does this `groupby(['cell_id'])` work for you? When I was doing the coordinate work I kept getting the error that `cell_id` was not valid

ambrosejcarr: I pushed an addition to the ISS vignette that shows the sequence of pipeline component calls needed for this to work, updated the documentation with a requirement that spots be previously assigned to cells, and added an error in the case that the correct path is not followed.

PR #1080, opened 2019-03-12
Bump certifi

ttung:
It seems like Docker build can't deal with requests to install older versions of certifi other than is installed by travis by default.

### DIFF
--- a/REQUIREMENTS-STRICT.txt
+++ b/REQUIREMENTS-STRICT.txt
@@ -2,7 +2,7 @@
 appnope==0.1.0
 backcall==0.1.0
 bleach==3.1.0
-certifi==2018.11.29
+certifi==2019.3.9
 chardet==3.0.4
 Click==7.0
 cloudpickle==0.8.0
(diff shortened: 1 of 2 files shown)


ttung:
@joshmoore any idea why this is necessary?

here's the [log from a failed run](https://travis-ci.com/spacetx/starfish/jobs/184292232).

joshmoore:
>  any idea why this is necessary?

This last time I looked into this, I think I got dug long enough to find out that "conda and pip don't work flawlessly together". (cF. https://github.com/pypa/pip/issues/5247)

This is where the `--ignore-installed` in https://github.com/spacetx/starfish/pull/713 came from. What's changed here is that we now have  a call to `pip` in [environment.yml](https://github.com/spacetx/starfish/blob/master/environment.yml). I've had _some_ success with adding flags to environment.yml files.  I'd suggest trying:

```
  - pip:
    - --ignore-installed
    - ...
```

Another option would be to try to force conda use the version of certifi that pip is expecting.

----

NB: something I found while searching around that might be worth a second look -- https://picky-conda.readthedocs.io/en/latest/

ttung:
Looks like flags in the pip section of `environment.yml` is [not supported](https://github.com/conda/conda/issues/6805).

further confirmed with https://travis-ci.com/spacetx/starfish/builds/104144068

ttung:
I prefer this approach.

ttung:
Closing in favor of #1085

PR #1093, opened 2019-03-18
Enable appending to existing napari viewer in display()

### LINKED ISSUE #1086: Feature request: display() should be able to add images to an existing viewer
I think it would be useful if `display()` had the option to add the image as a new layer to an existing Napari `viewer`. This is super useful for comparing the image before/after a processing step (e.g., filtering).

Thoughts, @kne42 and @ambrosejcarr ?

kevinyamauchi:
# Overview
This pull request adds the ability to display multiple `ImageStack` objects in the same napari viewer, as discussed in #1086. Towards this, I added the option to pass in a napari `Viewer` object to `display()`. If the user passes in a `Viewer` object, the stack and/or spots are added as new layers in the existing `Viewer`. If `None` (default) is passed, the stack and/or spots are displayed in a new `Viewer`.  Closes #1086.

# Question
I am not sure what to do for the `viewer` type hint (line 111). It seems to me like we are trying to avoid importing napari at the top of `_display.py` to ease the dependency, so I am not sure how to give the `Viewer` class as a type hint. Thoughts?

# Usage
```python
from starfish import data, FieldOfView, display

# Download the ISS sequencing and dot images
experiment = data.ISS(use_test_data=False)
fov = experiment.fov()
primary_image = fov.get_image(FieldOfView.PRIMARY_IMAGES)
dots = fov.get_image('dots')

# View the dots overlaid on the sequencing images
viewer = display(primary_image)
viewer = display(stack=primary_image, viewer=viewer)
```

### DIFF
--- a/starfish/_display.py
+++ b/starfish/_display.py
@@ -128,7 +128,7 @@ def display(
         ImageStack to display
     spots : IntensityTable
         IntensityTable containing spot information that was generated from the submitted stack.
-    viewer : napari.components._viewer.model.Viewer
+    viewer : napari.Viewer
         Napari viewer to append the ImageStack and/or spots to. If None, creates a new viewer.
         Note: appending is only supported in interactive environments.
     project_axes : Optional[Set[Axes]]
--- a/starfish/_display.py
+++ b/starfish/_display.py
@@ -9,6 +9,12 @@ from starfish.imagestack.imagestack import ImageStack
 from starfish.intensity_table.intensity_table import IntensityTable
 from starfish.types import Axes, Features
 
+try:
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", module="vispy.visuals.isocurve", lineno=22)
+        from napari import Window, Viewer
+except ImportError:
+    Window, Viewer = None, None
 
 INTERACTIVE = not hasattr(__main__, "__file__")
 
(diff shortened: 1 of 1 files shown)


starfish/_display.py
```diff
@@ -108,6 +108,7 @@ def _spots_to_markers(intensity_table: IntensityTable) -> Tuple[np.ndarray, np.n
 def display(
         stack: Optional[ImageStack] = None,
         spots: Optional[IntensityTable] = None,
+        viewer=None,
```
kevinyamauchi: Is it possible to give a type hint here without importing `Viewer` from `napari`?

kevinyamauchi: It seems to me like we are trying to avoid importing napari at the top of _display.py to ease the dependency. Is this the case?

kne42: Sort of. The `try-except` block can be moved up to the top but I didn't at the time because the warning would've been generated every time on start-up.

I suppose you could instead put the following at the top:
```python
try:
    from napari import Window, Viewer
except ImportError:
    Window, Viewer = None, None
```

and replace the warning code at the bottom with:
```python
if Window is None or Viewer is None:
    warnings.warn(...)
    return
```

kevinyamauchi: Sounds good. I'll do this. Thanks!

Issue #921, opened 2018-12-31
[RFC] Cleanup of the starfish package organization


ambrosejcarr:
## Making starfish more intuitive

I am finding it a bit confusing to navigate _parts_ of the _starfish_ codebase. Specifically, I sometimes feel I have to work harder than I should to find the files that contain important classes or functions.

I've identified 2 things that cause confusion for me: 
1. inconsistent use of `__init__.py` files
2. inconsistent file naming

Below I describe the issues I'm having and suggest some solutions but would like to get some opinions from others, in particular @ttung and @shanaxel42 . 

## inconsistent use of `__init__.py` files. 
Our codebase uses `__init__.py` files to create packages, structure the import structure of the public API, and definition of classes and functions. 

From my googling, best practices of what to include span from "nothing, ever" to import statements, versions, and some important constants. Everything I've read _prohibits including code_. 

I would like to propose that we keep our existing import _structure_ (subject to change when we finalize the public API), but move code out of `__init__.py` modules wherever possible. 

Some examples of modules that currently contain code, which could be adjusted: 
```
starfish.data.__init__.py
starfish.config.__init__.py  # not sure about this one
starfish.experiment.builder.__init__.py
starfish.imagestack.__init__.py  # multiprocessing stuff, move to multiprocessing module? 
starfish.image.<all pipeline component base classes>
starfish.spots.<all pipeline component base classes>
```

It's not clear to me that I would find `write_experiment_json` in `starfish.config.builder.__init__.py` or `starfish.data.BaristaSeq` in `starfish.data.__init__.py`. 

## Inconsistent file naming: 
Some of the starfish code base uses separate files for key object classes. I like this and find it very clear. Other parts of the codebase are named in ways where it is less clear what is contained. The `starfish.imagestack` package has examples of both:

```
starfish/imagestack/
├── __init__.py
├── _mp_dataarray.py
├── dataorder.py
├── imagestack.py
├── indexing_utils.py
├── parser
│   ├── __init__.py
│   ├── _key.py
│   ├── _tiledata.py
│   └── tileset
│       ├── __init__.py
│       └── _parser.py
└── physical_coordinate_calculator.py
```

It is very clear from observation that one would find the `ImageStack` class in `starfish.imagestack.imagestack`

It is not clear that one would find the ImageStack API contract `TileSetData` in `starfish.imagestack.tileset._parser`

I propose that we go through the codebase and restructure the imports to make the paths more intuitive.

joshmoore:
[...]

PR #1095, opened 2019-03-19
Restructure the relationship between PipelineComponent and AlgorithmBase

ttung:
Restructure the relationship between PipelineComponent and AlgorithmBase

Rather than each specific PipelineComponent have a link to the AlgorithmBase that describes the contract that the algorithm must implement, each AlgorithmBase implementation points to the PipelineComponent that it serves.

The current arrangement forces all the algorithms to be defined and evaluated in the interpreter before the pipeline component is defined and evaluated.  If we want to create an algorithm that's only used in tests, this is not possible.

The new arrangement ensures that any implementation of a specific AlgorithmBase is registered with its corresponding PipelineComponent.

This adds explicit abstract classes to the AlgorithmBase class hierarchy to make it easier to detect what's an actual algorithm implementation.

Depends on #1094

Test plan: `make -j fast`


ttung:
This also partially addresses #921

starfish/pipeline/algorithmbase.py
```diff
@@ -1,17 +1,37 @@
...
 from starfish.imagestack.imagestack import ImageStack
 from starfish.intensity_table.intensity_table import IntensityTable
 from starfish.types import LOG
 from starfish.types._constants import STARFISH_EXTRAS_KEY
 from starfish.util.logging import LogEncoder
+from .pipelinecomponent import PipelineComponent
 
 
-class AlgorithmBaseType(type):
+DISALLOWED_BASES: Set[Type] = set()
```
shanaxel42: what's an example of a disallowed base?

shanaxel42: what's an example of a disallowed base?

ttung: Let me improve the documentation.  Basically, we want the subclasses of `FilterAlgorithmBase`, `RegistrationAlgorithmBase`, etc., and not the subclasses of `AlgorithmBase`.

ttung: I went with a different approach that removed the need for this.

PR #1097, opened 2019-03-19
Introduce Runnable

ttung:
Runnable is the base unit of a recipe.  Each runnable constitutes a PipelineComponent with a specific algorithm.  Constructor arguments that are FileProviders are loaded into memory and passed in.  Inputs to the `run()` method can also include FileProviders.  In all cases, FileProviders are associated with a file path or url.  The type is inferred from the typing parameters of the method (currently supported: ImageStack, IntensityTable, ExpressionMatrix, and Codebook).

Runnables will be wired together to constitute a pipeline recipe.

Test plan: Added tests to verify a simple Runnable, chained Runnables, and Runnables that has constructor arguments that are FileProviders.

Depends on #1095

### DIFF
--- a/starfish/recipe/filesystem.py
+++ b/starfish/recipe/filesystem.py
@@ -1,0 +1,101 @@
+import enum
+from typing import Any, Callable, Type
+
+from starfish.codebook.codebook import Codebook
+from starfish.expression_matrix.expression_matrix import ExpressionMatrix
+from starfish.imagestack.imagestack import ImageStack
+from starfish.intensity_table.intensity_table import IntensityTable
+from starfish.util.indirectfile import (
+    convert,
+    GetCodebook,
+    GetCodebookFromExperiment,
+    GetImageStack,
+    GetImageStackFromExperiment,
+)
+
+
+def imagestack_convert(indirect_path_or_url: str) -> ImageStack:
+    return convert(
+        indirect_path_or_url,
+        [
+            GetImageStack(),
+            GetImageStackFromExperiment(),
+        ],
+    )
+
+
+def codebook_convert(indirect_path_or_url: str) -> Codebook:
+    return convert(
+        indirect_path_or_url,
+        [
+            GetCodebook(),
+            GetCodebookFromExperiment(),
+        ],
+    )
+
+
+class FileTypes(enum.Enum):
+    """These are the filetypes supported as inputs and outputs for recipes.  Each filetype is
+    associated with the implementing class, the method to invoke to load such a filetype, and the
+    method to invoke to save back to the filetype.
+
+    The load method is expected to be called with a string, which is the file or url to load from,
+    and is expected to return an instantiated object.
+
+    The save method is expected to be called with the object and a string, which is the path to
+    write the object to.
+    """
+    IMAGESTACK = (ImageStack, imagestack_convert, ImageStack.export)
+    INTENSITYTABLE = (IntensityTable, IntensityTable.load, IntensityTable.save)
+    EXPRESSIONMATRIX = (ExpressionMatrix, ExpressionMatrix.load, ExpressionMatrix.save)
+    CODEBOOK = (Codebook, codebook_convert, Codebook.to_json)
+
+    def __init__(self, cls: Type, loader: Callable[[str], Any], saver: Callable[[Any, str], None]):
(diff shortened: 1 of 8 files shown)


starfish/recipe/filesystem.py
```diff
@@ -0,0 +1,32 @@
...
+
+def save(file_path: str, object: Any) -> None:
+    if isinstance(object, ImageStack):
+        object.export(file_path)
+    elif isinstance(object, IntensityTable):
+        object.save(file_path)
+    elif isinstance(object, ExpressionMatrix):
+        object.save(file_path)
+    elif isinstance(object, Codebook):
+        object.to_json(file_path)
```
shanaxel42: If we refactored every class to name their methods export/load could we avoid needing these methods? Just call object.load()/object.export(). I'm not sure if the refactor's worth it, but it would keep things consistent and avoid the need to add more cases if we add more object types.

starfish/recipe/runnable.py
```diff
@@ -0,0 +1,95 @@
...
+            elif isinstance(_input, str):
+                # it's a path.
+                annotation = signature.parameters[key].annotation
+                reformatted_inputs.append((_input, annotation))
+        self._inputs = reformatted_inputs
+
+    @staticmethod
+    def _get_actual_run_method_signature(run_method: Callable) -> inspect.Signature:
+        if hasattr(run_method, "__closure__"):
+            # it's a closure, probably because of AlgorithmBaseType.run_with_logging.  Unwrap to
```
shanaxel42: eeesh, that's annoying

starfish/test/recipe/test_runnable.py
```diff
@@ -0,0 +1,170 @@
...
+    file_input_types = runnable.file_input_types
+    assert len(file_input_types) == 1
+    assert file_input_types["should_be_codebook"] == Codebook
+
+
+def test_run():
+    """Verify that we can run a single runnable and get its result.
+    """
+    filter_runnable = Runnable(
+        "filter", "FakeFilterAlgorithm",
```
shanaxel42: why not just use a real filter class here?

starfish/test/recipe/test_runnable.py
```diff
@@ -0,0 +1,170 @@
...
+        "https://d2nhj9g34unfro.cloudfront.net/20181005/ISS-TEST/fov_001/hybridization.json",
+        multiplicand=.5,
+    )
+    result = filter_runnable.run({})
+    assert isinstance(result, ImageStack)
+
+    # pick a random part of the registered image and assert on it
```
shanaxel42:
```suggestion
    # pick a random part of the filtered image and assert on it
```

starfish/test/recipe/test_runnable.py
```diff
@@ -0,0 +1,170 @@
...
+
+    filter_runnable = Runnable(
+        "filter", "FakeFilterAlgorithm",
+        dependency_runnable,
+        multiplicand=2.0,
+    )
+    result = filter_runnable.run({dependency_runnable: result})
+    assert isinstance(result, ImageStack)
+
+    # pick a random part of the registered image and assert on it
```
shanaxel42:
```suggestion
    # pick a random part of the filtered image and assert on it
```

ttung:
Please avoid reviewing this for now.  I need to make some design changes to this.

starfish/recipe/runnable.py
ttung: Yeah. :(

starfish/test/recipe/test_runnable.py
ttung: It's easier b/c getting the expected results are easier. :)

starfish/test/recipe/test_runnable.py
ttung: Fixed

starfish/test/recipe/test_runnable.py
ttung: Fixed

ttung:
Ready for re-review :) 

If you want to see how this is used in practice, you can wait for the next PR for reviewing.

ttung:
I've expanded the documentation significantly for the `Runnable` class.  Please let me know if anything is unclear.

PR #1100, opened 2019-03-21
[easy] - adding make clean to docs build

shanaxel42:
So when we delete/rename/reorganize pipeline components the docs build doesn't freakout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ help-unit:
 ### DOCS #####################################################
 #
 docs-%:
-	make -C docs $*
+	make -C docs clean && make -C docs $*
 
 help-docs:
 	$(call print_help, docs-TASK, alias for 'make TASK' in the docs subdirectory)
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ all:	fast
 
 ### UNIT #####################################################
 #
-fast:	lint mypy fast-test docs-html
+fast:	lint mypy fast-test clean-docs docs-html
 
 lint:   lint-non-init lint-init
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -61,7 +61,7 @@ help-unit:
 ### DOCS #####################################################
 #
 docs-%:
-	make -C docs $*
+	make -C docs clean && make -C docs $*
```
joshmoore: This would [surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) me (and in case I was doing an incremental docs build probably freak _me_ out :smile:). Would:

```
-clean:
+clean: docs-clean
```
get you some of what you want? i.e. encourage folks to `make clean` more frequently (e.g. between branch changes and before refactorings)

shanaxel42: why would it surprise you? I think regenerating the docs each time we build is a good way to keep them up to date. The clean also needs to be a part of the travis build because even if I clean-docs in my local env. travis will still fail

joshmoore: >  I think regenerating the docs each time we build is a good 

One of benefits of sphinx is that it does incremental builds. If I were to see `make docs` trigger a full build each time, I would initially assume I was doing something wrong. I'd certainly learn to live with it, but I'd likely do that by no longer using the top level `docs` target and instead going into the docs directory to workaround the clean.

> The clean also needs to be a part of the travis build 

:+1: for having travis cleaning.

> even if I clean-docs in my local env, travis will still fail

I don't follow.

shanaxel42: okok I'll make a separate docs-clean command

PR #1107, opened 2019-03-22
Clean up the algorithm base classes and run() methods.

ttung:
1. Ensure that the algorithm base classes's `run()` method is consistent with the implementations.
2. Fix missing documentation.
3. Add `*args` to all the run methods such that the implementations' run methods can take additional parameters.

Test plan: `make -j fast lint mypy`

### DIFF
--- a/starfish/image/_filter/clip.py
+++ b/starfish/image/_filter/clip.py
@@ -59,8 +59,12 @@ class Clip(FilterAlgorithmBase):
         return image.clip(min=v_min, max=v_max)
 
     def run(
-            self, stack: ImageStack, in_place: bool=False, verbose: bool=False,
-            n_processes: Optional[int]=None
+            self,
+            stack: ImageStack,
+            in_place: bool=False,
+            verbose: bool=False,
+            n_processes: Optional[int]=None,
+            *args,
     ) -> ImageStack:
         """Perform filtering of an image stack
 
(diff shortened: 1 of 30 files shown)


starfish/image/_filter/clip.py
```diff
@@ -59,8 +59,12 @@ def _clip(image: Union[xr.DataArray, np.ndarray], p_min: int, p_max: int) -> np.
...
 
     def run(
-            self, stack: ImageStack, in_place: bool=False, verbose: bool=False,
-            n_processes: Optional[int]=None
+            self,
+            stack: ImageStack,
+            in_place: bool=False,
+            verbose: bool=False,
+            n_processes: Optional[int]=None,
+            *args,
```
shanaxel42: where would we actually use these extra *args? Should we pass them into image.clip as well?

shanaxel42: where would we actually use these extra *args? Should we pass them into image.clip as well?

starfish/image/_filter/linear_unmixing.py
```diff
@@ -12,7 +12,7 @@
 
 class LinearUnmixing(FilterAlgorithmBase):
 
-    def __init__(self, coeff_mat: np.ndarray, clip_method: int=1) -> None:
+    def __init__(self, coeff_mat: np.ndarray) -> None:
```
shanaxel42: was clip_method supposed to be deleted?

starfish/image/_filter/clip.py
ttung: This is to make the signature match the superclass.

starfish/image/_filter/linear_unmixing.py
ttung: It's unused.

ttung: @kevinyamauchi I think you added this.  Let me know if it's a problem.

PR #1111, opened 2019-03-26
Move all the tests into the module they are testing.

ttung:
For example, starfish/test/experiment/builder/* is getting moved to starfish/experiment/builder/test.

This simplifies navigation in the IDE, and also ensures we put the tests in the right place (c.f., all the filter tests in starfish/test/pipeline).

Test plan: `make -j fast lint mypy`


kne42:
just saw this but 😍 😍 😍

PR #1118, opened 2019-03-29
Intensity Table Concat Processing

shanaxel42:
This PR introduces the idea of processing a list of IntensityTables with an overlap strategy before concatenating them. It also explicitly adds the TAKE_MAX strategy described by @berl in which we compare overlapping intensity tables and remove spots from the one with less spots in the overlap.
 
The PR also includes unit tests for overlapping_util methods.  

NOTE: 
In an effort to make the code easier to understand I went with O(n^2) approach to finding overlaps within a list of IntensityTables (just comparing each on to each other one). But there is a O(nlogn) approach that involves sorting the list by x/y coordinates first. If we think we'll need to optimize this process for large lists I can refactor.


starfish/intensity_table/intensity_table.py
```diff
@@ -389,10 +394,29 @@ def from_image_stack(
...
     @staticmethod
-    def concatanate_intensity_tables(intensity_tables: List["IntensityTable"]):
-        # TODO VARY CONCAT LOGIC IF TILES OVERLAP
-        # This method is a starting point for handling tile overlap, right now
-        # it does a simple concat but people want other overlap logic implmented
+    def process_overlaps(its: List["IntensityTable"],
+                         overlap_strategy: OverlapStrategy
+                         ) -> List["IntensityTable"]:
+        """Find the overlapping sections between IntensityTables
+        and process them according to the give overlap strategy
```
ttung:
```suggestion
        and process them according to the given overlap strategy
```

you can probably unwrap this a bit as well.

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+from starfish.types import Coordinates, Features
+from starfish.types._constants import OverlapStrategy
+
+
+class Area:
+    """
+    Small class that defines rectangular area of physical space by
+    its bottom left and top right coordinates.
+    """
+    def __init__(self, min_x, max_x, min_y, max_y):
```
ttung:
```suggestion
    def __init__(self, min_x: Number, max_x: Number, min_y: Number, max_y: Number) -> None:
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+            return False
+        if (area1.max_y < area2.min_y) or (area1.min_y > area2.max_y):
+            return False
+        return True
+
+    @staticmethod
+    def find_intersection(area1: "Area", area2: "Area") -> Union[None, "Area"]:
+        """
+        Find the overlap area of two rectangles and return as new Area object.
+        If no overlap return none.
```
ttung:
```suggestion
        If the two rectangles do not overlap, return None.
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    def _overlap(area1: "Area", area2: "Area") -> bool:
+        """Return True if two rectangles overlap"""
+        if (area1.max_x < area2.min_x) or (area1.min_x > area2.max_x):
+            return False
+        if (area1.max_y < area2.min_y) or (area1.min_y > area2.max_y):
+            return False
+        return True
+
+    @staticmethod
+    def find_intersection(area1: "Area", area2: "Area") -> Union[None, "Area"]:
```
ttung:
```suggestion
    def find_intersection(area1: "Area", area2: "Area") -> Optional["Area"]:
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    Parameters
+    ----------
+    it: xr.DataArray
+        The xarray to modify
+    area: Area
+        The area to not include in the modified xarray
+
+    Returns
+    -------
+     xr.DataArray :
```
ttung:
```suggestion
    xr.DataArray :
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+                     min(xr2[Coordinates.Y.value]).data,
+                     max(xr2[Coordinates.Y.value]).data)
+        intersection = Area.find_intersection(area1, area2)
+        if intersection:
+            all_overlaps[(idx1, idx2)] = intersection
+    return all_overlaps
+
+
+def remove_area_of_xarray(it: xr.DataArray, area: Area) -> xr.DataArray:
+    """Return everything in the xarray defined OUTSIDE the input area
```
ttung: It would be good to explain the boundary condition.  It looks like spots on the line are removed.

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    return it.where((it.xc <= area.min_x)
+                    | (it.xc >= area.max_x)
+                    | (it.yc <= area.min_y)
+                    | (it.yc >= area.max_y),
+                    drop=True)
+
+
+def sel_area_of_xarray(it: xr.DataArray, area: Area) -> xr.DataArray:
+    """Return everything in the xarray defined WITHIN the input area
```
ttung: ditto re the boundary conditions.

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    Parameters
+    ----------
+    intersection_rect : Area
+        The area of physical overalap between two xarrays
+    it1 : xr.DataArray
+        The first overlapping xarray
+    it2 : xr.DataArray
+        The second overlapping xarray
+    """
+    # # compare to see which section has more spots
```
ttung:
```suggestion
    # compare to see which section has more spots
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+                    & (it.yc >= area.min_y)
+                    & (it.yc <= area.max_y), drop=True)
+
+
+def take_max(intersection_rect: Area,
+             it1: xr.DataArray,
+             it2: xr.DataArray
+             ):
+    """
+    Compare two overlapping xarrays and remove spots from whichever
```
ttung: you can probably unwrap this.

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    has less in the overlapping section.
+
+    Parameters
+    ----------
+    intersection_rect : Area
```
ttung: Why not calculate this internally?

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,162 @@
...
+        The area of physical overalap between two xarrays
```
ttung:
```suggestion
        The area of physical overlap between the two xarrays
```

starfish/test/test_overlap_utils.py
```diff
@@ -0,0 +1,157 @@
...
+from starfish.types._constants import OverlapStrategy
+from starfish.util.overlap_utils import (
+    Area,
+    find_overlaps_of_xarrays,
+    remove_area_of_xarray,
+    sel_area_of_xarray
+)
+
+
+def create_intenisty_table_with_coords(area: Area, n_spots=10):
```
ttung:
```suggestion
def create_intensity_table_with_coords(area: Area, n_spots: int=10) -> IntensityTable:
```

starfish/test/test_overlap_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    it1 = create_intenisty_table_with_coords(Area(min_x=0, max_x=2,
+                                                  min_y=0, max_y=2), n_spots=10)
+    it2 = create_intenisty_table_with_coords(Area(min_x=1, max_x=2,
+                                                  min_y=1, max_y=3), n_spots=20)
+
+    concatenated = IntensityTable.concatanate_intensity_tables(
+        [it1, it2], overlap_strategy=OverlapStrategy.TAKE_MAX)
+
+    # The overlap section hits half of the spots from each intensity table, 5 from it1
```
ttung: wait what?  if it hits 5 of the spots from it1, then shouldn't we get a total of 25 spots?

starfish/test/test_overlap_utils.py
ttung: I'm a bit concerned that this method only creates spots along a diagonal.  Are there any shortcomings from this approach?

starfish/util/overlap_utils.py
shanaxel42: I guess I just didn't want to calculate it twice since we'd already done it to find the overlap pairs. But you're right it might be better for find_overlaps_of_xarrays() to just output a list of overlapping pairs and for take_max to actually find the overlap.

shanaxel42: This does make my tests a little less useful since I was validating that find_overlaps_of_xarrays() finds the right overlap pairs AND the right overlapping area, and now it would just verify it finds the overlapping pairs. What do you think?

starfish/test/test_overlap_utils.py
shanaxel42: both the sel and remove_area_of_xarray methods are inclusive...so we get one spot in the comparison count and the concatenation...maybe this is wrong?

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,169 @@
...
+
+    Returns
+    -------
+    list[Tuple[int, int]] :
+        A list of tuples containing the indices of two overlapping
+        IntensityTables.
+
+    """
+    all_overlaps: List[Tuple[int, int]] = list()
+    for idx1, idx2 in itertools.combinations(range(len(xarrays)), 2):
```
ttung: so this is, as you pointed out, a n^2 operation.  but each operation requires scanning the table to find the min/max for the coordinates.  can you get some numbers from @ambrosejcarr for realistic FOV and spot counts, and build a set of tables that reflect that, and see what the perf is like?  we can likely significantly shrink the cost by precomputing a min-max for each xarray and reusing that.  it would also help inform whether we need to do the nlogn approach.

Finally, it might be worth trying to actually merge a large set of intensity tables, even if synthetic, to see if there are any performance implications to any of the xarray ops used during the merge.

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    Find all the overlap areas within a list of xarrays.
+
+    Parameters
+    ----------
+    xarrays : List[xr.DataArray]
+        The list of xarrays to find overlaps in.
+
+    Returns
+    -------
+    list[Tuple[int, int]] :
```
ttung:
```suggestion
    List[Tuple[int, int]] :
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        if Area._overlap(area1, area2):
+            return Area(max(area1.min_x, area2.min_x),
+                        min(area1.max_x, area2.max_x),
+                        max(area1.min_y, area2.min_y),
+                        min(area1.max_y, area2.max_y))
+        return None
+
+
+def find_overlaps_of_xarrays(xarrays: List[xr.DataArray]) -> List[Tuple[int, int]]:
```
ttung: Not super important, but when typing with containers, it's usually advisable to type with the least specific container type that suits your requirements.  Rationale: it allows you to change your backing data structure without changing type everywhere.  In the java world, it's considered good practice to declare variables as List<> even if you assign to it a LinkedList<> or an ArrayList<>.

In this case, I think `xarrays` would be `Sequence[xr.DataArray]` or `Iterable[xr.DataArray]`, and the return type is probably also a sequence or an iterable (depending on if `len()` is required).

starfish/intensity_table/intensity_table.py
```diff
@@ -389,10 +394,29 @@ def from_image_stack(
         return IntensityTable.from_spot_data(intensity_data, pixel_coordinates)
 
     @staticmethod
-    def concatanate_intensity_tables(intensity_tables: List["IntensityTable"]):
-        # TODO VARY CONCAT LOGIC IF TILES OVERLAP
-        # This method is a starting point for handling tile overlap, right now
-        # it does a simple concat but people want other overlap logic implmented
+    def process_overlaps(its: List["IntensityTable"],
```
ttung: nit: would consider expanding `its` to `intensity_tables`

starfish/intensity_table/intensity_table.py
```diff
@@ -389,10 +394,29 @@ def from_image_stack(
...
+                         ) -> List["IntensityTable"]:
+        """Find the overlapping sections between IntensityTables and process them according
+        to the given overlap strategy
+        """
+        overlap_pairs = find_overlaps_of_xarrays(its)
+        for indices in overlap_pairs:
+            overlap_method = OVERLAP_STRATEGY_MAP[overlap_strategy]
+            idx1, idx2 = indices
+            # modify IntensityTables based on overlap strategy
+            it1, it2 = overlap_method(its[idx1], its[idx2])
```
ttung: I think this strategy might break down where you have three intensity tables that overlap in one area.  It might be easier to illustrate this if I'm not making sense.

starfish/intensity_table/intensity_table.py
```diff
@@ -389,10 +394,29 @@ def from_image_stack(
...
+            # replace IntensityTables in list
+            its[idx1] = it1
+            its[idx2] = it2
+        return its
+
+    @staticmethod
+    def concatanate_intensity_tables(intensity_tables: List["IntensityTable"],
+                                     overlap_strategy: OverlapStrategy = None):
```
ttung:
```suggestion
                                     overlap_strategy: Optional[OverlapStrategy] = None):
```

starfish/intensity_table/intensity_table.py
```diff
@@ -9,7 +9,12 @@
 
 from starfish.expression_matrix.expression_matrix import ExpressionMatrix
 from starfish.types import Axes, DecodedSpots, Features, LOG, SpotAttributes, STARFISH_EXTRAS_KEY
+from starfish.types._constants import OverlapStrategy
```
ttung: consider importing `OverlapStrategy` into `starfish.types.__init__`, especially if you foresee users actually using this enum.   If it's going to be for internal use only, this is okay.

starfish/test/test_overlap_utils.py
ttung: I would dump the table to make sure it is consistent with your understanding, though I suspect you are correct. :)

starfish/test/test_overlap_utils.py
```diff
@@ -0,0 +1,153 @@
...
+                                                  min_y=0, max_y=2), n_spots=10)
+    it2 = create_intensity_table_with_coords(Area(min_x=1, max_x=2,
+                                                  min_y=1, max_y=3), n_spots=20)
+
+    concatenated = IntensityTable.concatanate_intensity_tables(
+        [it1, it2], overlap_strategy=OverlapStrategy.TAKE_MAX)
+
+    # The overlap section hits half of the spots from each intensity table, 5 from it1
+    # and 10 from i21. It2 wins and the resulting concatenated table should have all the
+    # spots from it2 (20) and 6 (one on the boarder) from it1 (6) for a total of 26 spots
```
ttung:
```suggestion
    # spots from it2 (20) and 6 (one on the border) from it1 (6) for a total of 26 spots
```

starfish/util/overlap_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    Small class that defines rectangular area of physical space by
+    its bottom left and top right coordinates.
+    """
+    def __init__(self, min_x: Number, max_x: Number, min_y: Number, max_y: Number):
+        self.min_x = min_x
+        self.max_x = max_x
+        self.min_y = min_y
+        self.max_y = max_y
+
+    def __eq__(self, other):
```
ttung:
```suggestion
    def __eq__(self, other) -> bool:
```

PR #1119, opened 2019-03-30
Fixes for spots in starfish.display

ambrosejcarr:
* Add ability of starfish.display to adjust z-size of spots to account for anisotropy
* Enable n-d spots in starfish.display
* Change default display colors to gray colormap + red spots

### DIFF
--- a/starfish/_display.py
+++ b/starfish/_display.py
@@ -107,6 +107,10 @@ def _spots_to_markers(intensity_table: IntensityTable) -> Tuple[np.ndarray, np.n
     coords[:, 4] = np.repeat(intensity_table[Features.AXIS][Axes.ZPLANE.value].values, code_length)
 
     sizes = np.repeat(intensity_table.radius.values, code_length)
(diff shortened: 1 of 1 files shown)


starfish/_display.py
```diff
@@ -246,9 +253,12 @@ def display(
             return viewer
 
         coords = coords[mask, :]
-        sizes = sizes[mask]
+        sizes = sizes[mask, :]
```
kne42: shouldn't this be equivalent to `sizes = sizes[mask]`?

ambrosejcarr: In this regime, sizes is a (spots, 5) matrix, just like coords.  Can you point me at `napari` tests? I'd like to ideally add some to test various edge cases (2d, 3d, pseudo-3d where `z` is a singleton)

kne42: I mean those two lines behave exactly the same, assuming `sizes` is of the right shape; just don't see the need for the explicit `, :`. There are no napari tests for high-level things like that other than creating a script and looking at it yourself so that's just what I would do.

PR #1123, opened 2019-04-02
Refactor docs to support sphinx-gallery

ambrosejcarr:
* Converts the 3d smFISH notebook for use with sphinx gallery. 
* Updates the notebook text. 

Testing strategy: 
View docs here (click through data processing examples)
https://spacetx-starfish.readthedocs.io/en/ajc-sphinx-gallery/usage/index.html#data-processing-examples

@kevinyamauchi @berl @neuromusic @dganguli from a user perspective, is this helpful? It's certainly way easier from a development perspective. 

If `sphinx-gallery` works, I'll transition our existing notebooks over while updating the text, and will write the tutorials up in the same form.

I will figure out how to add plots, but @kne42 made that sound like it might be a tricky (although one-time) thing to set up.


kevinyamauchi:
This works for me! A few thoughts:

- Is it possible to create a table of contents at the top that links to the different sections (e.g., "Dowloading Data", "Constructing a Pipeline")? The page is very long, so it doesn't feel very "browseable".
- I think there needs to be something at the end the users can compare their results to (e.g., image of detected spots, detected spots/gene table). Currently, it is difficult for the user to know if it "worked" after working through the example.
- If you are planning to update the text, I think it would be helpful to make the language in the text descriptions and scripts match (e.g., use class/method names). Currently, I think it is a bit difficult to map the descriptions to the corresponding lines of code.

neuromusic:
This looks great @ambrosejcarr.

Moving forward, I suspect that it is not ONLY the full pipelines that we want in sphinx gallery... each item in a step-by-step tutorial could be implemented in sphinx gallery, along with any other demos or vignettes we want.

To support this, I'd like to propose the following:
- keep the gallery_dir at the top level `gallery`
- for gallery examples that we want to be part of a "set" like the data processing examples, add sphinx targets at the top of the gallery script
- rather than a toctree to navigate the data processing examples, use references to manually create the list of examples

.travis.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       script: make docker
     - name: 3d smFISH Notebook
       if: type = push and branch =~ /^(master|merge)/
-      script: make install-dev run__notebooks/py/3d_smFISH.py
+      script: make install-dev && ipython docs/source/_static/data_processing_examples/3d_smFISH.py
```
ttung: Do you care that we don't run this with the `run-notebooks` rule?  I suspect we want to because run-notebooks is run as part of the release process.

neuromusic:
> Do we have a short-term strategy for building these since it will time out on readthedocs?

1. For sphinx, leave the data processing demo scripts as they are here. Sphinx will render them (but not the outputs) in HTML. This will happen in << 10 min. Users can download the notebook that sphinx generates
2. Use travis execute each of the scripts in the sphinx gallery folder at the same cadence as we currently run the notebooks

If we want the sphinx to render the outputs in HTML, we rename the scripts to `plot_<script_name>.py` and sphinx will execute them when it builds the docs. This is the part that will take more than 10 minutes.

PR #1124, opened 2019-04-02
Indirect File click types

ttung:
Add the ability to reference an imagestack or a codebook through a set of conversion recipes.  Each conversion recipe can read the input given on the CLI and make an attempt to convert it to the desired type.

For instance, a primary image can be referred to by the json describing the slicedimage set, _or_ @experiment.json[my_fov_name][primary].

### DIFF
--- a/starfish/util/click/indirectparams.py
+++ b/starfish/util/click/indirectparams.py
@@ -1,0 +1,63 @@
+import abc
+from typing import Generic, Iterable, TypeVar
+
+from starfish.codebook.codebook import Codebook
+from starfish.imagestack.imagestack import ImageStack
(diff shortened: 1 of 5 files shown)


ttung:
Here's an example of how this works:

ttung:
I refactored this a bit to separate out the conversion logic from the click logic.  Rationale: I want to be able to use this in the pipeline recipes.  However, the recipes will not be able to use the click params, because at the time we generate the `click.command` calls, we do not know the types!  Only when we parse the recipe do we know what type they will be.

I will wait for this to be green before landing, but if you want to do a post-merge review, please do! :)

Issue #1116, opened 2019-03-27
`write_experiment_json()` doesn't handle non-flat experiment directory structures correctly


njmei:
Hi all,

So I've encountered an issue when running `write_experiment_json()` in 'no copy' mode (see: 
https://github.com/spacetx/starfish/pull/1053)

The resulting output .json files that starfish produces all assume that image tiles reside in the experiment root directory (see: https://github.com/spacetx/slicedimage/blob/3456f86a10cc9c89996a394cb75a03457a9bf6bc/slicedimage/io.py#L303).

However, our data (and likely many other group's data) do not live in a flat directory structure. Ours happens to look like:

```
expt_root_dir
| - Round_1_dir
|    | - FOV_1_dir
|    |   | - channel_1_z_1.tiff
|    |   ...
|    | - FOV_2_dir
|    |   | - channel_1_z_1.tiff
|    |   ...
|    ...
| - Round_2_dir
|   | - FOV_1_dir
|   |    | - channel_1_z_1.tiff
|   |    ...
|   ...
...
```

Although this issue arises because Starfish can no longer dictate a flat data structure (by forcing copying), copying **has** to be optional as the overhead of creating a new copy of huge 'local' datasets is unreasonable.

From an end-user perspective, having image data organized into subdirectories is also much easier to deal with than having to deal with one folder with thousands of tiffs.

berl:
@ttung should this be linked to an issue in `slicedimage` or is this the right place?  The 'no copy' mode plus this directory fix will give us the ability to create `SpaceTx-format` datasets from data already present on disk.

ttung:
Are auxiliary images mixed in with your primary images or are they separately stored?

ttung:
Are auxiliary images mixed in with your primary images or are they separately stored?

Also, the `partition_path_generator` option of `Writer.write_to_path` might solve your problems.

njmei:
@ttung Thanks for the reply Tony! It's very likely that our primary images will be mixed with our aux images.

I'll take a look at the `partition_path_generator` parameter. There wouldn't happen to be any documentation for that would there?

ttung:
With this patch, I can get FOVs to be written to their individual directories

```
[tt (.venv)]:~/microscopy/starfish:tonytung-default-shape> hg diff
diff -r 891395f82e10 starfish/experiment/builder/__init__.py
--- a/starfish/experiment/builder/__init__.py   Fri Mar 29 14:39:38 2019 -0700
[...]

PR #1131, opened 2019-04-03
Test to verify that inplace experiment construction works.

ttung:
This should nail the coffin on #1116, though there may still be usability improvements we can deliver.

Depends on #1127


ttung:
cc: @njmei

PR #1137, opened 2019-04-05
Change all ImageStack and Codebook cli params to use the click params

ttung:
#1124 introduced new click param types that directly return an ImageStack or a Codebook.  This refactors the existing cli params to use those types.

Test plan: `make -j fast`

Depends on #1124


starfish/image/_filter/_base.py
```diff
@@ -5,7 +5,7 @@
 from starfish.pipeline.algorithmbase import AlgorithmBase
 from starfish.pipeline.pipelinecomponent import PipelineComponent
 from starfish.util import click
-
+from starfish.util.click.indirectparams import ImageStackParamType
```
shanaxel42: Did I miss the PR where ImageStackParamType gets defined?

ttung: https://github.com/spacetx/starfish/pull/1124

PR #1140, opened 2019-04-05
[easy] Move tests that are not full pipeline tests into their respective packages

ttung:
Test plan: travis


ambrosejcarr:
lol, you should probably tag the rest of us on these types of PRs so we know about them. 😆

PR #1146, opened 2019-04-08
This is what black does to our codebase

ambrosejcarr:
Here's what the black linter would change about starfish. 

This is an informational PR, not for merging.


ttung:
I generally like it.  It's a bit newline happy than I am, but I can live with it.  @kne42 @shanaxel42, any thoughts?

shanaxel42:
Why are we considering this?

kne42:
oh I hate`black`. I get that it's opinionated; it just has the wrong opinions :)

PR #1151, opened 2019-04-08
Use a threadpoolexecutor to load data.

ttung:
This seems to speed things up, especially when we are doing network IO.

Depends on #1156


ttung:
There are some issues with this.  I'm seeing some weird pandas issues, which makes me suspect pandas is not thread-safe.

ambrosejcarr:
https://pandas.pydata.org/pandas-docs/stable/user_guide/gotchas.html#thread-safety

ttung:
I'm going to close this until I can figure out how to make it thread-safe. :)

PR #1159, opened 2019-04-09
Demonstrate that starfish outputs can be used in R

### LINKED ISSUE #841: Verify that all output formats are easily loaded into R

ambrosejcarr:
Adds documentation demonstrating how to use starfish outputs in R

* docs for IntensityTable and DecodedSpots
* adds example data files and README.rst declaring file contents

Review strategy: 

I think the reviewer probably needs to fire up R and verify this works. There's no R CI on starfish. 

closes #841


docs/source/help_and_reference/working_with_starfish_outputs/index.rst
```diff
@@ -0,0 +1,60 @@
...
+    example_netcdf_file: str = "docs/source/_static/example_data_files/decoded.nc"
+    intensity_table: starfish.IntensityTable = starfish.IntensityTable.load(example_netcdf_file)
+    print(intensity_table)
+
+in R_, the ncdf4 library allows the :code:`.nc` archive, which is based on hdf5, to be opened.
+It will contain a number of variables, each of which can be accessed by name:
+
+.. code-block:: R
+
+    install.packages("ncdf4")
```
dganguli: this fails out for me 

```
> install.packages("ncdf4")
Installing package into ‘/usr/local/lib/R/3.4/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://cran.rstudio.com/src/contrib/ncdf4_1.16.1.tar.gz'
Content type 'application/x-gzip' length 121046 bytes (118 KB)
==================================================
downloaded 118 KB

* installing *source* package ‘ncdf4’ ...
** package ‘ncdf4’ successfully unpacked and MD5 sums checked
configure.ac: starting
checking for nc-config... no
-----------------------------------------------------------------------------------
Error, nc-config not found or not executable.  This is a script that comes with the
netcdf library, version 4.1-beta2 or later, and must be present for configuration
to succeed.
 
If you installed the netcdf library (and nc-config) in a standard location, nc-config
should be found automatically.  Otherwise, you can specify the full path and name of
the nc-config script by passing the --with-nc-config=/full/path/nc-config argument
flag to the configure script.  For example:
 
./configure --with-nc-config=/sw/dist/netcdf4/bin/nc-config
 
Special note for R users:
-------------------------
To pass the configure flag to R, use something like this:
 
R CMD INSTALL --configure-args="--with-nc-config=/home/joe/bin/nc-config" ncdf4
 
where you should replace /home/joe/bin etc. with the location where you have
installed the nc-config script that came with the netcdf 4 distribution.
```

dganguli: do i need to install a netcdf library to use starfish?

ambrosejcarr: I believe Starfish installs netcdf4 for the user. 

You don't _need_ netcdf; it's also possible to open the `.nc` files with `rhdf5`. I'll figure out the solution to this before I merge.

ambrosejcarr:
> [...] However, we can separate these two elements into two csv files. One for spots, and one for intensities. 

The "spots csv" is essentially the decoded spots table that's linked in this PR. The intensities are not yet provided as a separate csv, in part because they're 3-dimensional. Do you have a sense of how to best represent those data? We're using `.nc` as a convenience, but as you point out, there may be better ways.

PR #1160, opened 2019-04-10
Recipe runner CLI entry point

### LINKED ISSUE #311: make a local pipeline runner
Starfish users will expect to be able to process multiple FOVs efficiently, but at the moment we are focusing on single FOVs. 

The simplest multi-field of view orchestration would be to process them sequentially with a for loop. More sophisticated would be a multi-processing based approach. Finally, we will want to link with the HCA cromwell/wdl architecture. 

Design an API to enable these kinds of multi-field of view processing. This will interact with #310 and depends on #302

ttung:
1. Add the recipe runner CLI entry point.
2. Add a recipe for the ISS pipeline and a test that exercises the recipe.

Test plan: `pytest starfish/test/full_pipelines/recipe`

Depends on #1192 

Fixes #311

### DIFF
--- a/starfish/recipe/cli.py
+++ b/starfish/recipe/cli.py
@@ -1,0 +1,28 @@
+from slicedimage.io import resolve_path_or_url
+
+from starfish.config import StarfishConfig
+from starfish.starfish import starfish
+from starfish.util import click
+from .recipe import Recipe
+
+
+@starfish.command("recipe")
+@click.option("--recipe", required=True, type=str, metavar="RECIPE_PATH_OR_URL")
+@click.option(
+    "--input", type=str, multiple=True, metavar="INPUT_FILE_PATH_OR_URL",
+    help="input file paths or urls to map to the recipe input parameters")
+@click.option(
+    "--output", type=str, multiple=True, metavar="OUTPUT_FILE_PATH",
+    help="output file paths to write recipe outputs to")
+@click.pass_context
+def run_recipe(ctx, recipe, input, output):
+    """Runs a recipe with a given set of inputs and outputs."""
+    config = StarfishConfig()
+
+    backend, relativeurl, _ = resolve_path_or_url(
+        recipe, backend_config=config.slicedimage)
+    with backend.read_contextmanager(relativeurl) as fh:
+        recipe_str = fh.read()
+
+    recipe_obj = Recipe(recipe_str, input, output)
+    recipe_obj.run_and_save()
--- a/starfish/test/full_pipelines/recipe/iss_recipe.txt
+++ b/starfish/test/full_pipelines/recipe/iss_recipe.txt
@@ -1,0 +1,55 @@
+primary_image = file_inputs[0]
+dots = file_inputs[1]
+nuclei = file_inputs[2]
+codebook = file_inputs[3]
+
+max_proj = compute(
+    Filter.MaxProject,
+    primary_image,
+    dims=['c', 'z'])
+
+transformation_list = compute(
+    LearnTransform.Translation,
+    max_proj,
+    reference_stack=dots, upsampling=1000, axes=Axes.ROUND)
+
+transformed = compute(
+    ApplyTransform.Warp,
+    primary_image,
+    transformation_list)
+
+filtered_primary = compute(
+    Filter.WhiteTophat,
+    transformed,
+    masking_radius=15)
+
+filtered_nuclei = compute(
+    Filter.WhiteTophat,
+    nuclei,
+    masking_radius=15)
+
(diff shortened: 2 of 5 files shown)


starfish/recipe/cli.py
```diff
@@ -0,0 +1,29 @@
...
+@starfish.command("recipe")
+@click.option("--recipe", required=True, type=str, metavar="RECIPE_PATH_OR_URL")
+@click.option(
+    "--input", type=str, multiple=True, metavar="INPUT_FILE_PATH_OR_URL",
+    help="input file paths or urls to map to the recipe input parameters")
+@click.option(
+    "--output", type=str, multiple=True, metavar="OUTPUT_FILE_PATH",
+    help="output file paths to write recipe outputs to")
+@click.pass_context
+def run_recipe(ctx, recipe, input, output):
```
ambrosejcarr: This needs usage examples. It's also not clear what any of the parameters must contain for the run to be valid. That should probably be documented here.

starfish/recipe/cli.py
```diff
@@ -0,0 +1,29 @@
...
+    """Runs a recipe with a given set of inputs and outputs."""
+    config = StarfishConfig()
+
+    backend, relativeurl, _ = resolve_path_or_url(
```
ambrosejcarr: General style question: We have a bunch of custom classes, and in this case I needed to follow 3 wrapper functions  in slicedimage.io to find out that this `backend` is a subclass of `slicedimage.backends.Backend`. 

Since we have type hints, I would find it helpful to use them in the cases of our custom classes, so that developers have a better sense of what the code is doing. 

In this case: 

```python
    backend: slicedimage.io.Backend
    relativeurl: str
    backend, relativeurl, _ = resolve_path_or_url(
        recipe, backend_config=config.slicedimage)


```

starfish/recipe/cli.py
```diff
@@ -0,0 +1,29 @@
+from slicedimage.io import resolve_path_or_url
+
+from starfish.config import StarfishConfig
+from starfish.starfish import starfish
+from starfish.util import click
+from .recipe import Recipe
+
+
+@starfish.command("recipe")
+@click.option("--recipe", required=True, type=str, metavar="RECIPE_PATH_OR_URL")
```
ambrosejcarr: I wonder why we have a public `run` method now that we can run recipes and you've shown this to be faster.

starfish/test/full_pipelines/recipe/iss_recipe.txt
```diff
@@ -0,0 +1,54 @@
+primary_image = file_inputs[0]
```
ambrosejcarr: This is great. We need to doc how this works and what your mini-language can do.

starfish/test/full_pipelines/recipe/iss_recipe.txt
```diff
@@ -0,0 +1,54 @@
...
+
+target_assignment = compute(
+    "target_assignment", "Label",
+    segmentation, spots)
+
+decoded = compute(
+    "decode", "PerRoundMaxChannelDecoder",
+    target_assignment, codebook)
+
+file_outputs[0] = decoded
```
shanaxel42: does this mean you only output once at the end of the pipeline? 
That would definitely explain why this process is way faster the exporting and importing after each step

starfish/test/full_pipelines/recipe/iss_recipe.txt
ttung: #1245

starfish/recipe/cli.py
ttung: For API users.

starfish/recipe/cli.py
ttung: Sure, I can do this.

ttung: Correction: `Backend` was never exposed.

I will correct this.

ttung: https://github.com/spacetx/starfish/issues/1246

starfish/recipe/cli.py
ttung: It is recipe dependent.  Please let me know if #1245 is adequate.

starfish/test/full_pipelines/recipe/iss_recipe.txt
ttung: Yes, that is true.  The speed improvements (IMO) is mostly from not starting up the starfish process repeatedly.

ttung:
I'm going to land this, free free to leave any post-land comments, @ambrosejcarr / @kne42

PR #1169, opened 2019-04-11
Clean up the giant mess of factories.

ttung:
Clean up the giant mess of factories.

End result:

1. Each object (imagestack, intensity table, and codebook) have their own test factory modules.  They produce objects of their respective types.  For example, starfish/codebook/test/factories.py produces `Codebook` objects.
2. Synthetic factories that produce multiple objects live in starfish/test/factories.py
3. Remove a lot of `pytest.fixtures`.  They complicate comprehension and static analysis tools.
4. Remove unused test fixtures.  Believe me, there was a bunch.
5. Got rid of `test_synthetic_spot_creation_produces_an_imagestack`, as that's something that static type analysis should already do.


starfish/imagestack/test/factories.py
```diff
@@ -0,0 +1,120 @@
...
+        range(num_ch),
+        range(num_z),
+        tile_fetcher,
+    )
+    tileset = list(collection.all_tilesets())[0][1]
+
+    return ImageStack.from_tileset(tileset)
+
+
+def imagestack_with_coords_factory(stack_shape: OrderedDict, coords: OrderedDict) -> ImageStack:
```
ambrosejcarr: [Future] Should we work towards removing objects without coordinates? I imagine our code should all be tested for robustly dealing with and propagating coordinates.

ttung: I do think we should probably move towards one factory that builds very realistic examples.  By putting all the factories together, I'm hoping we can start seeing the common stuff and avoid creating more. :)

PR #1173, opened 2019-04-11
Dg docs merfish notebook

dganguli:
Generally cleaned up the text and visualizations in this notebook to make this more user friendly. Test plan: this should pass CI before merging. And someone should read the text I wrote to make sure there's nothing obviously wrong.


notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+# EPY: END markdown
+
+# EPY: START code
+from starfish.image import Filter
+ghp = Filter.GaussianHighPass(sigma=3)
+high_passed = ghp.run(imgs, verbose=True, in_place=False)
+# EPY: END code
+
+# EPY: START markdown
+#The below algorithm deconvolves out the point spread function (PSF) introduced by the microcope. The goal of deconvolution is to enable the resolution of more spots, especially in high transcript density regions of the data. For this assay, the PSF is well approximated by a 2D isotropic gaussian with standard deviation (sigma) of 2. This The number of iterations (here 15) is an important parameter that needs careful optimization.
```
ambrosejcarr:
```suggestion
#The below algorithm deconvolves the point spread function (PSF) introduced by the microcope. The goal of deconvolution is to enable the resolution of more spots, especially in high transcript density regions of the data. For this assay, the PSF is well approximated by a 2D isotropic gaussian with standard deviation (sigma) of 2. This The number of iterations (here 15) is an important parameter that needs careful optimization.
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+glp = Filter.GaussianLowPass(sigma=1)
+low_passed = glp.run(deconvolved, in_place=False, verbose=True)
+# EPY: END code
+
+# EPY: START markdown
+#Image intensities vary across color channels and imaging rounds. We use the author's computed scale factors to appropriately scale the data to correct for this variation. Right now we have to extract this information from the metadata and apply this transformation manually.
+# EPY: END markdown
+
+# EPY: START code
+if use_test_data:
```
ambrosejcarr: We should fix the locations of these data so they are consistent and replace the code with the ElementWiseMultiply pipeline component. This can happen in a future PR (see #1182)

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+
+# EPY: START code
+single_plane_filtered = filtered_imgs.sel({Axes.ROUND: 0, Axes.CH: 0, Axes.ZPLANE: 0})
+single_plane_filtered = single_plane_filtered.xarray.squeeze()
+
+plt.figure(figsize=(10,10))
+plt.subplot(121)
+plt.imshow(single_plane, cmap='gray', clim = list(np.percentile(single_plane.data, [5, 99])))
+plt.axis('off')
+plt.title('Original data, Round:0, Channel: 0')
```
ambrosejcarr:
```suggestion
plt.title('Original data, Round: 0, Channel: 0')
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+plt.subplot(122)
+plt.imshow(single_plane_filtered, cmap='gray', clim = list(np.percentile(single_plane_filtered.data, [5, 99])))
+plt.title('Filtered data, Round:0, Channel: 0')
```
ambrosejcarr:
```suggestion
plt.title('Filtered data, Round: 0, Channel: 0')
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+plt.axis('off');
+# EPY: END code
+
+# EPY: START markdown
+### Decode the processed data into spatially resolved gene expression profiles
+#
+#Here, we decode each pixel value, across all rounds and channels into the corresponding target (gene) it corresponds too. Contiguous pixels that map to the same target gene are called as one RNA molecule. Intuitively, pixel vectors are matched to the codebook by computing the euclidean distance between the pixel vector and all codewords. The minimal distance gene target is then selected.
```
ambrosejcarr:
```suggestion
#Here, we decode each pixel value, across all rounds and channels, into the corresponding target (gene) it corresponds too. Contiguous pixels that map to the same target gene are called as one RNA molecule. Intuitively, pixel vectors are matched to the codebook by computing the euclidean distance between the pixel vector and all codewords. The minimal distance gene target is selected if it lies within `distance_threshold` of a code. 
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+
+initial_spot_intensities, prop_results = psd.run(filtered_imgs)
+
+spot_intensities = initial_spot_intensities.loc[initial_spot_intensities[Features.PASSES_THRESHOLDS]]
+# EPY: END code
+
+# EPY: START markdown
+### Compare to results from paper
+#
+#The below plot aggregates gene copy number across single cells in the field of view and compares the results to the published intensities in the MERFISH paper. Note that Starfish detects a lower number of transcripts than the authors' results. This can probably be improved by tweaking any of the above parameters.
```
ambrosejcarr:
```suggestion
#The below plot aggregates gene copy number across single cells in the field of view and compares the results to the published counts in the MERFISH paper. Note that Starfish detects a lower number of transcripts than the authors' results. This can likely be improved by tweaking the parameters of the algorithms above.
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+# EPY: END markdown
+
+# EPY: START code
+bench = pd.read_csv('https://d2nhj9g34unfro.cloudfront.net/MERFISH/benchmark_results.csv',
+                    dtype = {'barcode':object})
```
ambrosejcarr:
```suggestion
                    dtype={'barcode': object})
```

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+
+from starfish.util.plot import (
+    imshow_plane, intensity_histogram, overlay_spot_calls
+)
+# EPY: END code
+
+# EPY: START markdown
+### Load Data into Starfish from the Cloud
+#
+#The data from one field of view correspond to 16 images from 8 hybridzation rounds (r)  2 color channels (c) one z plane (z). Each image is 2048X2048 (y,x). These data correspond to U2OS cells in culture. O(100) transcripts are barcoded for subsequent spatial resolution. Pixel intensity values for one pixel, across all rounds and channels, can be decoded into the nearest barcode, thus resolving each pixel into a particular gene.
```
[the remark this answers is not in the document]
dganguli: yep. will tweak.

notebooks/py/MERFISH.py
dganguli: 👍

notebooks/py/MERFISH.py
```diff
@@ -0,0 +1,252 @@
...
+#A nice way to page through all this data is to use the ```display``` command. We have commented this out for now, because it will not render in Github. Instead, we simply show an image from the first round and color channel.
+# EPY: END markdown
+
+# EPY: START code
+# Display all the data in an interactive pop-up window. Uncomment to have this version work.
+
+# %gui qt5
+# display(imgs)
+
+# Display a single plane of data
```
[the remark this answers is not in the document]
dganguli: ah, first and foremost, i didn't quite understand how **kwargs were propagated from imshow_plane to the underlying call to im_show. second, not here but in other examples, I needed to pass a data driven call to clim, e.g., ```clim = list(np.percentile(single_plane.data, [5, 99])``` which happens after the imagestack is sliced into a numpy array. adding this functionality into ```imshow_plane``` would be awesome! the api should be more like percentiles=None, but if you specify these values it will compute the percentiles appropriately and wrap them up in the clim kwarg.

PR #1180, opened 2019-04-12
Dg docs dartfish notebook

dganguli:
Updates text around the Dartfish notebook


notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
 
-# EPY: START code
-print(stack.shape)
+print(imgs)
 # EPY: END code
 
+# EPY: START markdown
+### Visualize codebook
+#
+#The DARTFISH codebook maps pixel intensities across the rounds and channels to the corresponding barcodes and genes that those pixels code for. For this example dataset, the codebook specifies 96 possible barcodes.
```
ambrosejcarr:
```suggestion
#The DARTFISH codebook maps pixel intensities across the rounds and channels to the corresponding barcodes and genes that those pixels code for. For this example dataset, the codebook specifies 96 possible barcodes. The codebook used in this experiment has 3 color channels and one blank channel, each of which contribute to codes. The presence of the blank channel will be important later when the filtering is described.
```

notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
+plt.imshow(single_plane, cmap='gray', clim=list(np.percentile(single_plane, [1, 99.9])))
+plt.title('Round: 0, Chanel:0')
+plt.axis('off');
 # EPY: END code
 
 # EPY: START markdown
-#### Load copy number benchmark results
+### Filter and scale raw data before decoding into spatially resolved gene expression
+#
+#First, we normalize the intensity of each image by the maximum intensity of that image, which is equivalent to scaling by the 100th percentile value of the pixel values in each image.
```
ambrosejcarr:
```suggestion
#First, we equalize the intensity of the images by scaling each image by its maximum intensity, which is equivalent to scaling by the 100th percentile value of the pixel values in each image.
```

notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
 # EPY: START code
-cnts_benchmark = pd.read_csv('https://d2nhj9g34unfro.cloudfront.net/20181005/DARTFISH/fov_001/counts.csv')
-cnts_benchmark.head()
+sc_filt = Filter.ScaleByPercentile(p=100)
+norm_imgs = sc_filt.run(imgs)
 # EPY: END code
 
 # EPY: START markdown
-#### Filter Image Stack
+#Next, for each imaging round, and each pixel location, we zero out the values across all three color channels if the magnitude of this 3 vector is below a threshold. As such, the code letters in these rounds for these pixels will be 0
```
ambrosejcarr:
```suggestion
#Next, for each imaging round, and each pixel location, we zero the intensity values across all three color channels if the magnitude of this 3 vector is below a threshold. As such, the code value associated with these pixels will be the blank. This is necessary to support euclidean decoding for codebooks that include blank values. 
```

This approach feels statistically unprincipled to me. My interpretation of this method is that it's accounting for the fact that autofluorescence will bias the decoding towards non-blank channels, but this feels like it should be corrected through background removal or global intensity thresholding, not a magnitude-based zeroing. Do you recall why we didn't try those approaches when you were creating this? 

If you agree, I'd like to explore replacing this with existing filters as a follow-up.

notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
-norm_stack = sc_filt.run(stack)
-zero_norm_stack = z_filt.run(norm_stack)
+filtered_imgs = z_filt.run(norm_imgs)
 # EPY: END code
 
 # EPY: START markdown
-##### Visualize barcode magnitudes to help determine an appropriate threshold for decoding
+### Decode the processed data into spatially resolved gene expression profiles
+#
+#Here, we decode each pixel value, across all rounds and channels into the corresponding target (gene) it corresponds too. Contiguous pixels that map to the same target gene are called as one RNA molecule. Intuitively, pixel vectors are matched to the codebook by computing the euclidean distance between the pixel vector and all codewords. The minimal distance gene target is then selected.
```
ambrosejcarr:
```suggestion
#Here, starfish decodes each pixel value, across all rounds and channels, into the corresponding target (gene) it corresponds too. Contiguous pixels that map to the same target gene are called as one RNA molecule. Intuitively, pixel vectors are matched to the codebook by computing the euclidean distance between the pixel vector and all codewords. The minimal distance gene target is then selected, if it is within `distance_threshold` of any code. 
```

notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
+filtered_imgs = z_filt.run(norm_imgs)
 # EPY: END code
 
 # EPY: START markdown
-##### Visualize barcode magnitudes to help determine an appropriate threshold for decoding
+### Decode the processed data into spatially resolved gene expression profiles
+#
+#Here, we decode each pixel value, across all rounds and channels into the corresponding target (gene) it corresponds too. Contiguous pixels that map to the same target gene are called as one RNA molecule. Intuitively, pixel vectors are matched to the codebook by computing the euclidean distance between the pixel vector and all codewords. The minimal distance gene target is then selected. 
+#
+#This decoding operation requires some parameter tuning, which is described below. First, we look at a distribution of pixel vector barcode magnitudes to determine the minimum magnitude threshold at which we will attempt to decode the pixel vector.
```
ambrosejcarr: Can you say anything about how you select the threshold from the histogram?

notebooks/py/DARTFISH.py
```diff
@@ -139,60 +154,25 @@ def compute_magnitudes(stack, norm_order=2):
...
 
 # EPY: START markdown
-#### QC Plots
-# EPY: END markdown
-
-# EPY: START markdown
-##### parameter tuning plots
+### Compare to benchmark results
+#
+#The below plot aggregates gene copy number across single cells in the field of view and compares the results to the same copy numbers from the Matlab based pipeline. This can probably be improved by tweaking any of the above parameters.
```
ambrosejcarr:
```suggestion
#The below plot aggregates gene copy number across cells in the field of view and compares the results to the same copy numbers from the authors' pipeline. This can likely be improved by tweaking parameters in the above algorithms.
```

notebooks/py/DARTFISH.py
```diff
@@ -139,60 +154,25 @@ def compute_magnitudes(stack, norm_order=2):
...
-plt.xlabel('min distance to code');
-plt.vlines(min_dist, ymin=plt.gca().get_ylim()[0], ymax=plt.gca().get_ylim()[1])
-sns.despine(offset=2)
-plt.title('Set minimum distance threshold');
-# EPY: END code
-
-# EPY: START markdown
-##### Copy number comparisons
-# EPY: END markdown
+# load results from Matlab based pipeline
```
ambrosejcarr:
```suggestion
# load results from authors' pipeline
```

notebooks/py/DARTFISH.py
```diff
@@ -26,76 +28,89 @@
...
-sns.set_context('talk')
-sns.set_style('ticks')
 # EPY: END code
 
 # EPY: START markdown
-#### Load image stack
+### Load data into Starfish from the Cloud
 #
-#Note that the data here corresopond to DARTFISHv1 2017. The group is actively working on improving the protocol.
+#The example data here corresopond to DARTFISHv1 2017. The group is actively working on improving the protocol. The data represent human brain tissue from the human occipital cortex from 1 field of view (FOV) of larger experiment. The data from one field of view correspond to 9 images from 6 imaging rounds (r) 3 color channels (c) and 1 z-plane (z). Each image is 988x988 (y,x)
```
ambrosejcarr:
```suggestion
#The example data here corresopond to DARTFISHv1 2017. The group is actively working on improving the protocol. The data represent human brain tissue from the human occipital cortex from 1 field of view (FOV) of larger experiment. The data from one field of view correspond to 18 images from 6 imaging rounds (r) 3 color channels (c) and 1 z-plane (z). Each image is 988x988 (y,x)
```

notebooks/py/DARTFISH.py
dganguli: Many steps in many pipelines seem statistically unprincipled to me 🙃 My memory of this pipeline is fuzzy. I do vaguely remember thinking this wasn't that unreasonable of a pre-processing step. And looking at the actual data, I'm not so sure filtering would have helped much. The statistical properties of the images seem a bit odd to me.

PR #1181, opened 2019-04-14
Allow for intensity tables with labeled axes

### LINKED ISSUE #1168: Add test to verify that ImageStack with labeled axes -> IntensityTable works correctly

ttung:
Currently, we use the offset of r/c/z to label the axes on the intensity table.  We should instead be using the axes labels from the ImageStack.  This PR makes that change in three key areas:

1. `IntensityTable.empty_intensity_table` now accepts the labels for the axes.
2. `IntensityTable.from_spot_data` now accepts the labels for the axes and verifies that the number of labels matches the intensity data.
3. `IntensityTable.from_image_stack` reads the labels for the zplane axes and assigns them correctly.

Test plan: Add a test that instantiates a labeled ImageStack and derives an IntensityTable from it.  Also ran `make test && make -j run-notebooks`.

Depends on #1178
Fixes #1168

### DIFF
--- a/starfish/intensity_table/intensity_table.py
+++ b/starfish/intensity_table/intensity_table.py
@@ -1,6 +1,6 @@
 from itertools import product
 from json import loads
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Sequence, Union
 
 import numpy as np
 import pandas as pd
@@ -72,7 +72,9 @@ class IntensityTable(xr.DataArray):
 
     @staticmethod
     def _build_xarray_coords(
-            spot_attributes: SpotAttributes, channel_index: np.ndarray, round_index: np.ndarray
+            spot_attributes: SpotAttributes,
+            channel_values: Sequence[int],
+            round_values: Sequence[int]
     ) -> Dict[str, np.ndarray]:
         """build coordinates for intensity-table"""
         coordinates = {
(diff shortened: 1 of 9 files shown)


starfish/intensity_table/intensity_table.py
```diff
@@ -146,6 +155,10 @@ def from_spot_data(
             number of dimensions. If this argument is omitted, dimension names
             are taken from ``coords`` (if possible) and otherwise default to
             ``['dim_0', ... 'dim_n']``.
+        ch_values : Sequence[int]
+            The possible values for the channel number, in sequential order.
```
ambrosejcarr: what does "sequential order" mean? Does it mean they need to be numerically sorted, or match the input? I feel like we require the latter but not the former.

ttung: You're right.

ttung: Fixed.

starfish/intensity_table/intensity_table.py
```diff
@@ -365,10 +392,10 @@ def from_image_stack(
         assert crop_y * 2 < image_stack.shape['y']
         assert crop_x * 2 < image_stack.shape['x']
 
-        zmin = crop_z
+        zmin = image_stack.axis_labels(Axes.ZPLANE)[crop_z]
         ymin = crop_y
         xmin = crop_x
-        zmax = image_stack.shape['z'] - crop_z
+        zmax = image_stack.axis_labels(Axes.ZPLANE)[-crop_z - 1]
```
shanaxel42: `-crop_z - 1` this is weird, what value is this?

shanaxel42: -crop_z - 1 this is weird, what value is this?

ttung: So this is to allow for cropping from the maximum zlayer.  crop_z=0 => zmax=-1 (i.e., last plane).  crop_z=1 => zmax=-2 (i.e., second to last plane).

PR #1186, opened 2019-04-15
Add back ability to set number of processes in spot finders

ambrosejcarr:
Adds ability to set `n_processes` for 3/4 spot finders which were missing the flag.

### DIFF
--- a/starfish/spots/_detector/blob.py
+++ b/starfish/spots/_detector/blob.py
@@ -150,7 +150,7 @@ class BlobDetector(SpotFinderAlgorithmBase):
         blobs_axes : Optional[Tuple[Axes, ...]]
             If blobs_image is provided, blobs_axes must be provided as well.  blobs_axes represents
             the axes across which the blobs image is max projected before spot detection is done.
-        n_processes : Optional[int] = 0,
+        n_processes : Optional[int] = None,
             Number of processes to devote to spot finding.
         """
 
--- a/starfish/spots/_detector/blob.py
+++ b/starfish/spots/_detector/blob.py
@@ -133,6 +133,7 @@ class BlobDetector(SpotFinderAlgorithmBase):
             primary_image: ImageStack,
             blobs_image: Optional[ImageStack] = None,
             blobs_axes: Optional[Tuple[Axes, ...]] = None,
+            n_processes: Optional[int] = None,
             *args,
     ) -> IntensityTable:
         """
@@ -149,6 +150,8 @@ class BlobDetector(SpotFinderAlgorithmBase):
         blobs_axes : Optional[Tuple[Axes, ...]]
             If blobs_image is provided, blobs_axes must be provided as well.  blobs_axes represents
             the axes across which the blobs image is max projected before spot detection is done.
+        n_processes : Optional[int] = 0,
+            Number of processes to devote to spot finding.
         """
 
         intensity_table = detect_spots(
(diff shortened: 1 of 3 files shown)


starfish/spots/_detector/blob.py
```diff
@@ -149,6 +150,8 @@ def run(
         blobs_axes : Optional[Tuple[Axes, ...]]
             If blobs_image is provided, blobs_axes must be provided as well.  blobs_axes represents
             the axes across which the blobs image is max projected before spot detection is done.
+        n_processes : Optional[int] = 0,
```
dganguli: 0 seems like an unchill number of processes? did you mean 1? or None?

starfish/spots/_detector/local_max_peak_finder.py
```diff
@@ -289,13 +290,16 @@ def run(
         blobs_axes : Optional[Tuple[Axes, ...]]
             If blobs_image is provided, blobs_axes must be provided as well.  blobs_axes represents
             the axes across which the blobs image is max projected before spot detection is done.
+        n_processes : Optional[int] = 0,
```
dganguli: see comment above

starfish/spots/_detector/trackpy_local_max_peak_finder.py
```diff
@@ -157,13 +158,16 @@ def run(
         blobs_axes : Optional[Tuple[Axes, ...]]
             If blobs_image is provided, blobs_axes must be provided as well.  blobs_axes represents
             the axes across which the blobs image is max projected before spot detection is done.
+        n_processes : Optional[int] = 0,
```
dganguli: see comment above

starfish/spots/_detector/blob.py
ambrosejcarr:
```suggestion
        n_processes : Optional[int] = None,
```

starfish/spots/_detector/local_max_peak_finder.py
ambrosejcarr:
```suggestion
        n_processes : Optional[int] = None,
```

starfish/spots/_detector/trackpy_local_max_peak_finder.py
ambrosejcarr:
```suggestion
        n_processes : Optional[int] = None,
```

starfish/spots/_detector/blob.py
```diff
@@ -133,6 +133,7 @@ def run(
             primary_image: ImageStack,
             blobs_image: Optional[ImageStack] = None,
             blobs_axes: Optional[Tuple[Axes, ...]] = None,
+            n_processes: Optional[int] = None,
```
shanaxel42: I don't think you need the = None if it's an optional param

ttung: You do still need it.

Strictly speaking, `Optional` is optional, because mypy developers have decided to relax the rules for now.

ambrosejcarr: Can you say more / link to the documentation that defines the valid uses (or lack of uses) of `Optional`?

ttung: https://mypy.readthedocs.io/en/latest/kinds_of_types.html#no-strict-optional

Apparently the optional Optional is gone though, except in the case of a parameter that defaults to None.

PR #1188, opened 2019-04-15
Update BaristaSeq text, fix LinearUnmixing

ambrosejcarr:
This PR addresses scientific issues in BaristaSeq: 

* LinearUnmixing was doing the wrong thing (summing over the wrong axis, which led to incorrect unmixing)
* Updates text in BaristaSeq.py
* BaristaSeq is small enough that it can be rendered by readthedocs as it stands, so testing it there and removing the out-of-date .py and .ipynb

If we decide to keep checking in notebooks, this can be reversed by downloading the .ipynb file and checking it back in.


dganguli:
should un-delete the notebook for spacejam. pretty please?

PR #1189, opened 2019-04-16
Fix the restoration of non-indexed axes.

ttung:
When the selection removes a dimension, xarray.expand_dims does not expand the non-indexed dimensions that were removed.  For example, if one selects only a single zplane, it reduce the z physical coordinate to a coordinate scalar, and not an array of size 1.

When we detect that an axis was reduced out of existence, we ensure that the dependent axes are restored to arrays.

Test plan: Wrote a test that selects an ImageStack using a variety of selectors.  In each case, the length of each indexed dimension (i.e., X, Y, ZPLANE) must match its dependent dimension (i.e., XC, YC, ZC).

### DIFF
--- a/starfish/imagestack/indexing_utils.py
+++ b/starfish/imagestack/indexing_utils.py
@@ -2,6 +2,7 @@ from typing import Dict, Mapping, MutableMapping, Tuple, Union
 
 import numpy as np
 import xarray as xr
+from xarray.core.utils import is_scalar
 
 from starfish.types import Axes, Coordinates, Number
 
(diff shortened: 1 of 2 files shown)


ttung:
> I don't totally understand the issue? If you sel on value of the zplane then that results in the zc dimension only having a scalar value? But does that matter? You could still access that value with indexed.xarray['zc'].data

`ImageStack.export` indexes into `xarray['zc']` to retrieve the coordinates of each tile.  If it's a scalar, that operation does not work.

PR #1163, opened 2019-04-11
Introduce Runnable

ttung:
Runnable is the base unit of a recipe.  Each runnable constitutes a PipelineComponent with a specific algorithm.  Constructor arguments that are FileProviders are loaded into memory and passed in.  Inputs to the `run()` method can also include FileProviders.  In all cases, FileProviders are associated with a file path or url.  The type is inferred from the typing parameters of the method (currently supported: ImageStack, IntensityTable, ExpressionMatrix, and Codebook).

Runnables will be wired together to constitute a pipeline recipe.

Test plan: Added tests to verify a simple Runnable, chained Runnables, and Runnables that has constructor arguments that are FileProviders.

This is a continuation of #1097

PR #1117, opened 2019-03-28
Recipe and recipe execution

ttung:
Builds on top of #1097 to construct and run recipes with one or more runnables.

A recipe wires together one or more runnables.  Inputs to recipes are available to the recipe itself as the magic variable name "file_inputs".  A runnable is invoked by the magic function "compute".  Anything written to the magic variable "file_outputs" is matched and written out to a set of output paths.

Fixes #311

PR #1192, opened 2019-04-16
Recipe and recipe execution

### LINKED ISSUE #311: make a local pipeline runner

ttung:
Builds on top of #1163 to construct and run recipes with one or more runnables.

A recipe wires together one or more runnables.  Inputs to recipes are available to the recipe itself as the magic variable name "file_inputs".  A runnable is invoked by the magic function "compute".  Anything written to the magic variable "file_outputs" is matched and written out to a set of output paths.

This is a continuation of #1117
Fixes #311

### DIFF
--- a/starfish/recipe/recipe.py
+++ b/starfish/recipe/recipe.py
@@ -1,0 +1,234 @@
+from typing import (
+    AbstractSet,
+    Any,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+    Optional,
+    Sequence,
+    Set,
+)
+
+import numpy as np
+
+from starfish import image, spots
+from starfish.types import Axes, Coordinates
+from .filesystem import FileProvider, FileTypes
+from .runnable import Runnable
+
+
+class ExecutionComplete(Exception):
+    """Raised by :py:class:`Execution` when it is complete.  We don't rely on catching StopIteration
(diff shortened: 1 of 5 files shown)


starfish/recipe/recipe.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(
+            self,
+            recipe_str: str,
+            input_paths_or_urls: Sequence[str],
+            output_paths: Sequence[str],
+    ):
+        ordered_sequence = OrderedSequence()
+        file_outputs: MutableMapping[int, Runnable] = {}
+        recipe_scope = {
+            "Axes": Axes,
```
shanaxel42: Why do we need this info?

ttung:
> Looks great. Did I miss documentation of the deviations from python-style code (e.g. the @ symbol)?

you did.  that is a CLI-ism.  Not a recipe thing.

PR #1198, opened 2019-04-16
add documentation for use of conda environments

### LINKED ISSUE #704: Document use of environments with conda to reduce likelihood of user errors
Conda has a similar approach to virtual environments. We should specify that users of Conda should leverage this capability in our documentation and installation instructions. 

[https://conda.io/docs/user-guide/tasks/manage-environments.html](https://conda.io/docs/user-guide/tasks/manage-environments.html)

kne42:
add a section under the installation docs on how to set up virtual environments with conda

closes #704

### DIFF
--- a/docs/source/getting_started/installation/index.rst
+++ b/docs/source/getting_started/installation/index.rst
@@ -16,9 +16,10 @@ The output should look similar to this:
 Using virtual environments
 --------------------------
    
-Starfish names its dependencies and lists explicit versions, due to sensitivity to subtle algorithm
-changes.  For that reason, it is strongly encouraged, but not required, that you set up a
-virtualenv_. Create a work folder and set up the virtual environment like:
+Starfish lists minimum versions for its dependencies for access to new features and algorithms.
+These more up-to-date packages may create conflicts in your existing scripts or other packages,
+so you may want to set up a virtualenv_.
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
I haven't used conda in like 10 years. I'll try this tonight and let you know how I fare. :-)

docs/source/getting_started/installation/index.rst
```diff
@@ -13,11 +13,11 @@ The output should look similar to this:
    % python --version
    Python 3.6.5
 
-Installation for users
-----------------------
-
+Using virtual environments
+--------------------------
+   
 Starfish names its dependencies and lists explicit versions, due to sensitivity to subtle algorithm
```
ttung: > lists explicit versions

I don't think this is still true today.  There is an option to install the versions we use for CI, but it's no longer mandatory.

kne42: ah, that's a good point! do we maybe want to just convert this into a strict/nonstrict installation guide? cc @ambrosejcarr @neuromusic

ambrosejcarr: Yes, I would add both options to this.

ttung:
Landing!

PR #1200, opened 2019-04-16
Restructuring soc layout

shanaxel42:
Moved API to be a top level directory and data structures to be beneath it. Moved both from help and references. I think this makes more sense logically but let me know if not. 
Screen Shot 2019-04-16 at 4 25 00 PM


ttung:
👍

Issue #1208, opened 2019-04-17
De-clutter starfish namespace


ambrosejcarr:
Stuff that gets dropped into the starfish namespace when you import starfish right now:
 ```
 'Codebook',
 'Experiment',
 'FieldOfView',
 'ImageStack',
 'IntensityTable',
 'codebook',
 'compat',
 'config',
 'display',
 'errors',
 'experiment',
 'expression_matrix',
 'image',
 'imagestack',
 'intensity_table',
 'multiprocessing',
 'pipeline',
 'pkg_resources',
 'spacetx_format',
 'spots',
 'starfish',
 'types',
 'util',
 'warnings
```      

Of that list, I think relevant things to import are:
```
Codebook
Experiment
IntensityTable
FieldOfView
ImageStack
image
spots
display
types
```
I find the ambiguity of `experiment` and `Experiment` troubling, for example. 

In addition, some of these packages have no clear import path in `starfish/__init__.py` so it is not clear to me why they are found in the top-level namespace (e.g. `util, types, pipeline, multiprocessing`)

Can we clean this up by making some of our packages private? Is that the preferred method?

ambrosejcarr:
@ttung this seems like the kind of ticket you'd be least likely to mess up. Think this kind of clean-up would be possible pre-hackathon?

ttung:
So one way to do this would be to nest everything implementation-related in a `_impl` folder.  It clutters the codebase but might make for a nice UX.  There's probably other approaches, but I'm not super knowledgeable about this. :(

ambrosejcarr:
Ok. Let's get a third opinion. @kne42 ? :-)

kne42:
Sorry for the late reply, just got around to seeing this email notification.

So the general rule of thumb is:
- if a module contains only one *main* class/function, do not import it and import that object instead
- if a module contains only one public class/function, additionally make it private
- if a module will be heavily used, it can be imported
- objects that are not meant to be provided by a namespace should be deleted or not included in the package's `__all__` list

I don't really know the structure of `starfish` too well but what @ambrosejcarr suggests seems right to me.

PR #1209, opened 2019-04-17
[Easy] remove unused packages from top-level namespace. 

ambrosejcarr:
Remove some easy packages that shouldn't be in the package's top-level namespace

See: https://github.com/spacetx/starfish/issues/1208 for the harder stuff that we should also attend to.

### DIFF
--- a/starfish/__init__.py
+++ b/starfish/__init__.py
@@ -32,3 +32,5 @@ del get_versions
 
 if __name__ == "__main__":
     starfish()
+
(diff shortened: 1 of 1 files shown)


ttung:
I believe the canonical way to do this is to declare an __all__ list which enumerates the symbols you want to export.

Another valid approach is to import warnings and pkg_resources as underscored names, i.e.:

```
import pkg_resources as _pkg_resources
import warnings as _warnings
```

PR #1218, opened 2019-04-18
Normalize names for PipelineComponents

ttung:
The CLI name for PipelineComponent should match the API name.

Test plan: travis


starfish/spots/_target_assignment/_base.py
```diff
@@ -30,7 +22,7 @@ def _cli_run(cls, ctx, instance):
         assigned.to_netcdf(os.path.join(output))
 
     @staticmethod
-    @click.group(COMPONENT_NAME)
+    @click.group("TargetAssignment")
```
ambrosejcarr:
```suggestion
    @click.group("AssignTargets")
```

(and all dependent refactoring).

This is action oriented like the rest of the names.

ttung: done.

PR #1220, opened 2019-04-18
Clone a repo, not checkout.

ttung:
When starting up a dev environment, we should clone the github repo, not check it out.

Furthermore, we should clone from git:// rather than ssh://

Test plan: visual.


ttung:
Pushing. What could go wrong?  🤷‍♂️

PR #1223, opened 2019-04-19
RFC/WIP: document use of segmentation & target assignment

kne42:
please let me know how to make this example shorter/actually correct (expression matrix returned is null)

### DIFF
--- a/docs/source/_static/tutorials/exec_feature_identification_and_annotation.py
+++ b/docs/source/_static/tutorials/exec_feature_identification_and_annotation.py
@@ -27,49 +27,135 @@ BlobDetector should be favored over LocalMaxPeakFinder if you are unsure of the 
 the spots are uniformly gaussian in shape. LocalMaxPeakFinder, by contrast, can help find the
 correct threshold.
 
-Spot Decoding
-=============
-
-Starfish's decoding methods are based on its :ref:`Codebook` class. There are two primary decoding
-methods exposed by starfish, a metric based decoder, and a per-channel max decoder.
-
-Metric Decoder
---------------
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
I would suggest a few changes: 

Outside the example: run the script to the point of decoding and save the resulting intensity table as a netcdf object. Put that object on s3 somewhere. 

The script can then:
1. Load the intensity table
2. Load any auxiliary images needed for segmentation
3. Run segmentation
4. Assign spots
5. Create an expression matrix

If the expression matrix is empty, it could be because any of the earlier steps are wrong. I'd look at intermediate steps. I'll be available today on slack if you want to ping me with questions.

ambrosejcarr:
@kne42 checking in on this PR. What's the plan here?

kne42:
Since whatever pipeline was on here is not working properly, I thought I'd apply some of the same techniques you were talking about but with a different dataset:
1. run ISS.ipynb
2. save out registered images to s3

Then the tutorial structure would look like:
1. load up ISS data and registered images from s3
2. _Spot Finding_: use `BlobDetector` on registered images to get intensity table
3. _Spot Decoding_: use `experiment.codebook.decode_per_round_max`
4. _Segmenting Cells_: use `WatershedSegmenter` on registered images and nuclei
5. _Assigning Spots to Cells_: use `Label` on masks and decoded table

we can also forget about 2 & 3 and precompute them on s3 if you feel it isn't useful information, but the gist of this example will be to regurgitate ISS.ipynb

ambrosejcarr:
Sounds good!

PR #1238, opened 2019-04-20
3d smFISH notebook

ambrosejcarr:
* add back notebook
* add missing verbose parameter
* link to readme

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ to see starfish in action on particular assay types. Or, browse our our notebook
  MERFISH              [x]         [x] mer_             in process
  ISS                  [x]         [x] iss_             in process
  osmFISH              [x]         [x] osm_             [ ]
- smFISH               [x]         [x] 3ds_             [ ]
+ smFISH               [x]         [x] 3ds_             [x]
  BaristaSeq           [x]         [x] bar_             [ ]
  DARTFISH             [x]         [x] dar_             [ ]
  ex-FISH              [x]         [ ]                  [ ]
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -39,7 +39,7 @@ to see starfish in action on particular assay types. Or, browse our our notebook
  MERFISH              [x]         [x] mer_             in process
  ISS                  [x]         [x] iss_             in process
  osmFISH              [x]         [x] osm_             [ ]
- smFISH               [x]         [x]                  [ ]
+ smFISH               [x]         [x] 3ds_             [ ]
```
dganguli: place an x under the last column? because we have an entire experiment processed for smFISH?

docs/source/_static/data_processing_examples/3d_smFISH.py
```diff
@@ -4,6 +4,15 @@
 
 This notebook walks through a work flow that analyzes one field of view of a mouse gene panel from
 the Allen Institute for Cell Science, using the starfish package.
+
```
dganguli: this is a nit, but all our notebooks are named <assay acronym>.ipyb/py. this one is not. what if we do 3d_smFISH -> either smFISH. or smFISH_3d. i don't like making a big deal out of 3D for just one notebook, especially when not all our stuff works in 3D yet, e.g., segmentation and target assignment for example.

neuromusic: Also, this isn't from the Allen Institute for Cell Science

dganguli: lol. nice catch. Cell Science -> Brain Science. Brains do have cells though i think.

PR #1240, opened 2019-04-21
Refactor packages to exclude from public namespace

ambrosejcarr:
This PR has two main commits: 
* The first carries out the refactor, which is just a series of moves
* The second appeases flake8, which reorders a bunch of imports. 
* Following commits address changes that had broken master related to documentation and are not relevant to the refactor. 

The post-refactor namespace is a *sub-set* of the pre-factor namespace; no paths _change_, paths are only eliminated that are either duplicates, or which should not be exposed to users (e.g. starfish.multiprocessing)

It's also worth noting this was a clean refactor, I did not have to replace the packages whose imports are broken. For example, despite the fact that you can access `starfish.multiprocessing` from `import starfish` we, the developers, *always* use `import starfish.multiprocessing`. Thus, neither users nor developers expect these packages to be present. 

Before clean-up:

image

After clean-up: 

image

* Potentially worth noting for follow-up PRs that we don't include ExpressionMatrix, DecodedSpots, or SegmentationMaskCollection in our default namespace. 
* Follow-ups could potentially simplify our suggested import strategy in our examples, it should now be possible to exclusively:

```
import starfish
from starfish import types
```

Question: if we want to continue importing types this way, should we also refactor `starfish.types -> starfish._types`?


ambrosejcarr:
Closing in favor of #1244

PR #1244, opened 2019-04-22
RFC: public/private separation

ttung:
Some basic rules of thumb:

1. `starfish_impl` code should never reference `starfish`.
1. tests in `starfish_impl` should generally reference `starfish` whenever possible.  exception: imports in tests for package X should use relative imports to its enclosing package whenever possible.  for example, importing `ImageStack` from `ImageStack` tests should be done as `from ..imagestack import ImageStack`.  Rationale: it's easy to move code around without breaking references.
1. notebook/vignette code should never use `starfish_impl`.

I also created some public APIs where they did not previously exist.  Basically, go check out what was added in `starfish/`


ttung:
Contents of `starfish`:

```
In [2]: dir(starfish)                                                                                                                                                                                                                                                     
Out[2]: 
['Codebook',
 'Experiment',
 'ExpressionMatrix',
 'FieldOfView',
 'ImageStack',
 'IntensityTable',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__is_release_tag__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 '__version__',
 'config',
 'display',
 'starfish']

In [3]:                                                                                                                                                                                                                                                                   ```

**Comment by ttung** [comment]
starfish autocomplete:

```
In [3]: starfish. 
                  Codebook           config             display()          Experiment         ExpressionMatrix   FieldOfView        ImageStack         IntensityTable     starfish           starfish.egg-info/ starfish.wdl                                         
```

ambrosejcarr:
I see that I dropped it too, but I think `starfish.data` would be good to drop into the main namespace. 

Otherwise this LGTM from the end-user perspective.

ttung:
> I see that I dropped it too, but I think starfish.data would be good to drop into the main namespace.

I didn't move `starfish.data`...?

ambrosejcarr:
I mean: it should be imported by default.

ttung:
Was it done so previously?

neuromusic:
I want to let @ambrosejcarr or @shanaxel42 do the actual PR review, but I really dig the API changes here

ttung:
By the way, this is passing all notebooks now.

PR #1249, opened 2019-04-23
Remove .core from some docblocks.

ttung:
I'm not sure this is the right thing to do, so feel free to reject.


ambrosejcarr:
cc @shanaxel42 in case you have opinions on imports.

PR #1251, opened 2019-04-23
Remove clean-docs rule.

ttung:
The solution is the wipe button on readthedocs


ttung:
> Please leave the clean-docs rule, but I think it makes sense to remove from fast so it's not automatically cleaned on travis.

You can use the docs-clean rule.  `make docs-XYZ` gets automagically translated to `make -C docs XYZ`.

PR #1255, opened 2019-04-24
Fix pixel decoding

shanaxel42:
Fixes edge case where centroid of spot evaluates to 0 in decoded image, resulting in setting the target of some spots to nan. @dganguli and I aren't sure if this is the best fix but thought we'd PR for discussion.


ambrosejcarr:
Profiling for this part of the code. Are you fixing this slow spot in this PR? 

image

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -248,4 +248,4 @@ def test_merfish_pipeline_cropped_data():
 
     corrcoef = np.corrcoef(tmp[:, 1], tmp[:, 0])[0, 1]
 
-    assert np.round(corrcoef, 4) == 0.9647
+    assert np.round(corrcoef, 4) == 0.9963
```
dganguli: @ambrosejcarr i think @shanaxel42 finally got to the bottom of why Starfish was 'under-counting' relative to the original MERFISH results.

starfish/test/full_pipelines/api/test_merfish.py
```diff
@@ -232,8 +232,8 @@ def test_merfish_pipeline_cropped_data():
 
     # assert that number of high-expression detected genes are correct
     expected_counts = pd.Series(
-        [120, 101, 54, 43, 30],
-        index=('nan', 'MALAT1', 'SRRM2', 'FASN', 'PRKDC')
+        [104, 63, 55, 45, 32],
+        index=('MALAT1', 'nan', 'SRRM2', 'FASN', 'TLN1')
```
dganguli: why is 'nan' still showing up? it's certainly reduced by ~50% but it should still not appear as long as these spots pass thresholds. right?

starfish/core/spots/_detect_pixels/combine_adjacent_features.py
```diff
@@ -268,7 +268,17 @@ def _single_spot_attributes(
             }
 
         # we're back to 3d or fake-3d here
-        target_index = decoded_image[spot_attrs['z'], spot_attrs['y'], spot_attrs['x']]
```
dganguli: may want to leave a comment in the code here. something like: 

```# define the target index as the most repeated value in the bounding box of the spot. it appears there is no simpler way to do this with a regionprops object```

File: starfish/core/spots/_detect_pixels/combine_adjacent_features.py [hunk 4]
```diff
@@ -268,7 +268,17 @@ def _single_spot_attributes(
...
+        if len(bbox) == 6:
+            # 3d bbox
+            bb_non_zero = np.trim_zeros(np.ndarray.flatten(
+                decoded_image[bbox[0]:bbox[3], bbox[1]:bbox[4], bbox[2]:bbox[5]]))
+        else:
+            # flatten and remove zeros
+            bb_non_zero = np.trim_zeros(np.ndarray.flatten(
+                decoded_image[0, bbox[0]:bbox[2], bbox[1]:bbox[3]]))
+        # get the most repeated value
+        target_index = np.argmax(np.bincount(bb_non_zero))
```
dganguli: this may be returning 0, which then gets resolved to a 'nan' in the MERFISH unit test that i pointed out below. one way to quickly check for this is to find the non-zero most repeated value.  then re-run the merfish test and verify that the 'nan's go away.

dganguli: oh wait nm, np.trim_zeros gets rid of the 0s. hmmm. i'm not sure why you're still getting 'nans' in the merfish test case.

dganguli: oh wait, double nm. np.trim_zeros only removes leading and lagging 0s. 

so ```np.trim_zeros([0, 0, 1, 1, 0 0]) = [1, 1]```

but ```np.trim_zeros([1, 1,, 0, 0, 0, 0, 1]) = [1, 1,, 0, 0, 0, 0, 1]```

so you may not be removing 0s before finding the most repeated element. one way to fix this is like this:

```
target_candidates = np.ndarray.flatten( decoded_image[0, bbox[0]:bbox[2], bbox[1]:bbox[3]]))
non_zero_target_candidates = target_candidates[target_candidates != 0]
target_index = np.argmax(np.bincount(non_zero_target_candidates)))
```

starfish/core/spots/_detect_pixels/combine_adjacent_features.py
```diff
@@ -267,8 +267,20 @@ def _single_spot_attributes(
...
-        target_index = decoded_image[spot_attrs['z'], spot_attrs['y'], spot_attrs['x']]
+        # define the target index as the most repeated value in the bounding box of the spot.
+        #  it appears there is no simpler way to do this with a regionprops object
+        bbox = spot_property.bbox
+        if len(bbox) == 6:
+            # 3d bbox
+            target_candidates = np.ndarray.flatten(
+                decoded_image[bbox[0]:bbox[3], bbox[1]:bbox[4], bbox[2]:bbox[5]])
+        else:
+            # flatten and remove zeros
```
ambrosejcarr: Not clear to me how this is removing zeros

starfish/core/spots/_detect_pixels/combine_adjacent_features.py
```diff
@@ -267,8 +267,20 @@ def _single_spot_attributes(
...
+            target_candidates = np.ndarray.flatten(
+                decoded_image[0, bbox[0]:bbox[2], bbox[1]:bbox[3]])
+        # get the most repeated nonzero value
+        non_zero_target_candidates = target_candidates[target_candidates != 0]
+        target_index = np.argmax(np.bincount(non_zero_target_candidates))
```
ambrosejcarr: This requires quite a lot of introspection to understand how it's working. It'd be good to drop a comment in here. Why remove zeros?

starfish/test/full_pipelines/cli/test_dartfish_cli.py
```diff
@@ -104,4 +104,4 @@ def verify_results(self, intensities):
         corrcoef = np.corrcoef(x, y)
         corrcoef = corrcoef[0, 1]
 
-        assert np.round(corrcoef, 5) == 0.04422
+        assert np.round(corrcoef, 5) == 0.03028
```
ambrosejcarr: I didn't realize our Dartfish pipeline was so bad?

PR #1267, opened 2019-04-25
Add missing files for starfish initialization

ttung:
`starfish/__main__.py` is so that we can do `python -m starfish`, which is what pycharm does.
`starfish/test/__init__.py` is so that setuptools includes the pipeline tests in the zip file.


ttung:
I removed it and it seems to be cool.  Thanks for flagging!

PR #1271, opened 2019-04-25
Enable zero-spot decoding

ambrosejcarr:
* add missing metadata to empty tables when zero spots are detected
* add missing metadata to empty tables when zero spot tables are passed to a decoder
* verify that empty intensity tables can be converted to decoded_spots
* verify those tables can be loaded and dumped to disk
* write down tests

Fixes #1271

### DIFF
--- a/starfish/core/codebook/codebook.py
+++ b/starfish/core/codebook/codebook.py
@@ -552,6 +552,13 @@ class Codebook(xr.DataArray):
 
         self._validate_decode_intensity_input_matches_codebook_shape(intensities)
 
+        # add empty metadata fields and return
+        if intensities.sizes[Features.AXIS] == 0:
+            intensities[Features.TARGET] = (Features.AXIS, np.empty(0, dtype='U'))
+            intensities[Features.DISTANCE] = (Features.AXIS, np.empty(0, dtype=float))
+            intensities[Features.PASSES_THRESHOLDS] = (Features.AXIS, np.empty(0, dtype=bool))
+            return intensities
+
         # normalize both the intensities and the codebook
         norm_intensities, norms = self._normalize_features(intensities, norm_order=norm_order)
         norm_codes, _ = self._normalize_features(self, norm_order=norm_order)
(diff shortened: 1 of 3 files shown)


starfish/core/intensity_table/intensity_table_coordinates.py
```diff
@@ -19,15 +19,19 @@ def transfer_physical_coords_from_imagestack_to_intensity_table(
...
     pairs = (
         (Axes.X.value, Coordinates.X.value),
         (Axes.Y.value, Coordinates.Y.value),
         (Axes.ZPLANE.value, Coordinates.Z.value)
     )
+
+    # make sure the intensity table gets empty metadata if there are no intensities
+    if intensity_table.sizes[Features.AXIS] == 0:
+        for axis, coord in pairs:
+            intensity_table[coord] = xr.DataArray(np.zeros((0)), dims='features')
```
shanaxel42:
```suggestion
            intensity_table[coord] = xr.DataArray(np.zeros((0)), dims=Features.AXIS)
```

shanaxel42:
```suggestion
            intensity_table[coord] = xr.DataArray(np.zeros((0)), dims=Features.AXIS)
```

ambrosejcarr: I think this was duplicated; committed 1/2 :-)

starfish/core/codebook/codebook.py
```diff
@@ -552,6 +552,13 @@ def decode_metric(
 
         self._validate_decode_intensity_input_matches_codebook_shape(intensities)
 
+        # add empty metadata fields and return
+        if intensities.sizes[Features.AXIS] == 0:
```
ttung: what does the code below do (prior to your change)? I assume it didn't create a zero-length array since you're doing this, but what did it create?

starfish/core/codebook/codebook.py
```diff
@@ -634,6 +641,13 @@ def _view_row_as_element(array: np.ndarray) -> np.ndarray:
 
         self._validate_decode_intensity_input_matches_codebook_shape(intensities)
 
+        # add empty metadata fields and return
```
ttung: ditto here.

starfish/core/codebook/codebook.py
ttung: @ambrosejcarr

PR #1272, opened 2019-04-25
Conda pip

joshmoore:
Take 2 of gh-1256 with a fix for `make docker`. (I was bit by the fact that it's not run on regular PRs). The motivation for this is to keep docker/environment.yml generally useful as a way to setup a local conda environment. For development, `pip install -r REQUIREMENTS-CI.txt` is needed:

Test plan:
```
conda env create -n sf -f docker/environment.yml
source activate sf
pip install -r REQUIREMENTS-CI.txt
```

### DIFF
--- a/docker/environment.yml
+++ b/docker/environment.yml
@@ -5,8 +5,9 @@ channels:
 - conda-forge
 dependencies:
 - python>=3.6
+- pip
 - matplotlib==2.1.2
 - numpy==1.16.1
 - pip:
-  - -r REQUIREMENTS-STRICT.txt
+  - -r ../REQUIREMENTS-STRICT.txt
   - jsonpath-rw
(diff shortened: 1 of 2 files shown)


docker/environment.yml
```diff
@@ -5,8 +5,9 @@ channels:
 - conda-forge
 dependencies:
 - python>=3.6
+- pip
 - matplotlib==2.1.2
 - numpy==1.16.1
 - pip:
-  - -r REQUIREMENTS-STRICT.txt
+  - -r ../REQUIREMENTS-STRICT.txt
```
ttung: it's not weird / broken to reach up to a parent directory?

joshmoore: I agree it's slightly weird, yeah, but since the docker directory is effectively self-contained, the impact is pretty low, and it keeps the environment.yml file usable outside of docker.

Issue #1263, opened 2019-04-24
Base calling Filter for in situ sequencing


kevinyamauchi:
# Description
I think it would be useful to create a Filter component for performing base calls. In short, this would compute an image from an `ImageStack` where each channel corresponds to a nucleotide (i.e., C, G, T, A) and each pixel contains the base "quality" The base quality would be a score with range [0, 1] that is the likelihood a particular pixel corresponds to a particular nucleotide.

# Implementation
The filter transforms an image to a nucleotide image by applying a function across the channel dimension to call each pixel for a particular base and returns the confidence score.

In terms of how to score bases, we could adopt something similar to the PHRED ([originally used in capillary-based sequencing](https://en.wikipedia.org/wiki/Phred_base_calling)) or the relative intensity of the channels ([Chen et al. uses the intensity of the strongest channel divided by the root mean square of all four channels for BaristaSeq](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5829746/)). It could be nice to allow users to pass a custom function, but that may not play nice with the logging.

@dganguli , @ambrosejcarr , thoughts?

dganguli:
If I understand correctly, and define space as Z x Y x X, this filter is a RxC x space => RxC x space operation. In the input ImageStack, we have pixel intensity values for each R and C. In the output ImageStack, we have a quality score for each nucleotide (across C) for all rounds R. 

A subsequent decoder could then first perform a transformation of RxC x space => R x space. An example of such a transformation could be an argmax over C.  Finally, each 'called' nucleotide base across R is mapped to the codebook via a lookup table thus creating a R x space => space mapping.

The decoder mentioned above is implemented [here](https://github.com/spacetx/starfish/blob/a35a6233d125e2be7b0e3d8d4be6c201e7e755ae/starfish/core/codebook/codebook.py#L577)  

If I understand this correctly, this Filter may result in better decoding accuracy than simply running the decoder on the raw pixel values themselves. If that's true, I'm all for it! The magic, however, is in specifying the the base scoring function. I really like the idea of allowing users to pass a custom function. This could be done with a lambda from the API which takes a C vector and returns a C vector. We can figure out the logging later.

kevinyamauchi:
[...]

PR #1281, opened 2019-04-26
RFC: Base calling filter for in situ sequencing

kevinyamauchi:
# Overview
As discussed in #1263, this filter component performs base calling for in situ sequencing. The filter determines which base (if any) is represented in each X, Y location in a (round, z plane) slice (i.e., in a (channel, x, y) image). Each pixel value in the resulting image is the base quality score for a given base call.

Currently, the base quality score is calculated as the intensity of the called base divided by the L2 norm of the intensities in all 4 channels (this is how it is done in the original BaristaSeq paper). However, I think it could be useful to add the capability for the user to provide their own custom functions. However, this will likely present challenges for provenance logging. Thoughts, @ambrosejcarr and @dganguli?

# Usage
```python
import numpy as np

from starfish import ImageStack

# Make the test image
test = np.ones((1, 4, 1, 2, 2), dtype='uint16') * 0.5

x = [0, 0, 1, 1]
y = [0, 1, 0, 1]

for i in range(4):
    test[0, i, 0, x[i], y[i]] = 1
test[0, 0, 0, 0, 0] = 0.75

# Make the ImageStack
test_stack = ImageStack.from_numpy(test)

# Filter
bc = Filter.CallBases(intensity_threshold=0.8)
base_call = bc.run(test_stack, in_place=False)

base_call.xarray.values
```

In the example above, the test image has intensities as shown below. The field over view is 2 pixels x 2 pixels. Base for channel 0 is "on" in the upper left corner (intensity=0.75), the base for channel 1 is "on" in the upper right hand  corner (intensity=1), the base for channel 2 is "on" in lower left-hand corner (intensity=1), and the base for channel 3 is "on" in the lower right hand corner (intensity=1). The background intensity is uniformly 0.5 for all channels.
```python
array([[[[[0.75, 0.5 ],
          [0.5 , 0.5 ]]],

        [[[0.5 , 1.  ],
          [0.5 , 0.5 ]]],

        [[[0.5 , 0.5 ],
          [1.  , 0.5 ]]],

        [[[0.5 , 0.5 ],
          [0.5 , 1.  ]]]]], dtype=float32)
```
After filtering, we obtain the following. Note that the base 0 intensity in the upper left hand corner was 0.75, so no base was called due to the intensity threshold we set for the filter (`intensity_threshold=0.8`). The base quality for all other bases is `1 / (1^2 + 0.5^2 + 0.5^2 + 0.5^2)^0.5 = 0.755929`.
```python
array([[[[[0.      , 0.      ],
          [0.      , 0.      ]]],

        [[[0.      , 0.755929],
          [0.      , 0.      ]]],

        [[[0.      , 0.      ],
          [0.755929, 0.      ]]],

        [[[0.      , 0.      ],
          [0.      , 0.755929]]]]], dtype=float32)
```
# Testing
I added tests for the base calling and filtering in `starfish/core/image/_filter/test/test_call_bases.py`

### DIFF
--- a/starfish/core/image/_filter/call_bases.py
+++ b/starfish/core/image/_filter/call_bases.py
@@ -1,11 +1,10 @@
 from functools import partial
-from typing import Optional, Union
+from typing import Optional
 
 import numpy as np
 import xarray as xr
 
 from starfish.core.imagestack.imagestack import ImageStack
-from starfish.core.types import Clip
 from starfish.core.util import click
 from starfish.types import Axes
 from ._base import FilterAlgorithmBase
--- a/starfish/core/image/_filter/call_bases.py
+++ b/starfish/core/image/_filter/call_bases.py
@@ -1,0 +1,118 @@
+from functools import partial
+from typing import Callable, Optional, Union
+
+import numpy as np
+import xarray as xr
+
+from ._base import FilterAlgorithmBase
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.types import Clip
+from starfish.core.util import click
+from starfish.types import Axes
+
+
+class CallBases(FilterAlgorithmBase):
+    def __init__(self, intensity_threshold: float = 0, quality_threshold: float = 0,
+        is_volume: bool = False, expand_dynamic_range: bool = False,
+        clip_method: Union[str, Clip] = Clip.CLIP
+    ) -> None:
+
+        self.intensity_threshold = intensity_threshold
+        self.quality_threshold = quality_threshold
+        self.is_volume = is_volume
+        self.expand_dynamic_range = expand_dynamic_range
+        self.clip_method = clip_method
+
+    _DEFAULT_TESTING_PARAMETERS = {"intensity_threshold": 0, "quality_threshold": 0}
+
+    def _vector_norm(self, x, dim, ord=None):
+        return xr.apply_ufunc(np.linalg.norm, x,
+                              input_core_dims=[[dim]],
+                              kwargs={'ord': ord, 'axis': -1})
+
+    def _call_bases(
+        self, image: xr.DataArray, intensity_threshold: float,
+        quality_threshold: float
+    ) -> xr.DataArray:
+
+        # Get the maximum value for each round/z
+        max_chan = image.argmax(dim=Axes.CH.value)
+        max_values = image.max(dim=Axes.CH.value)
+
+        # Get the norms for each pixel
+        norms = self._vector_norm(x=image, dim=Axes.CH.value)
+
+        # Calculate the base qualities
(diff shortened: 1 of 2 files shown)


starfish/core/image/_filter/call_bases.py
```diff
@@ -0,0 +1,189 @@
...
+            base_qualities < quality_threshold, 0, base_qualities
+        )
+
+        # Threshold the intensity values
+        base_qualities_filtered = xr.where(
+            max_values < intensity_threshold, 0, base_qualities_filtered
+        )
+
+        # Put the base calls in place
+        base_calls = xr.full_like(other=image, fill_value=0)
```
ambrosejcarr: Oh I didn't know about `xr.full_like`. This is great!

starfish/core/image/_filter/test/test_call_bases.py
```diff
@@ -0,0 +1,90 @@
...
+                [[0, 0], [0, 0]]], dtype='float32'
+        )
+
+    bases_xr = xr.DataArray(base_calls, dims=(Axes.CH.value, Axes.X.value, Axes.Y.value))
+
+    return bases_xr
+
+
+@pytest.mark.parametrize("int_thresh, qual_thresh", [(0.8, 0), (0, 0.75), (0.8, 0.75)])
+def test_call_bases(int_thresh: float, qual_thresh: float):
```
ambrosejcarr: Would it be hard to move your example from the PR introduction into the test here? I'm a bit nervous about not having a test that directly consumes the IntensityTable.

kevinyamauchi: No problem. I can add that as an additional test. I like having an isolated test that tests just the base calling method, so I will also keep the tests currently there.

starfish/core/image/_filter/call_bases.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Parameters
+    ----------
+    intensity_threshold : float
+        Minimum intensity a pixel must have to be called a base.
+        Set to zero for no thresholding.
+    quality_threshold : float
+        Minimum base quality score a pixel must have to be called a base.
+        Set to zero for no thresholding.
+    clip_method : Union[str, Clip]
```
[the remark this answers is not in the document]
kevinyamauchi: Sounds good!

starfish/core/image/_filter/test/test_call_bases.py
ambrosejcarr: Oh yes, definitely. Don't remove these tests. :-)

kevinyamauchi:
@ambrosejcarr , thanks for the comments! I've addressed them. Would you mind re-reviewing?

PR #1291, opened 2019-04-29
Fixing coordinate issues with max_proj

### LINKED ISSUE #1214: max_proj does weird things to physical coordinates
Reproduce: run the ISS.ipynb file, but really you just need to max proj the starfish.data.ISS() nuclei image. 

image

shanaxel42:
Fixes #1214


ttung:
Why do we still want this since the `MaxProject` filter exists?

shanaxel42:
MaxProject

> Why do we still want this since the `MaxProject` filter exists?
`MaxProject`  just uses Imagestack.max_proj, so it'll have the same issues

ttung:
> MaxProject just uses Imagestack.max_proj, so it'll have the same issues

LOL ok my bad. :)

starfish/core/imagestack/parser/numpy/__init__.py
```diff
@@ -142,8 +142,8 @@ def get_tile(self, r: int, ch: int, z: int) -> TileData:
                 max_selectors[PHYSICAL_COORDINATE_DIMENSION] = max_selector_value.value
 
                 coordinates[coordinate_type] = (
-                    self.coordinates.loc[min_selectors].item(),
```
ttung: can't you get rid of L139-L142 now?

shanaxel42: oh yea good call

shanaxel42:
> Kind of curious: if you take a max projection across >1 ZPLANES, what is the correct Z coordinate of the result? I think we should have a rational answer for that, and test for it.

Every zplane has it's own zc value so would't it just be that value?

neuromusic:
> if you take a max projection across >1 ZPLANES, what is the correct Z coordinate of the result?

the mean of the coordinates of the individual zplanes

ttung:
> Every zplane has it's own zc value so would't it just be that value?

When you max-project along Z, all the zplanes get flattened into one.  So which zplane would the zc be obtained from?

> the mean of the coordinates of the individual zplanes

Furthermore, if the zplanes are not evenly spaced, then you would need to average all of the individual z coordinates, rather than just take the first and last and average them. :)

shanaxel42:
> When you max-project along Z, all the zplanes get flattened into one. So which zplane would the zc be obtained from?

But doesn't this happen by just choosing the max value from whatever dim your max_projecting on? And we've already calculated a value for each z_pane. So in this case we'd choose the max z_plane and then the z_coord is whatever we'd previously calculate for that z_plane?

ttung:
For each pixel, you could be sourcing the value from a different z plane.  However, we only store a coordinate value per plane, and so when you flatten to one, you have to store only one z coordinate.

kevinyamauchi:
@shanaxel42, what is the status of this PR? I am trying to add other types of projection (e.g., sum across axes) in #1342 and I was hoping to integrate the changes here for transferring physical coordinates.

shanaxel42:
> @shanaxel42, what is the status of this PR? I am trying to add other types of projection (e.g., sum across axes) in #1342 and I was hoping to integrate the changes here for transferring physical coordinates.

Hey! Sorry this one kinda fell off my radar, I'll get it merged today!

kevinyamauchi:
Awesome! Thanks!

starfish/core/imagestack/imagestack.py
```diff
@@ -299,7 +300,7 @@ def from_numpy(
             cls,
             array: np.ndarray,
             index_labels: Optional[Mapping[Axes, Sequence[int]]]=None,
-            coordinates: Optional[xr.DataArray]=None,
+            coordinates: Optional[DataArrayCoordinates]=None,
```
ttung: parameter docs not updated below.

starfish/core/imagestack/imagestack.py
```diff
@@ -1152,7 +1153,8 @@ def max_proj(self, *dims: Axes) -> "ImageStack":
         max_projection = self._data.max([dim.value for dim in dims])
         max_projection = max_projection.expand_dims(tuple(dim.value for dim in dims))
         max_projection = max_projection.transpose(*self.xarray.dims)
-        max_proj_stack = self.from_numpy(max_projection.values)
+        physical_coords = self.xarray.coords
+        max_proj_stack = self.from_numpy(max_projection.values, coordinates=physical_coords)
```
ttung: the coordinates passed in here should be shaped like the array passed in here.  the z averaging should happen here.

starfish/core/imagestack/parser/numpy/__init__.py
```diff
@@ -122,28 +122,19 @@ def get_tile(self, r: int, ch: int, z: int) -> TileData:
...
-                max_selectors = dict(selectors)
-                max_selectors[PHYSICAL_COORDINATE_DIMENSION] = max_selector_value.value
-
                 coordinates[coordinate_type] = (
-                    self.coordinates.loc[min_selectors].item(),
-                    self.coordinates.loc[max_selectors].item())
+                    float(self.coordinates[coordinate_type][0]),
+                    float(self.coordinates[coordinate_type][-1]))
+            # For the z coordinate, take the average.
+            average_z_coord = np.average(self.coordinates[Coordinates.Z])
```
ttung: If you pass this a full 5D tensor (i.e., where z has many layers) into `from_numpy`, this will mess up your z coordinates.

starfish/core/imagestack/imagestack.py
```diff
@@ -310,7 +311,7 @@ def from_numpy(
         index_labels : Optional[Mapping[Axes, Sequence[int]]]
             Mapping from axes (r, ch, z) to their labels.  If this is not provided, then the axes
             will be labeled from 0..(n-1), where n=the size of the axes.
-        coordinates : Optional[xr.DataArray]
+        coordinates : Optional[DataArrayCoordinates]
             DataArray indexed by r, ch, z, with xmin, xmax, ymin, ymax, zmin, zmax as columns.  If
```
ttung: This needs to be updated as well, presumably.

starfish/core/imagestack/imagestack.py
```diff
@@ -1147,12 +1148,18 @@ def max_proj(self, *dims: Axes) -> "ImageStack":
...
-
         """
         max_projection = self._data.max([dim.value for dim in dims])
         max_projection = max_projection.expand_dims(tuple(dim.value for dim in dims))
         max_projection = max_projection.transpose(*self.xarray.dims)
-        max_proj_stack = self.from_numpy(max_projection.values)
+        physical_coords = self.xarray.coords
+        if Axes.ZPLANE in dims:
+            # if we max proj by z, take average coord
+            physical_coords[Coordinates.Z].values[0] = \
```
ttung: don't calculate the average twice.

also, the physical_coords[Coordinates.Z] is the wrong size!  It should be size = 2, according to the contract for `from_numpy` (min, max).

furthermore, the size of the coords for X/Y is also wrong.  It's currently the size of the X/Y, and presumably it ought to also be 2. 

```
average_z_coordinate = np.average((self.xarray.coords[Coordinates.Z])
physical_coords[Coordinates.Z] = xr.DataArray(...)
```

Furthermore, if one max projects across another dimension where we assign coordinates (like x or y), we would presumably need to do the same thing.  Alternatively, we could explicitly disallow max projecting across x or y, but if we want to do that, we should clear it with Deep/Ambrose/Kevin/Brian.

shanaxel42: There is no min/max contract in from_numpy, it takes the first and last values from the array. So they can be any size

shanaxel42: But I can do that here if you think that's cleaner

starfish/core/imagestack/imagestack.py
ttung: contract is here. actually it's not even up to date. :/

shanaxel42: yea this is def not true anymore. it should just be DataArrayCoordinates object

ttung:
Going to close this as it landed with #1353

Issue #1310, opened 2019-05-04
`IntensityTable` is truncating x,y,z coordinates for spot detection methods that return subpixel results


njmei:
We've been in the process of comparing the results of our current pipeline vs. a starfish version and it looks like we've stumbled on a really unfortunate 'feature' of the current `IntensityTable` data structure for spot based detection methods.

Below is a visual example (diamonds are our pipeline, circles are starfish), with the takeaway being that the starfish circles are not centered on detected spots:
![Screenshot from 2019-05-02 14-44-25](https://user-images.githubusercontent.com/43766899/57182459-2941d800-6e54-11e9-8439-17f0be3da602.png)

Both our spot finder and the [Starfish equivalent](http://soft-matter.github.io/trackpy/dev/tutorial/uncertainty.html) are based on the Crocker-Grier centroid finding algorithm and can determine the coordinates of a spot with sub-pixel resolution.

So I ended up chasing a [red herring](https://github.com/spacetx/starfish/pull/1300) before discovering that `IntensityTable` x, y, z properties are `int64`, which would `floor()` any result returned by the Crocker-Grier algorithm. In fact, this flooring appears to happen [immediately after results are produced](https://github.com/spacetx/starfish/blob/master/starfish/core/spots/_detect_spots/trackpy_local_max_peak_finder.py#L137).

It looks like (xc, yc, zc) are then real world coordinates that map to each set of pixel indices (x, y, z). I think that this design is probably (?) fine for pixel based decoding methods, but throwing out sub-pixel information for spot methods is a big deal breaker.

CC: @berl @ambrosejcarr

dganguli:
@njmei pardon my ignorance here -- can this be fixed simply by removing the flooring you pointed out immediately after the results are produced? 

if that's true, does that break anything else? if not, then i think the fix is simple? 

but i'm sure i'm missing something here

ambrosejcarr:
Seems like we'll also need to make the IntensityTable pixel coordinates Floats instead of Ints. This feels like a sensible and relatively simple change to me.

njmei:
@dganguli I'm hoping resolving this is as simple as removing the flooring :)

But there was probably (?) a reason the spot values were converted to `int` which is why I opened this issue. I'm wondering if there are any specific indexing operations that rely on the spot locations being ints?

dganguli:
@njmei got it. some basic sleuthing can elucidate why this decision was made:

https://github.com/spacetx/starfish/pull/898

[...]

PR #1315, opened 2019-05-06
[Easy] Change IntensityTable x,y,z SpotAttributes to floats for spot-based methods

njmei:
Spot detection methods return spot coordinates with
sub-pixel resolution. Previously this sub-pixel information
was thrown away due to int conversion (see: #898).

The correct behavior should be to retain sub-pixel
information and modify downstream functions to
properly handle float values for x,y,z pixel
coordinates.

Relates to: #1310

NOTE: This will break functions like: https://github.com/spacetx/starfish/blob/master/starfish/core/intensity_table/intensity_table_coordinates.py and possibly others!


ttung:
This does not calculate physical coordinates correctly for the spots.  I'm working on a solution for that.

njmei:
@ttung I kinda had a suspicion that might be the case. Do you want me to go ahead and close this pull then?

PR #1325, opened 2019-05-07
upload ISS WDL

ambrosejcarr:
* Adds a workflow for processing ISS data using the starfish python API
* Adds inputs that trigger test-sized data and the complete spaceTx ISS mouse dataset.

### DIFF
--- a/workflows/iss/iss.wdl
+++ b/workflows/iss/iss.wdl
@@ -1,0 +1,133 @@
+
+task process_field_of_view {
+
+    String experiment
+    Int field_of_view
+
+    command <<<
+        python3 <<CODE
+
+        import numpy as np
+        import starfish
+        from starfish.types import Axes
+
+        fov: int = ${field_of_view}
+        fov_str: str = f"fov_{int(fov):03d}"
+
+        # load experiment
+        experiment = starfish.Experiment.from_json("${experiment}")
+
+        fov = experiment[fov_str]
+        imgs = fov.get_image(starfish.FieldOfView.PRIMARY_IMAGES)
+        dots = imgs.max_proj(Axes.CH)
+
+        # filter
+        filt = starfish.image.Filter.WhiteTophat(masking_radius=15, is_volume=False)
+        filtered_imgs = filt.run(imgs, verbose=True, in_place=False)
+        filt.run(dots, verbose=True, in_place=True)
+
+        # find threshold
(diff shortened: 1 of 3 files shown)


workflows/iss/iss.wdl
```diff
@@ -0,0 +1,134 @@
...
+
+        # blobs = dots; define the spots in the dots image, but then find them again in the stack.
+        intensities = p.run(filtered_imgs, blobs_image=dots, blobs_axes=(Axes.ROUND, Axes.ZPLANE))
+
+        # decode
+        decoded = experiment.codebook.decode_per_round_max(intensities)
+
+        # save results
+        df = decoded.to_decoded_spots()
+        df.save_csv("decoded.csv")
```
shanaxel42: Do you wanna save the file as like "decoded.csv+ fov" so we know which fov the output is for?

workflows/iss/iss.wdl
```diff
@@ -0,0 +1,134 @@
+
+task process_field_of_view {
+
+    String experiment
+    Int field_of_view
+
+    command <<<
+        python3 <<CODE
+
+        import numpy as np
```
ttung: move this into a .py file so we can test it.

even better, but entirely optional, would be to structure it as a recipe (rationale: other people would be less tempted to tinker with the pipeline in a way that makes it non-traceable from a provenance POV).

workflows/iss/iss.wdl
```diff
@@ -0,0 +1,134 @@
...
+        df = decoded.to_decoded_spots()
+        df.save_csv("decoded.csv")
+        CODE
+    >>>
+
+    runtime {
+        docker: "spacetx/starfish:0.1.0-simple"
+        memory: "16 GB"
+        cpu: "4"
+        disk: "local-disk 100 SDD"
```
ttung: I suspect you don't need a 100GB SSD to process this dataset.  There's literally nothing here that takes advantage of disk.

You may also want some config to disable the slicedimage caching.  Rationale: You will not load the data twice, so writing it to disk is counterproductive.

ambrosejcarr:
@shanaxel42 had some ideas about how to follow up on @ttung 's requests, so I'm going to reassign this to avoid duplicating work.

shanaxel42:
> @shanaxel42 had some ideas about how to follow up on @ttung 's requests, so I'm going to reassign this to avoid duplicating work.

I think this work is taken care of in https://github.com/spacetx/starfish/pull/1364/files. So I'm going to close this one out!

Issue #1328, opened 2019-05-08
Refactor FieldOfView.get_image() to FieldOfView.get_images()


shanaxel42:
After #1326 we should refactor FieldOfView.get_image() to be FieldOfView.get_images() that takes in an image name and optional round/chs/zplane,x,y selected axes parameters. If the selected axes given are within multiple aligned image groups FieldOfView.get_images() should return a list of Aligned Imagestacks

PR #1329, opened 2019-05-08
Adding back ability to crop on fov.get_image()

shanaxel42:
Unblocks @ambrosejcarr on seqFish work for now but will eventually be overridden by #1328

### DIFF
--- a/starfish/core/imagestack/parser/crop.py
+++ b/starfish/core/imagestack/parser/crop.py
@@ -58,6 +58,31 @@ class CropParameters:
         if axis_type == Axes.ZPLANE and self._permitted_zplanes:
             self._permitted_zplanes.add(permitted_axis)
 
+    def further_crop(self, further_crop_params) -> "CropParameters":
+        if further_crop_params._permitted_chs and \
+                not further_crop_params._permitted_chs.issubset(self._permitted_chs):
+            raise ValueError("Can only provide further ch crop params within the existing set of"
+                             "ch crop params")
+        else:
+            self._permitted_chs = further_crop_params._permitted_chs
+
+        if further_crop_params._permitted_rounds and \
+                not further_crop_params._permitted_rounds.issubset(self._permitted_rounds):
+            raise ValueError("Can only provide further round crop params within the existing set of"
+                             "round crop params")
+        else:
+            self._permitted_rounds = further_crop_params._permitted_rounds
+
+        if further_crop_params._permitted_zplanes and \
+                not further_crop_params._permitted_zplanes.issubset(self._permitted_zplanes):
+            raise ValueError("Can only provide further zplane crop params within the existing set "
+                             "of z_plane crop params")
+        else:
+            self._permitted_zplanes = further_crop_params._permitted_zplanes
+        self._x_slice = further_crop_params._x_slice
+        self._y_slice = further_crop_params._y_slice
+        return self
+
     def filter_tilekeys(self, tilekeys: Collection[TileKey]) -> Collection[TileKey]:
         """
         Filters tilekeys for those that should be included in the resulting ImageStack.
--- a/starfish/core/imagestack/parser/crop.py
+++ b/starfish/core/imagestack/parser/crop.py
@@ -59,24 +59,24 @@ class CropParameters:
             self._permitted_zplanes.add(permitted_axis)
 
     def further_crop(self, further_crop_params) -> "CropParameters":
-        if further_crop_params._permitted_chs and \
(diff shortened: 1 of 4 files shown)


shanaxel42:
> Needs tests. :)

Was the test in test_experiment.py not enough?

ttung:
> Was the test in test_experiment.py not enough?

I swear it wasn't there when I reviewed it, apologies. :(

starfish/core/imagestack/parser/crop.py
```diff
@@ -58,6 +58,31 @@ def _add_permitted_axes(self, axis_type: Axes, permitted_axis: int) -> None:
         if axis_type == Axes.ZPLANE and self._permitted_zplanes:
             self._permitted_zplanes.add(permitted_axis)
 
+    def further_crop(self, further_crop_params) -> "CropParameters":
+        if further_crop_params._permitted_chs and \
```
ttung: when you have a multiline condition clause, i think it's ok to do

```
if (xxxx
    and yyyy ...
):
```

starfish/core/imagestack/parser/crop.py
```diff
@@ -58,6 +58,31 @@ def _add_permitted_axes(self, axis_type: Axes, permitted_axis: int) -> None:
         if axis_type == Axes.ZPLANE and self._permitted_zplanes:
             self._permitted_zplanes.add(permitted_axis)
 
+    def further_crop(self, further_crop_params) -> "CropParameters":
```
ttung:
```suggestion
    def further_crop(self, further_crop_params: "CropParameters") -> "CropParameters":
```

A simple docblock here would be good.

starfish/core/imagestack/parser/crop.py
```diff
@@ -58,6 +58,31 @@ def _add_permitted_axes(self, axis_type: Axes, permitted_axis: int) -> None:
...
+                not further_crop_params._permitted_rounds.issubset(self._permitted_rounds):
+            raise ValueError("Can only provide further round crop params within the existing set of"
+                             "round crop params")
+        else:
+            self._permitted_rounds = further_crop_params._permitted_rounds
+
+        if further_crop_params._permitted_zplanes and \
+                not further_crop_params._permitted_zplanes.issubset(self._permitted_zplanes):
+            raise ValueError("Can only provide further zplane crop params within the existing set "
+                             "of z_plane crop params")
```
ttung:
```suggestion
                             "of zplane crop params")
```

^^ to match your earlier reference.

starfish/core/imagestack/parser/crop.py
ttung: You could rewrite the three if-else as one single if-else inside a loop if you want.  It's debatable if it's better.

dganguli:
Sorry for chiming in a bit late here, but I'd like to revisit the name "Crop" in this new context. In the field of image processing / image analysis, cropping implies selecting a subset of an image in *spatial coordinates*, e.g., X, Y, and Z. If I understand correctly, this PR _also_ allows me to select a subset of an image in non-spatial coordinates, e.g., R, C. 

I definitely want this functionality, but I find the name CropParameters and all the references to crop to be confusing. Here are some other ideas:

 ```crop``` ->  ```select``` or ```subset``` or ```index```

correspondingly, 
```CropParameters``` -> ```SelectionParameters``` or ```SubsetParameters``` or ```AxesParameters```

shanaxel42:
> Sorry for chiming in a bit late here, but I'd like to revisit the name "Crop" in this new context. In the field of image processing / image analysis, cropping implies selecting a subset of an image in _spatial coordinates_, e.g., X, Y, and Z. If I understand correctly, this PR _also_ allows me to select a subset of an image in non-spatial coordinates, e.g., R, C.
> 
> I definitely want this functionality, but I find the name CropParameters and all the references to crop to be confusing. Here are some other ideas:
> 
> `crop` -> `select` or `subset` or `index`
> 
> correspondingly,
> `CropParameters` -> `SelectionParameters` or `SubsetParameters` or `AxesParameters`

@dganguli does the suggestion that @ambrosejcarr made about just passing in the r/ch/z parameters directly work for your concerns?

dganguli:
@shanaxel42 @ambrosejcarr whoops! Yeah it seems like Ambrose and I are saying the same thing. After re-reading, I agree with Ambrose. Sorry!

dganguli:
i do think, we should avoid using the term ```crop``` if it refers to more than manipulating the X,Y,Z axes though. ambrose's solution indeed avoids this.

shanaxel42:
merging this as an intermediate step but immediately working on  #1328 and will address @ttung testing case

PR #1330, opened 2019-05-09
Preserve subpixel offsets from spot detection

### LINKED ISSUE #1310: `IntensityTable` is truncating x,y,z coordinates for spot detection methods that return subpixel results

ttung:
1. Remove the casts to int.

2. Interpolate the coordinates if there are more than one pixel location, otherwise use sel.

3. Update tests.

Fixes #1310

### DIFF
--- a/starfish/core/intensity_table/intensity_table_coordinates.py
+++ b/starfish/core/intensity_table/intensity_table_coordinates.py
@@ -33,11 +33,15 @@ def transfer_physical_coords_from_imagestack_to_intensity_table(
         return intensity_table
 
     for axis, coord in pairs:
+        imagestack_pixels: xr.DataArray = image_stack.xarray[axis]
+        intensity_table_pixel_offsets: np.ndarray = intensity_table[axis].values
 
-        imagestack_pixels: np.ndarray = image_stack.xarray[axis].values
-        intensity_table_pixels: np.ndarray = intensity_table[axis].values.astype(int)
-        pixel_inds: np.ndarray = imagestack_pixels[intensity_table_pixels]
-        coordinates: xr.DataArray = image_stack.xarray[coord][pixel_inds]
+        # can't interpolate if the axis size == 1, so just select in that case.
+        if len(imagestack_pixels) == 1:
+            coordinate_fetcher = imagestack_pixels.sel
+        else:
+            coordinate_fetcher = imagestack_pixels.interp
+        coordinates = coordinate_fetcher({axis: intensity_table_pixel_offsets})[coord]
 
         intensity_table[coord] = xr.DataArray(coordinates.values, dims='features')
 
--- a/starfish/core/spots/_detect_spots/blob.py
+++ b/starfish/core/spots/_detect_spots/blob.py
@@ -115,7 +115,7 @@ class BlobDetector(DetectSpotsAlgorithmBase):
         fitted_blobs[Features.SPOT_RADIUS] = converted_radius
 
         # convert the array to int so it can be used to index
-        rounded_blobs = SpotAttributes(fitted_blobs.astype(int))
+        rounded_blobs = SpotAttributes(fitted_blobs)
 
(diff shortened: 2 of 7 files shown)


starfish/core/spots/_detect_spots/blob.py
```diff
@@ -115,7 +115,7 @@ def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> SpotAtt
         fitted_blobs[Features.SPOT_RADIUS] = converted_radius
 
         # convert the array to int so it can be used to index
-        rounded_blobs = SpotAttributes(fitted_blobs.astype(int))
+        rounded_blobs = SpotAttributes(fitted_blobs)
```
dganguli: rename ```rounded_blobs``` -> ```spots```? since they're no longer rounded, and they're actually now spots as defined by SpotAttributes

starfish/core/intensity_table/intensity_table_coordinates.py
```diff
@@ -33,11 +33,15 @@ def transfer_physical_coords_from_imagestack_to_intensity_table(
...
-        imagestack_pixels: np.ndarray = image_stack.xarray[axis].values
-        intensity_table_pixels: np.ndarray = intensity_table[axis].values.astype(int)
-        pixel_inds: np.ndarray = imagestack_pixels[intensity_table_pixels]
-        coordinates: xr.DataArray = image_stack.xarray[coord][pixel_inds]
+        imagestack_pixels: xr.DataArray = image_stack.xarray[axis]
+        intensity_table_pixel_offsets: np.ndarray = intensity_table[axis].values
+
+        # can't interpolate if the axis size == 1, so just select in that case.
+        if len(imagestack_pixels) == 1:
+            func = imagestack_pixels.sel
```
dganguli: rename ```func``` to something a bit more expressive? like ```get_coordinates```?

starfish/core/spots/_detect_spots/blob.py
ttung: so would we do a search-and-replace of `blob`->`spot`?  i.e., would `fitted_blobs` be renamed to `fitted_spots`?

starfish/core/intensity_table/intensity_table_coordinates.py
ttung: will do.

starfish/core/spots/_detect_spots/blob.py
dganguli: i think fitted_blobs should stay fitted_blobs, as it's the output of a blob detector. my sense is no need for the search-and-replace.

PR #1342, opened 2019-05-13
Add Filter.Reduce (general dimension reduction for ImageStack)

### LINKED ISSUE #1302: Add support for other types of projections on ImageStack
# Proposal
I think it would be useful to expose other methods of projecting along axes for `ImageStack` (e.g., `DataArray.sum()` and `DataArray.mean()`). This would add support the ExFISH case for projecting across (R, C) before spot segmentation (#1301).

# Implementation
Some possibilities:
1. I think the thing that is the least work is to make a filter, `Filter.Project` that takes projection type as an input argument. This could replace `Filter.MaxProject`.
2. We could do the same as (1), but make it a method of `ImageStack`. I think a lot of the notebooks would have to be updated, as they use `ImageStack.max_proj()`.
3. We could also make one filter component per projection type, but this seems too granular.

As a related aside: I think there should be only one way to max project (i.e., `ImageStack.max_proj()` and `Filter.MaxProject` are redundant).

Thoughts, @dganguli , @ambrosejcarr ?

kevinyamauchi:
# Summary
As discussed in #1302, this component provides the ability to perform dimension reduction operations to `ImageStack` using `DataArray.reduce()`. Currently, max, mean, sum, and a user-provided  can be applied across specified dimensions. Closes #1302.

# Testing
Testing of sum, mean, and max are performed in `starfish/core/image/_filter/test/test_reduce.py`

# To Do
- [x] Add tests
- [x] Update physical coordinates


starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+import numpy as np
+
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.types import Axes
+from starfish.core.util import click
+from ._base import FilterAlgorithmBase
+
+
+class Reduce(FilterAlgorithmBase):
+    """
```
ttung: Did we decide to put the docs in the class and not the constructor?  I can't remember. :(

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+    Reduces the dimensions of the ImageStack by applying a function
+    along one or more axes.
+
+    Parameters
+    ----------
+    dims : Axes
+        one or more Axes to project over
+    func : Union[str, Callable]
+        function to apply across the dimension(s) specified by dims.
+        If a functino is provided, it follow the form specified by
```
ttung:
```suggestion
        If a function is provided, it follow the form specified by
```

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+class Reduce(FilterAlgorithmBase):
+    """
+    Reduces the dimensions of the ImageStack by applying a function
+    along one or more axes.
+
+    Parameters
+    ----------
+    dims : Axes
+        one or more Axes to project over
+    func : Union[str, Callable]
```
ttung: why not accept just Callable?

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+            if func == 'max':
+                func = np.amax
+            elif func == 'mean':
+                func = np.mean
+            elif func == 'sum':
+                func = np.sum
+            else:
+                raise ValueError('func should be max, mean, or sum')
+        elif callable(func):
+            warnings.warn('User-specific functions are not logged')
```
ttung: why is this true?

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+    @click.option(
+        "--func",
+        type=click.Choice(["max", "mean", "sum"]),
+        multiple=False,
+        help="The function to apply across dims"
+             "Valid function names: max, mean, sum."
+    )
+    @click.pass_context
+    def _cli(ctx, dims, func):
+        ctx.obj["component"]._cli_run(ctx, Reduce(dims, func))
```
ttung: if you make the constructor only accept callable, you can make it support a wide variety of cases by doing:

`func = getattr(np, func)`

the CLI would require someone to use `--func amax`, as that is the only one that doesn't map exactly from the expected name to name in numpy space.

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        if not in_place:
+            stack = deepcopy(stack)
+
+        # Apply the reducing function
+        reduced = stack._data.reduce(self.func, dim=[dim.value for dim in self.dims])
+
+        # Add the reduced dims back and align with the original stack
+        reduced = reduced.expand_dims(tuple(dim.value for dim in self.dims))
+        reduced = reduced.transpose(*stack.xarray.dims)
```
ttung: I wonder if the physical coordinates are correct here.  You may want to look at #1291.

starfish/core/image/_filter/reduce.py
dganguli: i like it. this is the more 'pythonic' way in scientific computing -- will be easier for numpy/pandas users to reason about i think.

starfish/core/image/_filter/reduce.py
dganguli: oh. good point. huh.

starfish/core/image/_filter/reduce.py
kevinyamauchi: This is how it's done in other filter components, but I can move it if the plan is to change it moving forward.

ttung: IMO that is a mistake, and we should rectify it. :)

ttung: (deleted earlier comment) I misread it.

Cool, if this is what is done elsewhere, that's fine.

starfish/core/image/_filter/reduce.py
ttung: @dganguli You like what?  The original or my suggestion?

kevinyamauchi:
@dganguli , I think @ttung suggested that we keep `Filter.max_proj()` and `ImageStack.max_proj()` for a bit for API stability. I would be in favor of removing `ImageStack.max_proj()` and refactoring `Filter.max_proj()` to use `Filter.reduce()`. I think it's worth keeping `Filter.max_proj()` around because maximum intensity projection is such a common operation. Thoughts?

starfish/core/image/_filter/reduce.py
kevinyamauchi: I think this is an interesting approach and I have a couple of questions:
- Is this really accepting only a callable? As I understand it, `func` would the string of a name of a callable.
- What if the user wants to define their own function to apply across an axis or specify something that is in a different library than numpy. Would we use a try/catch on `func = getattr(np, func)`? Or is there a way to check it before trying?
- I agree that the user will expect `max` even though they would mean `amax`. Perhaps we could map them together  behind the scenes  in the constructor?
- I can replace the `func` setting logic when a string is provided with `func = getattr(np, func)` and keep the ability of users to pass their own functions as well. What do you think?

starfish/core/image/_filter/reduce.py
kevinyamauchi: Ah I didn't inspect the log very closely and I thought it wasn't actually storing the callable. I now see that it does, so I can remove the warning. Thanks!

starfish/core/image/_filter/reduce.py
kevinyamauchi: Great point. Shannon said she would merge that soon and I can fix the physical coordinates.

ttung: I just landed #1353, which encompasses #1291

kevinyamauchi: Awesome! Thanks @ttung and @shanaxel42 !

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,151 @@
...
+                func = 'amax'
+            func = getattr(np, func)
+        self.func = func
+
+    _DEFAULT_TESTING_PARAMETERS = {"dims": ['r'], "func": 'max'}
+
+    def run(
+            self,
+            stack: ImageStack,
+            in_place: bool = False,
```
ttung: I would explicitly not support in_place.  It doesn't make any sense.

kevinyamauchi: Sounds good. In max_proj, the `run()` method still accepts an `in_place` argument and just doesn't do anything with it. Should I do the same here? Maybe give a warning?

kevinyamauchi:
@ttung , I think I've addressed your comments. However, the test is failing during doc generation (see error below) and I'm not sure what it is expecting because it passes `flake8`
```
Warning, treated as error:
/home/travis/build/spacetx/starfish/starfish/core/image/_filter/reduce.py:docstring of starfish.image.Filter.Reduce:21:Unexpected indentation.
make[1]: *** [html] Error 2
make[1]: Leaving directory `/home/travis/build/spacetx/starfish/docs'
make: *** [docs-html] Error 2
The command "make install-dev fast" exited with 2.
```

starfish/core/image/_filter/reduce.py
ttung: Just remove the in_place and n_processes arguments, as neither do what they are advertised to do.

kevinyamauchi:
@ttung, I think I have addressed your comments. Would you mind taking another look?

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,144 @@
...
+        max: maximum intensity projection (applies numpy.amax)
+        mean: take the mean across the dim(s) (applies numpy.mean)
+        sum: sum across the dim(s) (applies numpy.sum)
+    clip_method : Union[str, Clip]
+        (Default Clip.CLIP) Controls the way that data are scaled to retain skimage dtype
+        requirements that float data fall in [0, 1].
+        Clip.CLIP: data above 1 are set to 1, and below 0 are set to 0
+        Clip.SCALE_BY_IMAGE: data above 1 are scaled by the maximum value, with the maximum
+        value calculated over the entire ImageStack
+        Clip.SCALE_BY_CHUNK: data above 1 are scaled by the maximum value, with the maximum
```
ttung: group_by is not provided...?

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,144 @@
...
+        DataArray.reduce():
+        http://xarray.pydata.org/en/stable/generated/xarray.DataArray.reduce.html
+        If a string is provided, it should correspond to a numpy function that
+        matches the form specified above
+        (i.e., function is resolved: func = getattr(np, func)).
+        Some common examples below:
+        max: maximum intensity projection (applies numpy.amax)
+        mean: take the mean across the dim(s) (applies numpy.mean)
+        sum: sum across the dim(s) (applies numpy.sum)
+    clip_method : Union[str, Clip]
```
ttung: should just be
 
```suggestion
    clip_method : Clip
```

starfish/core/image/_filter/reduce.py
kevinyamauchi: I think it's nice to have the string option. I think a lot of users will find it annoying to need to import numpy.amax() and this component to perform a maximum intensity projection.

starfish/core/image/_filter/reduce.py
kevinyamauchi: Ah good point. I will remove this option.

starfish/core/image/_filter/reduce.py
ambrosejcarr: I support the string option, just for ecosystem continuity -- users are used to passing function names that are resolved by getattr, as this formalism is used a lot in `scipy.spatial.distance` among other packages.

starfish/core/image/_filter/reduce.py
```diff
@@ -0,0 +1,144 @@
...
+    func : Union[str, Callable]
+        function to apply across the dimension(s) specified by dims.
+        If a function is provided, it should follow the form specified by
+        DataArray.reduce():
+        http://xarray.pydata.org/en/stable/generated/xarray.DataArray.reduce.html
+        If a string is provided, it should correspond to a numpy function that
+        matches the form specified above
+        (i.e., function is resolved: func = getattr(np, func)).
+        Some common examples below:
+        max: maximum intensity projection (applies numpy.amax)
```
ambrosejcarr: @kevinyamauchi `max` -> `np.amax` is not consistent with the stated functionality of the documentation using `getattr`, and in the code, it's special cased. Can you resolve the logical discontinuity in the documentation?

starfish/core/image/_filter/reduce.py
kevinyamauchi: Thanks! Fixed.

starfish/core/image/_filter/reduce.py
kevinyamauchi: Ah good point. Fixed.

PR #1346, opened 2019-05-14
Refactor fov.getImage() to fov.getImages()

### LINKED ISSUE #1328: Refactor FieldOfView.get_image() to FieldOfView.get_images()

shanaxel42:
fixes #1328 
Refactors fov.getImage() to

 ```
 def get_images(self, item: str,
                   rounds: Optional[Collection[int]] = None,
                   chs: Optional[Collection[int]] = None,
                   zplanes: Optional[Collection[int]] = None,
                   x: Optional[Union[int, slice]] = None,
                   y: Optional[Union[int, slice]] = None,
                   ) -> Union[ImageStack, List[ImageStack]]:
``` 

The user now provides whatever axes they want to include. If no axes provided, we include them all. If the group of selected axes contain unaligned tiles we return an AlignedImageStackIterator where each stack is an aligned group. This PR also completely abstracts the idea of "AlignedCordGroups" from the user. They no longer need to provide an aligned group number anywhere or worry about the list. They only worry about which axes they want to include in there get_images() call.

### CODE REVIEW

File: starfish/core/experiment/experiment.py [hunk 1]
```diff
@@ -105,105 +96,74 @@ def name(self) -> str:
...
-            yield self.get_image(item=image_type, aligned_group=aligned_group)
-
-    def get_image(self, item: str, aligned_group: int = 0,
-                  rounds: Optional[Collection[int]] = None,
-                  chs: Optional[Collection[int]] = None,
-                  zplanes: Optional[Collection[int]] = None,
-                  x: Optional[Union[int, slice]] = None,
-                  y: Optional[Union[int, slice]] = None,
-                  ) -> ImageStack:
+    def get_images(self, item: str,
```
ttung: I would prefer that `get_images` always returns an iterator.  We can leave `get_image` in the codebase, and consider deprecating it (a la `@deprecated`, or whatever similar annotation exists for python).

shanaxel42: I think this was something @ambrosejcarr specifically asked for? I will say it's kinda nice when you know your params are going to just return one imagestack like this allen dataset where I use get_images(rounds=[1]) and you don't have to then do a next(imagesatcks)

ttung: It's probably convenient when you know exactly the nature of your data, and you just want to lazy-mode it.  In practice, I think this is likely going to be a lot of:

```
x = get_images(...)
if isinstance(x, ImageStack):
    do something
else:
    for img in x:
        do something
```

@ambrosejcarr you want to make the case for this? :)

ambrosejcarr: I think the UX I'd like is: 

`get_image() -> ImageStack`
If this 
`get_images() -> Iterator[ImageStack]`

ambrosejcarr: This might be a good place to get a consult from @kne42 as well.

ttung: I'm cool with what you suggested.

kne42: Yeah it should definitely always return an iterator for exactly the reason @ttung presented in his example.

If a user expects there to only be one `ImageStack`, they can unpack it like so:
```python
stack, = get_images(rounds=[1])
```
although I don't think this will be very intuitive to users unless you explicitly document it :)

I do think what @ambrosejcarr proposed with `get_image` and `get_images` is a good middle-ground in terms of convenience and intuitiveness.

starfish/core/imagestack/parser/crop.py
```diff
@@ -133,36 +108,104 @@ def _crop_axis(size: int, crop: Optional[Union[int, slice]]) -> Tuple[int, int]:
...
-                crop_params._add_permitted_axes(Axes.ROUND, tile.indices[Axes.ROUND])
-                if Axes.ZPLANE in tile.indices:
-                    crop_params._add_permitted_axes(Axes.ZPLANE, tile.indices[Axes.ZPLANE])
-            else:
-                coord_groups[x_y_coords] = CropParameters(
-                    permitted_chs=[tile.indices[Axes.CH]],
-                    permitted_rounds=[tile.indices[Axes.ROUND]],
-                    permitted_zplanes=[tile.indices[Axes.ZPLANE]] if Axes.ZPLANE in tile.indices
-                    else None)
+            if (rounds is None and chs is None and zplanes is None) \
```
ttung: why is this section (`(rounds is None and chs is None and zplanes is None)`) necessary?

starfish/core/imagestack/parser/crop.py
```diff
@@ -133,36 +108,104 @@ def _crop_axis(size: int, crop: Optional[Union[int, slice]]) -> Tuple[int, int]:
...
+                    crop_params = coord_groups[x_y_coords]
+                    crop_params._add_permitted_axes(Axes.CH, tile.indices[Axes.CH])
+                    crop_params._add_permitted_axes(Axes.ROUND, tile.indices[Axes.ROUND])
+                    if Axes.ZPLANE in tile.indices:
+                        crop_params._add_permitted_axes(Axes.ZPLANE, tile.indices[Axes.ZPLANE])
+                else:
+                    coord_groups[x_y_coords] = CropParameters(
+                        permitted_chs=[tile.indices[Axes.CH]],
+                        permitted_rounds=[tile.indices[Axes.ROUND]],
+                        permitted_zplanes=[tile.indices[Axes.ZPLANE]] if Axes.ZPLANE in tile.indices
```
ttung: consider dropping the RHS to the next line if it improves readability, i.e.,

```
                permitted_zplanes=
                    [tile.indices[Axes.ZPLANE]] if Axes.ZPLANE in tile.indices else None,
```

starfish/core/experiment/experiment.py
```diff
@@ -105,66 +96,16 @@ def name(self) -> str:
     def image_types(self) -> Set[str]:
         return set(self._images.keys())
 
-    def show_aligned_image_groups(self) -> None:
```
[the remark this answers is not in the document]
shanaxel42: I don't think anyone's using this method? I thought we wanted to abstract everything about alignedImageGroups from the user @ambrosejcarr @dganguli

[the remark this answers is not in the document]
shanaxel42: this would also require us to continue parsing aligned groups when we load an FOV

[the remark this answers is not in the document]
ambrosejcarr: 👍 I think we should hide aligned image groups from users, and that this should be an internal detail.

starfish/core/experiment/experiment.py
```diff
@@ -195,15 +136,75 @@ def get_image(self, item: str, aligned_group: int = 0,
...
+        aligned_groups = CropParameters.parse_aligned_groups(self._images[item],
+                                                             rounds=rounds, chs=chs,
+                                                             zplanes=zplanes,
+                                                             x=x, y=y)
+        aligned_stack_iterator = AlignedImageStackIterator(tileset=self._images[item],
+                                                           aligned_groups=aligned_groups)
+        return aligned_stack_iterator
+
+
+class AlignedImageStackIterator:
```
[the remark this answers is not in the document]
shanaxel42: I think Iterable is right, I don't think I could easily  implement a __contains__ for collection

starfish/core/imagestack/parser/crop.py
shanaxel42: If no selected axes are give we want to parse the tile, I could move this logic into `tile_in_selected_axes` though.

starfish/core/imagestack/parser/crop.py
```diff
@@ -133,36 +108,104 @@ def _crop_axis(size: int, crop: Optional[Union[int, slice]]) -> Tuple[int, int]:
...
+        zplanes: Optional[Collection[int]]
+            The allowed zplanes.
+
+        Returns
+        -------
+        Boolean
+            True if tile belongs with selected axes, False if not.
+        """
+        # If all none, no selected axes given
+        if rounds is None and chs is None and zplanes is None:
```
ttung: i think this can be removed.

starfish/core/experiment/experiment.py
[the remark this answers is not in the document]
shanaxel42: added!

starfish/core/experiment/experiment.py
```diff
@@ -195,15 +136,75 @@ def get_image(self, item: str, aligned_group: int = 0,
...
+        aligned_groups = CropParameters.parse_aligned_groups(self._images[item],
+                                                             rounds=rounds, chs=chs,
+                                                             zplanes=zplanes,
+                                                             x=x, y=y)
+        aligned_stack_iterator = AlignedImageStackIterator(tileset=self._images[item],
+                                                           aligned_groups=aligned_groups)
+        return aligned_stack_iterator
+
+
+class AlignedImageStackIterator(Iterable[ImageStack]):
```
ttung: Sorry, my bad, it looks like you're implementing both `Iterator` and `Iterable`.  It's probably good to know exactly which one you want to implement.  Right now, the Iterable implementation is broken as `iter` is supposed to return a new instance every time.

starfish/core/experiment/experiment.py
```diff
@@ -106,72 +99,29 @@ def image_types(self) -> Set[str]:
...
-                f" <starfish.ImageStack "
-                f"r={v._permitted_rounds if v._permitted_rounds else 1}, "
-                f"ch={v._permitted_chs if v._permitted_chs else 1}, "
-                f"z={v._permitted_zplanes if v._permitted_zplanes else 1}, "
-                f"(y, x)={y_size, x_size}>"
-                for k, v in enumerate(groups)
-            )
-            all_groups[name] = f'{info}'
-        pprint.pprint(all_groups)
+        raise DeprecationWarning("This method has been deprecated. Aligned groups are now parsed "
```
ttung: 👍

starfish/core/experiment/experiment.py
```diff
@@ -105,66 +96,16 @@ def name(self) -> str:
...
-        --------
-        Iterator
-            An iterator for the aligned subgroups of the image type
-
-        """
-        for aligned_group, _ in enumerate(self.aligned_coordinate_groups[image_type]):
-            yield self.get_image(item=image_type, aligned_group=aligned_group)
-
-    def get_image(self, item: str, aligned_group: int = 0,
+    def get_image(self, item: str,
```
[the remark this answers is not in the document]
ttung: What's the resolution of this?

starfish/core/experiment/experiment.py
```diff
@@ -195,15 +144,75 @@ def get_image(self, item: str, aligned_group: int = 0,
...
+    """Iterator class of AlignedImageStacks."""
+    def __init__(self, tileset: TileSet, aligned_groups: List[CropParameters]):
+        self.size = len(aligned_groups)
+        self.aligned_groups = iter(aligned_groups)
+        self.tileset = tileset
+
+    def __len__(self):
+        return self.size
+
+    def __iter__(self):
```
ttung: I don't believe this should be implemented as it is deceptive.

starfish/core/experiment/experiment.py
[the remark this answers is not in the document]
shanaxel42: oh I don't think we should continue to have this since we're trying to abstract the idea of aligned groups from the user, I can add it back in if you want to maintain the same parameters though?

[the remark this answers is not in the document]
ttung: If we want to maintain API compatibility, we should accept that parameter, default it to a singleton value, and raise an exception if someone passes in that parameter.

Example:

```
_SINGLETON = object()

def should_warn_if_someone_passes_in_abc(abc=_SINGLETON):
    if abc != _SINGLETON:
        raise RuntimeException("Hey, parameter `abc` is no longer accepted")
```

starfish/core/experiment/test/test_experiment.py
```diff
@@ -111,25 +111,18 @@ def test_crop_experiment():
...
-    assert len(fov.aligned_coordinate_groups['primary']) == 1
-    assert len(fov.aligned_coordinate_groups['nuclei']) == 1
+    primary_images = fov.get_images('primary')
+    nuclei_images = fov.get_images('nuclei')
+    # Assert that only one ImageStack in list
+    assert len(primary_images) == 1
+    assert len(nuclei_images) == 1
 
 
 def test_fov_un_aligned_tileset():
```
ttung: consider adding a test where we call get_images with subsets of round, ch, or zplanes.

PR #1347, opened 2019-05-14
[Easy] Fixing spot detection for labeled axes

shanaxel42:
This came up in my quest for the smFISH wdl.


ttung:
This is an incomplete fix, unfortunately.  :(

The complete fix is a serious PITA to test.  Let me send out the complete fix, and maybe the test afterwards. :)

ttung:
Actually, I like the way you did this such that the selector is labeled.  I'm updating this PR to cover both places where it is needed.

PR #1348, opened 2019-05-14
Create an all-purpose ImageStack factory

ttung:
1. Create a `LocationAwareFetchedTile` class, which is like `FetchedTile`, but is explicitly aware of its location in 5D space.
2. Create an all_purpose.imagestack_factory method that produces an ImageStack with the provided coordinate ranges.
3. Fix existing tests that did not deal with coordinates properly.
4. Fix `imagestack_test_utils.py::verify_physical_coordinates`, which used zplane as an index rather than a value.  In the case of labeled images, this makes a difference.

Test plan: `make -j all`

### DIFF
--- a/starfish/core/imagestack/test/imagestack_test_utils.py
+++ b/starfish/core/imagestack/test/imagestack_test_utils.py
@@ -26,9 +26,8 @@ def verify_physical_coordinates(stack: ImageStack,
                                 expected_y_coordinates: Tuple[float, float],
                                 expected_z_coordinates: Tuple[float, float],
                                 zplane: Optional[int] = None) -> None:
-    """Given an imagestack and a set of coordinate min/max values
-    verify that the physical coordinates on the stack match the expected
-    range of values for each coord dimension.
+    """Given an imagestack and a set of coordinate min/max values verify that the physical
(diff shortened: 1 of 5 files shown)


starfish/core/imagestack/test/imagestack_test_utils.py
```diff
@@ -41,6 +40,8 @@ def verify_physical_coordinates(stack: ImageStack,
     # If zplane provided, test expected_z_coordinates on specific plane.
     # Else just test expected_z_coordinates on entire array
     if zplane is not None:
-        assert np.isclose(stack.xarray[Coordinates.Z.value][zplane], expected_z_coordinates)
+        assert np.isclose(
+            stack.xarray.sel({Axes.ZPLANE.value: zplane})[Coordinates.Z.value],
```
shanaxel42: I don't think you need to do the sel, stack.xarray[Coordinates.Z.value].values should give you the array

shanaxel42: I don't think you need to do the sel, stack.xarray[Coordinates.Z.value].values should give you the array

ttung: I don't want the array.  I want a singular value for the plane.

Issue #1332, opened 2019-05-09
verify physical coordinates are consistent throughout spot finding 


ttung:
1. Document the implications.
2. Transferring physical coordinates to the intensity table should work for the last row/column of pixels.

ttung:
`transfer_physical_coords_from_imagestack_to_intensity_table` is probably also affected.

ttung:
1. if you have a pixel range of (0, 4), corresponding to physical coordinates of (0.0, 0.4), the actual valid range for centroids is (0, 4.9999..).  `interp` will break (return NaN, I believe) if you exceed the range of the original source.  Possible solution: extend the pixel->physical coordinates array by one element; in this case, (0, 5), and `interp` in that range.

2. when a third-party library reports (1, 1) as the pixel coordinates of a spot's centroid, does that mean the centroid is in the top-left corner of the pixel in the 2nd column, 2nd row?

ttung:
@shanaxel42 and I had a long conversation about this today, and we think there are basically two options:

# Example:
- 5x5 image
- center pixel is lit
- physical (xmin, xmax) is (0, 10)

# Option 1
## Rules
- Pixel's origin (0, 0) is at the center of a pixel
- A tile's physical min coordinate is the left/top extent of the tile, and the physical max coordinate is the right/bottom of the tile.

## Implications
- Spot finding output is correct (a 5x5 image with the center pixel lit would have its centroid called as (2, 2).
- Plotting the centroid makes sense (matplotlib plots (2, 2) as the absolute center of the 5x5 image).
- The physical coordinates in the xarray will be based on the physical coordinates of the pixel's center, and thus are somewhat surprising.  The top-left pixel in the 5x5 image will be assigned physical coordinates (1, 1), instead of (0, 0).  The bottom-right pixel in the 5x5 image will be assigned physical coordinates (9, 9) instead of (10, 10).
- Valid pixel coordinate range in the 5x5 example: (0, 5]
- Valid physical coordinate range in the 5x5 example: (0, 10)

## Doodle

![IMG_4335](https://user-images.githubusercontent.com/280924/60850763-9a2fa400-a1a5-11e9-968a-d7672903b7f9.jpg)


# Option 2
## Rules
- Pixel's origin (0, 0) is at the top-left of a pixel.
- A tile's physical min coordinate is the left/top extent of the tile, and the physical max coordinate is the top-left corner of the bottommost/rightmost pixel.

## Implications
[...]

PR #1350, opened 2019-05-15
Fix physical coordinate calculator

ttung:
It should reflect that the physical coordinate of a pixel is the top-left corner of a pixel.  Therefore, the last pixel's coordinate should be the top-left corner and is the max of the tile's range.

Test plan: added a test!

Part of #1332

### DIFF
--- a/starfish/core/imagestack/physical_coordinate_calculator.py
+++ b/starfish/core/imagestack/physical_coordinate_calculator.py
@@ -15,13 +15,11 @@ def _pixel_offset_to_physical_coordinate(
         physical_pixel_size: Number,
         pixel_offset: Optional[int],
         coordinates_at_pixel_offset_0: Number,
-        dimension_size: int,
 ) -> Number:
     """Calculate the physical pixel value at the given index"""
     if pixel_offset:
         # Check for negative index
-        if pixel_offset < 0:
-            pixel_offset = pixel_offset + dimension_size
+        assert pixel_offset >= 0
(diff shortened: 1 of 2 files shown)


starfish/core/imagestack/physical_coordinate_calculator.py
```diff
@@ -53,15 +51,19 @@ def recalculate_physical_coordinate_range(
...
-    max_pixel_index = indexer if isinstance(indexer, int) else indexer.stop
-    # Add one to max pixel index to get end of pixel
-    max_pixel_index = max_pixel_index + 1 if max_pixel_index else dimension_size
-    new_min = _pixel_offset_to_physical_coordinate(
-        physical_pixel_size, min_pixel_index, coord_min, dimension_size)
-    new_max = _pixel_offset_to_physical_coordinate(
-        physical_pixel_size, max_pixel_index, coord_min, dimension_size)
+    if isinstance(indexer, int):
+        max_pixel_index = indexer
+    elif isinstance(indexer.stop, int):
```
shanaxel42: would this ever not be an int?

ttung: I think if it's a range where the end is unbounded, it's a `None`.  In fact, that is not tested in my code. :(

ttung: Updated with a test case for stop=`None`!

PR #1357, opened 2019-05-16
Cap the number of tasks each child executes.

ttung:
This is just a guess at a reasonable number.  This prevents bugs in the memory management from persisting throughout the lifetime of the multiprocessing pool.

Test plan: travis


ttung:
I suspect we do want to cap the number of tasks, but a strict number seems rather useless.  It should really reflect some degree of resource utilization above a newly spawned task, and I don't know if that's possible.  Let's close this for now.

PR #1358, opened 2019-05-16
Always create Z coordinates, even on 4D datasets.

ttung:
When we parse a 4D tileset into an ImageStack, we always set up a ZPLANE and initialize it to 0.  Similarly, we should always create Z coordinates and set it to NaN.


ttung:
> I think there might be one more if Coordinates.Z in tile.coordinates: statements to get rid off in crop.py and on line 183 of Imagestack.py

These convert from TileSets to ImageStacks, and we need to keep them because Z coordinates are not required in tilesets.

shanaxel42:
> > I think there might be one more if Coordinates.Z in tile.coordinates: statements to get rid off in crop.py and on line 183 of Imagestack.py
> 
> These convert from TileSets to ImageStacks, and we need to keep them because Z coordinates are not required in tilesets.

Ah I'm sorry! I completely read this wrong. Ignore me

ttung:
Can this be approved then? :)

PR #1369, opened 2019-05-21
Tooling to support creation of 4D datasets

ttung:
This is a debatable PR.  This adds support to our factories to build 4D datasets.  This makes it easier to test our data loading paths for 4D datasets, but it does come at the cost of significantly complicating the codebase.  I wrote it and some tests, but was reluctant to send it out.  However, before I junk it, I wanted to check with the rest of you. :) 

If we want to test our code against datasets where Z is not provided, we need to be able to build synthetic datasets where Z is not provided.  This is the tooling to make such datasets (not ImageStacks!).  This means the TileSet is fundamentally 4D, with no zplanes, and no z coordinates.

### DIFF
--- a/starfish/core/experiment/builder/__init__.py
+++ b/starfish/core/experiment/builder/__init__.py
@@ -65,8 +65,22 @@ def _fov_path_generator(parent_toc_path: Path, toc_name: str) -> Path:
     return parent_toc_path.parent / "{}-{}.json".format(parent_toc_path.stem, toc_name)
 
 
+<<<<<<< variant A
 def build_irregular_image(
         tile_coordinates: Iterable[TileIdentifier],
+>>>>>>> variant B
+def build_image(
+        fovs: Sequence[int],
+        rounds: Sequence[int],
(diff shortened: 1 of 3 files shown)


ambrosejcarr:
What do you imagine is the use case for this? Is this because SpaceTx Format is technically valid without `r, z` information?

ttung:
> What do you imagine is the use case for this? Is this because SpaceTx Format is technically valid without r, z information?

This would facilitate the testing of `ImageStack`s without multiple zplanes.

ttung:
As I said, its utility is debatable.  We could make zplane mandatory to set.

ambrosejcarr:
I prefer keeping ImageStack 5d in all cases and having flexibility in the format. This is a weak preference though, so if others have stronger feelings I'm happy to defer.

ttung:
This doesn't make ImageStack 4D.  It makes it easy to construct 4D TileSets so we can test the 4D-5D promotion.

ttung:
I'm going to junk this.

Issue #1363, opened 2019-05-17
`IntensityTable.concatenate_intensity_tables` seems oddly slow


njmei:
Hi,

I tried merging `IntensityTable`s from a small test dataset with 3 FOVs (1 round, 3 fov, 3 ch, 37 z, 2048 y, 2048 x)

The 3 `IntensityTable`s have 1707, 1040, and 241 spots respectively. I can also provide netcdf'd versions of the tables if desired.

I did:
```
merged_decoded = IntensityTable.concatenate_intensity_tables([decoded_fov_000, decoded_fov_001, decoded_fov_002], overlap_strategy=OverlapStrategy.TAKE_MAX)
```

This somehow takes a wall time of ~3 minutes.

EDIT:
[decoded_intensity_table_netcdfs.tar.gz](https://github.com/spacetx/starfish/files/3193446/decoded_intensity_table_netcdfs.tar.gz)

ttung:
Yeah, that does seem pretty bad.  tables would be really helpful.  cc: @shanaxel42

shanaxel42:
`overlap_strategy=OverlapStrategy.TAKE_MAX` is doing a bit of computation so it makes sense it wouldn't be as fast as just a normal concat but that does seem especially slow. I can look

PR #1372, opened 2019-05-23
Pre-calculate xmin,xmax,ymin,ymax attrs on IntensityTable before overlap processing 

shanaxel42:
This specifically addresses #1363. The slowdown was due to repeatedly recalculating the min/max physical coordinates on the IntensityTables when trying to find overlaps. By calculating and storing the bounding box coordinates first, `IntensityTable.concatenate_intensity_tables(overlap_strategy=TAKE_MAX)` on the three tables provided by @njmei  was reduced from 479.860 seconds to 148.285 seconds. This should also scale much better (linearly) now since we only do the expensive min/max calculations once per intensity table at the beginning.

### DIFF
--- a/starfish/core/intensity_table/intensity_table.py
+++ b/starfish/core/intensity_table/intensity_table.py
@@ -153,8 +153,6 @@ class IntensityTable(xr.DataArray):
             5D tensor.
         round_values : Sequence[int]
             The possible values for the round number, in the order that they are in the ImageStack
-        bbox: Optional[Mapping[PhysicalCoordinateTypes, float]]
-            The bounding box of physical space the intensity table exists in.
         args :
             Additional arguments to pass to the xarray constructor.
         kwargs :
@@ -434,7 +432,7 @@ class IntensityTable(xr.DataArray):
             intensity_data,
             pixel_coordinates,
             image_stack.axis_labels(Axes.CH),
-            image_stack.axis_labels(Axes.ROUND)
+            image_stack.axis_labels(Axes.ROUND),
         )
 
     @staticmethod
@@ -454,7 +452,6 @@ class IntensityTable(xr.DataArray):
         Find the overlapping sections between IntensityTables and process them according
         to the given overlap strategy
         """
-
         overlap_pairs = find_overlaps_of_xarrays(intensity_tables)
         for indices in overlap_pairs:
             overlap_method = OVERLAP_STRATEGY_MAP[overlap_strategy]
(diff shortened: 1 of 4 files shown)


ttung:
Design question: is there anything stopping someone from mutating the IntensityTable after the bounding boxes are calculated?  If not, the bounding boxes could plausibly become outdated.  Would it make sense to compute the bounding boxes once per concat operation?

Separately, what's the critical path now?

shanaxel42:
> Design question: is there anything stopping someone from mutating the IntensityTable after the bounding boxes are calculated? If not, the bounding boxes could plausibly become outdated. Would it make sense to compute the bounding boxes once per concat operation?
> 
> Separately, what's the critical path now?

I mean the bounding boxes are calculated as part of the `IntensityTable.concatenate_intensity_tables()` call so I'm not sure how someone would mess with the intensity tables in the middle of that method unless they ripped out the code.

ttung:
> I mean the bounding boxes are calculated as part of the IntensityTable.concatenate_intensity_tables() call so I'm not sure how someone would mess with the intensity tables in the middle of that method unless they ripped out the code.

Yeah, I suspect that code path is fine, but the attrs are still on the IntensityTable objects.

shanaxel42:
> > I mean the bounding boxes are calculated as part of the IntensityTable.concatenate_intensity_tables() call so I'm not sure how someone would mess with the intensity tables in the middle of that method unless they ripped out the code.
> 
> Yeah, I suspect that code path is fine, but the attrs are still on the IntensityTable objects.

I can clear them after...or I guess store them separately from the intensity tables. The min/max calls take an incredibly long amount of time on the data nick sent though so calculating them more then once per table is significant.

ttung:
> I can clear them after...or I guess store them separately from the intensity tables. The min/max calls take an incredibly long amount of time on the data nick sent though so calculating them more then once per table is significant.

Why does it take so long to calculate a max of 1700 numbers?  Are we doing something wrong?

shanaxel42:
> > I can clear them after...or I guess store them separately from the intensity tables. The min/max calls take an incredibly long amount of time on the data nick sent though so calculating them more then once per table is significant.
> 
> Why does it take so long to calculate a max of 1700 numbers? Are we doing something wrong?

I'M SUPER DUMB and did not do enough profiling, this was a great point. Should have been using np.max/np.min instead of the built in min/max functions. Closing this PR, another one coming

PR #1375, opened 2019-05-24
Add inplace TileFetcher options back to public builder API

njmei:
This commit addresses an import error encountered
when trying to import inplace TileFetcher functionality.

Though if you want to keep the inplace stuff a `secret`
option for now, that's understandable too...

Example failing import:

```
from starfish.experiment.builder.inplace import (
    InplaceFetchedTile,
    enable_inplace_mode,
    inplace_tile_opener)
```

In the future these should be imported with:

```
from starfish.experiment.builder import (
    InplaceFetchedTile,
    enable_inplace_mode,
    inplace_tile_opener)
```


njmei:
@ttung: Rebased/squashed so that this is only 1 commit. Please merge if everything looks good and tests pass.

PR #1376, opened 2019-05-24
Add custom clip Filter classes

njmei:
This commit adds two custom clip `Filter` classes called:
ClipValueToZero and ClipPercentileToZero

They operate by clipping to a minimum value (or min percentile value)
and then subtracting the entire image with that minimum value.

This will allow us to [create/run](https://spacetx-starfish.readthedocs.io/en/latest/creating_an_image_processing_pipeline/tutorials/exec_running_a_pipeline.html#running-a-pipeline) recipes for our pipeline(s).

### DIFF
--- a/starfish/core/image/_filter/clip_value_to_zero.py
+++ b/starfish/core/image/_filter/clip_value_to_zero.py
@@ -1,0 +1,106 @@
+from functools import partial
+from typing import Optional, Union
+
+import numpy as np
+import xarray as xr
+
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.types import Number
+from starfish.core.util import click
+from ._base import FilterAlgorithmBase
+from .util import determine_axes_to_group_by
+
+
+class ClipValueToZero(FilterAlgorithmBase):
+    """
+    Image clipping filter that clips values below a minimum value and above a
+    maximum value. The filter then subtracts the minimum value from the
+    clipped image.
+
+    By default, the min and max values are set to 0.0 and None respectively,
+    which will result in the filter doing nothing.
+
+    This is a wrapper for :py:func:`numpy.clip`.
+
+    Parameters
+    ----------
+    v_min : float
+        Values below v_min are set to v_min. v_min is then subtracted from the
+        entire image (default 0)
+    v_max : Optional[Number]
+        Values above v_max are set to v_max (default None)
+    is_volume : bool
+        If True, 3d (z, y, x) volumes will be filtered. By default, filter 2D
+        (y, x) tiles
+    """
+
+    def __init__(self,
+                 v_min: float = 0.0,
+                 v_max: Optional[Number] = None,
+                 is_volume: bool = False) -> None:
+        self.v_min = v_min
+        self.v_max = v_max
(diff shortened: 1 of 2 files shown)


starfish/core/image/_filter/clip_value_to_zero.py
```diff
@@ -0,0 +1,105 @@
...
+    clipped image.
+
+    By default, the min and max values are set to 0.0 and None respectively,
+    which will result in the filter doing nothing.
+
+    This is a wrapper for :py:func:`numpy.clip`.
+
+    Parameters
+    ----------
+    v_min : float, optional
```
ttung: Optional in the context of python typing has a different meaning than one might expect.  Consider removing this and just leaving the default.

starfish/core/image/_filter/clip_value_to_zero.py
```diff
@@ -0,0 +1,105 @@
...
+        Values below v_min are set to v_min. v_min is then subtracted from the
+        entire image (default 0)
+    v_max : int
```
ttung: It should be this if your code is the source of truth.

```suggestion
    v_max : Optional[Number]
```

starfish/core/image/_filter/clip_value_to_zero.py
```diff
@@ -0,0 +1,105 @@
...
+        Values above v_max are set to v_max (default None)
+    is_volume : bool
+        If True, 3d (z, y, x) volumes will be filtered. By default, filter 2D
+        (y, x) tiles
+    """
+
+    def __init__(self,
+                 v_min: float = 0.0,
+                 v_max: Optional[float] = None,
```
ttung:
```suggestion
                 v_max: Optional[Number] = None,
```

starfish/core/image/_filter/clip_value_to_zero.py
njmei: Thanks for bringing this to my attention!

starfish/core/image/_filter/clip_value_to_zero.py
njmei: `int` is definitely wrong but I think I'll stick with `Optional[float]` since `Number` is probably too broad (I don't think `np.clip` can handle `Complex` types?)

Also:

> [Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable. This does not handle classes implementing the corresponding ABCs or the fractions.Fraction class, but we believe those use cases are exceedingly rare.](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower)

ttung: `Number` is a starfishism (starfish.core.types.Number) that means int or float.

njmei: oh, thought you were talking about python `Number` class from the `numbers` module.

njmei:
Okay, should be ready for merge now.

PR #1381, opened 2019-05-29
Add concatenate method for ExpressionMatrix

ambrosejcarr:
## Features:
* Add a method to concatenate ExpressionMatrix objects
* Add a test

## Style fixes: 
* Fix constant usage in ExpressionMatrix

## Bug fixes:
* Fix IntensityTable concatenate, which was returning a DataArray instead of an IntensityTable

### DIFF
--- a/starfish/core/expression_matrix/concatenate.py
+++ b/starfish/core/expression_matrix/concatenate.py
@@ -21,7 +21,7 @@ def concatenate(expression_matrices: Iterable[ExpressionMatrix]) -> ExpressionMa
     -------
     ExpressionMatrix :
         Concatenated expression matrix containing all input cells. Missing gene values are filled
-        with np.NaN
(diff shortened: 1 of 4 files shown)


starfish/core/expression_matrix/concatenate.py
```diff
@@ -0,0 +1,32 @@
...
+    Parameters
+    ----------
+    expression_matrices : Iterable[ExpressionMatrix]
+        iterable (list-like) of expression matrices to combine
+
+    Returns
+    -------
+    ExpressionMatrix :
+        Concatenated expression matrix containing all input cells. Missing gene values are filled
+        with np.NaN
```
ttung:
```suggestion
        with np.nan
```

even though I think of it as NaN, I believe it's nan in numpy.

PR #1382, opened 2019-05-30
Support for irregular images in the builder

### LINKED ISSUE #1322: (composite, ragged) experiment and codebook support
# Objective

*As a spaceTX data analyst with an experiment that does not have a matching number of rounds and channels, I want starfish to support this data model, so that  I can create a starfish benchmarking pipeline*

## Acceptance Criteria
* Simone can process his data with starfish
* Brian Long can process his data with starfish

## Notes
Encapsulates work that needs to be done to support data coming in from Ola, Simone, Brian Long, and presumably RNAscope users

ttung:
Both `build_image` and `write_experiment_json` get an irregular flavor.  In each case, the irregular case accepts an iterator of `TileCoordinate` objects, which record where the tile is in the 5D tensor.  `build_image` and `write_experiment_json` internally call their irregular flavors because they are very generalized.

#1374 introduced `write_labeled_experiment_json`, and this replaces that operation.  We are not deprecating that API because it never made it into a release; we will simply replace it.

Test plan: travis

Fixes #1322

### DIFF
--- a/starfish/core/experiment/builder/__init__.py
+++ b/starfish/core/experiment/builder/__init__.py
@@ -1,12 +1,18 @@
+import functools
 import json
 import os
+from dataclasses import astuple, dataclass
 from pathlib import Path
 from typing import (
     Any,
     BinaryIO,
     Callable,
     Dict,
+    Iterable,
     Mapping,
+    MutableMapping,
+    MutableSequence,
+    MutableSet,
     Optional,
     Sequence,
     Union,
@@ -20,6 +26,7 @@ from slicedimage import (
     Writer,
 )
 
+from starfish import FieldOfView
 from starfish.core.codebook.codebook import Codebook
 from starfish.core.experiment.builder.orderediterator import join_axes_labels, ordered_iterator
 from starfish.core.experiment.version import CURRENT_VERSION
@@ -28,11 +35,17 @@ from .defaultproviders import RandomNoiseTile, tile_fetcher_factory
 from .providers import FetchedTile, TileFetcher
 
 
-AUX_IMAGE_NAMES = {
-    'nuclei',
-    'dots',
-}
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
+
+
+@dataclass
+class TileIdentifier:
+    """Data class for encapsulating the location of a tile in a 6D tensor (fov, round, ch, zplane,
+    y, and x)."""
+    fov_id: int
+    round_id: int
+    ch_id: int
+    zplane_id: int
 
 
 def _tile_opener(toc_path: Path, tile: Tile, file_ext: str) -> BinaryIO:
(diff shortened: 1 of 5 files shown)


njmei:
Should users be able to import the `TileCoordinate` dataclass and `write_irregular_experiment_json()` from `starfish.experiment.builder`?

It looks like currently the only way to access is to import from `starfish.core.experiment.builder`.

ttung:
> Should users be able to import the `TileCoordinate` dataclass and `write_irregular_experiment_json()` from `starfish.experiment.builder`?

> It looks like currently the only way to access is to import from `starfish.core.experiment.builder`.

I pushed a fix for this.

njmei:
I think the `TileCoordinate` needs to be importable too from `starfish.experiment.builder`?

ttung:
> I think the TileCoordinate needs to be importable too from starfish.experiment.builder?

Fixed.

ttung:
Sorry, that was my mistake.  I thought this was the issue, and I also thought this PR was landed. :)

starfish/core/experiment/builder/__init__.py
```diff
@@ -28,13 +35,18 @@
...
-AUX_IMAGE_NAMES = {
-    'nuclei',
-    'dots',
-}
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
 
 
+@dataclass
+class TileCoordinate:
+    """Data class for encapsulating the location of a tile in a"""
```
shanaxel42:
```suggestion
    """Data class for encapsulating the location of a tile in an experiment"""
```

or at least I think that's what you meant.

starfish/core/experiment/builder/__init__.py
```diff
@@ -28,13 +35,18 @@
...
 
-AUX_IMAGE_NAMES = {
-    'nuclei',
-    'dots',
-}
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
 
 
+@dataclass
+class TileCoordinate:
```
shanaxel42: I think using the word Coordinates here is a little confusing since we already have a concept of tile coordinate...maybe `TileIndex` or `TileIndentifier`?

starfish/core/experiment/builder/__init__.py
```diff
@@ -52,55 +64,45 @@ def _fov_path_generator(parent_toc_path: Path, toc_name: str) -> Path:
...
-        (default = (Axes.Z, Axes.ROUND, Axes.CH))
 
     Returns
     -------
     The slicedimage collection representing the image.
     """
-    axes_sizes = join_axes_labels(
-        axes_order, rounds=rounds, chs=chs, zplanes=zplanes)
+    def reducer_to_sets(
+            accum: Sequence[MutableSet[int]], update: TileCoordinate,
```
shanaxel42: what's accum?

starfish/core/experiment/builder/__init__.py
ttung: I like `TileIdentifier`.

starfish/core/experiment/builder/__init__.py
ttung: accumulator

starfish/core/experiment/builder/__init__.py
ttung: Fixed.

PR #1383, opened 2019-05-30
Fix smFISH notebook.

ttung:
1. `iterate_image_type` is now deprecated.  Use `get_images` instead.
2. Remove extra call to `get_image(..)`.  The output is not used anywhere.

Test plan: python notebooks/py/smFISH.py


ttung:
Gonna unbreak master. #yolo

PR #1386, opened 2019-05-31
Refactor reduce to take an optional module and only a function name.

ttung:
Module defaults to np (== numpy).  Class no longer accepts a callable.  Alias list is a dict that relatively easy to extend.

Test plan: travis

### DIFF
--- a/starfish/core/image/_filter/reduce.py
+++ b/starfish/core/image/_filter/reduce.py
@@ -1,7 +1,12 @@
+import importlib
+from enum import Enum
 from typing import (
     Callable,
+    cast,
     Iterable,
+    Mapping,
     MutableMapping,
+    Optional,
     Sequence,
     Union
 )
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
A couple high-level comments: 
1. Is there any reason we don't accept `*args and **kwargs` and pass those to the function? I could imagine it being important, for example...
2. Can we have a couple examples in the class documentation? One simple one that uses `"mean"`, and a complex one that uses `scipy.linalg.norm` selecting the `2` norm?

```python
module=scipy.linalg
function="norm"
kwargs={"ord": 2}
```

kevinyamauchi:
This looks good to be. I agree with @ambrosejcarr that some examples in the documentation would be really helpful.

ttung:
I added a couple of tests, a couple of examples, and beefed up the docs.

starfish/core/image/_filter/test/test_reduce.py
```diff
@@ -18,10 +17,7 @@
...
 
 def make_image_stack():
-    '''
-        Make a test ImageStack
-
-    '''
+    """Make a test ImageStack."""
 
     # Make the test image
     test = np.ones((2, 4, 1, 2, 2), dtype='float32') * 0.1
```
shanaxel42: could we use imagestack_factory() here instead? Or maybe add a imagestack_from_numpy method to factories/all_purpose.py

starfish/core/image/_filter/reduce.py
```diff
@@ -17,53 +22,141 @@
...
+                f"{self.top_level_package}")
+
+        np = ("numpy", {'max': 'amax'})
+        """Function source for the numpy libraries"""
+        scipy = ("scipy",)
+
     def __init__(
-        self, dims: Iterable[Union[Axes, str]], func: Union[str, Callable] = 'max',
-        clip_method: Clip = Clip.CLIP
+        self,
```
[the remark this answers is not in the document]
ttung: This is the init for the Reduce class.  There is an init method for the FunctionSource class starting at L103.

starfish/core/image/_filter/reduce.py
```diff
@@ -17,53 +22,141 @@
 
 class Reduce(FilterAlgorithmBase):
     """
-    Reduces the dimensions of the ImageStack by applying a function
-    along one or more axes.
+    Reduces the dimensions of the ImageStack by applying a function along one or more axes.
```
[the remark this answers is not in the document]
ttung: I reworded this.  It now says, "Reduces the cardinality of one or more axes to 1 by applying a function across those axes."

starfish/core/image/_filter/test/test_reduce.py
ttung: > could we use imagestack_factory() here instead? 

Plausibly, though the inputs and outputs are very precise and not very numerous.  Not sure if it makes sense to.

> imagestack_from_numpy

ImageStack.from_numpy is basically that, isn't it?

starfish/core/image/_filter/reduce.py
```diff
@@ -17,53 +22,141 @@
 
 class Reduce(FilterAlgorithmBase):
```
ambrosejcarr: @ttung given how codebases typically mirror map & reduce, I wonder if `apply` should be exposed in a similar fashion.

ttung: Yes, I agree.  I thought there was already an issue for that, but maybe I'm wrong.

It's also possible that Reduce might work for this already.

PR #1390, opened 2019-06-03
[easy] Warn about the deprecation of the MaxProject filter

ttung:
Depends on #1386


ttung:
I can see this being controversial, so I'm flagging a bunch of ppl. :)

PR #1395, opened 2019-06-07
[EASY] updates to readme

neuromusic:


README.rst
```diff
@@ -55,7 +55,7 @@ to see starfish in action on particular assay types. Or, browse our our notebook
 .. _dar: https://github.com/spacetx/starfish/blob/master/notebooks/DARTFISH.ipynb
 .. _str: https://github.com/spacetx/starfish/blob/master/notebooks/STARmap.ipynb
 .. _seq: https://github.com/spacetx/starfish/blob/master/notebooks/SeqFISH.ipynb
-.. _3ds: https://github.com/spacetx/starfish/blob/master/notebooks/3d_smFISH.ipynb
+.. _3ds: https://github.com/spacetx/starfish/blob/master/notebooks/smFISH.ipynb
```
ttung: As I recall, we have another notebook of smFISH in the rendered docs

https://spacetx-starfish.readthedocs.io/en/stable/usage/data_processing_examples/3d_smFISH.html

Not sure which one is more appropriate though.

neuromusic: Everything in the README appears to be pointing to the items in the notebooks folder, so I think this is right for now.

But we should consolidate these into a single set of examples. I'll make an issue.

PR #1410, opened 2019-06-17
[Easy] fix merfish scale factors

shanaxel42:
I changed the starfish.data link for MERFISH to the full dataset so we no longer need two different ways of getting the scale factor info

### DIFF
--- a/notebooks/MERFISH.ipynb
+++ b/notebooks/MERFISH.ipynb
@@ -194,16 +194,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


ttung:
> I changed the starfish.data link for MERFISH to the full dataset so we no longer need two different ways of getting the scale factor info

does that mean the data on the server is now different?

ttung:
I suspect this is the reason master is broken?  https://api.travis-ci.com/v3/job/209190201/log.txt

Going to land this.

PR #1413, opened 2019-06-18
[easy] - fixing all expression matrix saves

### LINKED ISSUE #1411: ExpressionMatrix methods
After target assignment I created an expression matrix (a 2-dimensional cells x genes tensor). 

Screenshot 2019-06-18 at 19 23 41

I need to save it for the following usage with single-cell analysis software packages but ExpressionMatrix methods return these kinds of errors.

matrix.save('expression_matrix')

DataArray.name or Dataset key must be either a string or None for serialization to netCDF files

matrix.save_loom('expression_matrix1')

invalid indexer array, does not have integer dtype: array(<Axes.X: 'x'>, dtype=object)

matrix.save_anndata('expression_matrix2')

invalid indexer array, does not have integer dtype: array(<Axes.X: 'x'>, dtype=object)

shanaxel42:
Should fix #1411, also added small test

### DIFF
--- a/REQUIREMENTS-STRICT.txt
+++ b/REQUIREMENTS-STRICT.txt
@@ -1,16 +1,19 @@
 # You should not edit this file directly.  Instead, you should edit one of the following files (REQUIREMENTS.txt) and run make REQUIREMENTS-STRICT.txt
-attrs==19.1.0
 anndata==0.6.21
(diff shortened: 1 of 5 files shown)


ttung:
note that travis is not green, because the test environment does not have anndata installed.

PR #1415, opened 2019-06-19
Fix more misuse of TileCoordinates

ttung:
It's been renamed to TileIdentifier.

Test plan: I grepped the code for "TileCoordinates" and "tile coor" and verified that all remaining cases are valid.


ttung:
I folded this into #1417

PR #1417, opened 2019-06-20
Improvements to experiment builder

ttung:
1. Make TileIdentifier an immutable dataclass.
2. Use a f-string to format paths in `_tile_opener`
3. Make `fov_path_generator` and `tile_opener` arguments of write*experiment_json default to None.  This allows user code to cleanly call these methods sometimes with overrides without crufty code[a].
4. Improve documentation a teeeeeeeeeeeny bit.
5. Rename remaining `TileCoordinates` to `TileIdentifiers`

Test plan: travis

[a]: If a caller sometimes needs to invoke write_experiment_json with a tile_opener, we can now do:

```
if needs_tile_opener:
    tile_opener = abc
else:
    tile_opener = None

write_experiment_json(blah_blah, tile_opener=tile_opener)
```

Previously, it would require something like:

```
if needs_tile_opener:
    write_experiment_josn(blah_blah, tile_opener=abc)
else:
    write_experiment_josn(blah_blah)
```

Functionally, they are the same, but the latter repeats itself more.

### DIFF
--- a/starfish/core/experiment/builder/__init__.py
+++ b/starfish/core/experiment/builder/__init__.py
@@ -38,7 +38,7 @@ from .providers import FetchedTile, TileFetcher
 DEFAULT_DIMENSION_ORDER = (Axes.ZPLANE, Axes.ROUND, Axes.CH)
 
 
-@dataclass
+@dataclass(eq=True, order=True, frozen=True)
 class TileIdentifier:
     """Data class for encapsulating the location of a tile in a 6D tensor (fov, round, ch, zplane,
     y, and x)."""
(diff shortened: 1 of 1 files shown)


starfish/core/experiment/builder/__init__.py
```diff
@@ -51,13 +51,8 @@ class TileIdentifier:
...
     return open(
-        "{}-Z{}-H{}-C{}.{}".format(
-            str(base),
-            tile.indices[Axes.ZPLANE],
-            tile.indices[Axes.ROUND],
-            tile.indices[Axes.CH],
-            ImageFormat.TIFF.file_ext,
-        ),
+        f"{os.fspath(base)}-Z{tile.indices[Axes.ZPLANE]}-"
+        f"R{tile.indices[Axes.ROUND]}-C{tile.indices[Axes.CH]}.{file_ext}",
```
shanaxel42: this is so much easier to read!

PR #1418, opened 2019-06-20
Fix mypy errors

ttung:
mypy 0.710 has some new errors.  Fix them.  There's one that's spurious, so I'm just blindly ignoring for now.


ttung:
Ninja-landing this because it's making every PR go red.

PR #1421, opened 2019-06-24
data set formatter with fixed filenames

### LINKED ISSUE #1318: Formatting tool for user friendly converting to spaceTX format
- [ ] Define file name convention

- [ ] Have tile fetcher example that works with above convention

- [ ] Document in readthedocs

ttung:
code and docs for creating experiments from tiles with a regular naming schema.

very simple test included, but more testing of the bullet proofing from weird user choices is probably needed.  (bullet proofing is theoretically in the design already).

Depends on #1403, #1409, #1417, #1419 

Fixes #1318

### DIFF
--- a/docs/source/_static/data_formatting_examples/exec_format_structured_data.py
+++ b/docs/source/_static/data_formatting_examples/exec_format_structured_data.py
@@ -1,0 +1,158 @@
+"""
+Format Structured Data
+======================
+
+Format a dataset where the filenames of the tiles provide most of the metadata required to
+organize them into the 5D set (image_type, fov, round, ch, zplane) of 2D tiles  (y, x).  The
+remaining metadata, namely physical coordinates of the tiles, are provided in a CSV file.
+
+Image types generally at least includes 'primary' (defined in the code by
+:py:attr:`~starfish.experiment.experiment.FieldOFView.PRIMARY_IMAGES`).  It may also be 'nuclei' or
+'dots'.
+
+The stem of the tiles' filenames must be
+`<image_type>-f<fov_id>-r<round_label>-c<ch_label>-z<zplane_label>`.  The extension should be one
+of the supported tile formats (:py:class:`~slicedimage.ImageFormat`).  For example, the file
+`nuclei-f0-r2-c3-z33.tiff` would belong to the nuclei image, fov 0, round 2, channel 3, zplane 33.
+
+The CSV file must have a header, and must contain the following columns:
+
+====== ======== ====================================================================================
+Column Required Notes
+------ -------- ------------------------------------------------------------------------------------
+fov    yes
+round  yes
+ch     yes
+zplane yes
+xc_min yes      This should be the minimum value of the x coordinate of the tile.
+xc_max yes      This should be the maximum value of the x coordinate of the tile.
+yc_min yes      This should be the minimum value of the y coordinate of the tile.
+yc_max yes      This should be the maximum value of the y coordinate of the tile.
+zc_min no       This should be the minimum value of the z coordinate of the tile.
+zc_max no       This should be the maximum value of the z coordinate of the tile.
+====== ======== ====================================================================================
+"""
+
+###################################################################################################
+# Create some synthetic data to form into a trivial experiment.
+# -------------------------------------------------------------
+# fov0 contains r0, ch0, zplane0-2.  All tiles share the same physical coordinates.
+# fov1 contains r0-1, ch0, zplane0-2.  All tiles share the same physical coordinates.
+import csv
+import os
+import numpy as np
+import skimage.io
+import tempfile
+
+# columns: r, ch, zplane
+fovs = [
+    [
+        (0, 0, 0),
+        (0, 0, 1),
+        (0, 0, 2),
+    ],
+    [
+        (0, 0, 0),
+        (0, 0, 1),
+        (0, 0, 2),
+        (1, 0, 0),
+        (1, 0, 1),
+        (1, 0, 2),
+    ],
+]
+
+coordinates_of_fovs = [
+    {
+        'xc_min': 0.0,
+        'xc_max': 0.1,
+        'yc_min': 0.0,
+        'yc_max': 0.1,
+        'zc_min': 0.005,
+        'zc_max': 0.005,
+    },
+    {
+        'xc_min': 0.1,
+        'xc_max': 0.2,
+        'yc_min': 0.0,
+        'yc_max': 0.1,
+        'zc_min': 0.005,
+        'zc_max': 0.005,
+    },
+]
+
+data = np.zeros((10, 10), dtype=np.float32)
+
+# create example image tiles that adhere to the file naming schema
+inputdir = tempfile.TemporaryDirectory()
+for fov_id, fov in enumerate(fovs):
+    for round_label, ch_label, zplane_label in fov:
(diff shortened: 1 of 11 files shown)


ttung:
This needs https://github.com/spacetx/slicedimage/pull/100 to be landed and released before travis tests will pass.  But they pass locally, I promise!

ttung:
Here's the generated documentation: https://spacetx-starfish.readthedocs.io/en/tonytung-structured/usage/data_formatting_examples/exec_format_structured_data.html#sphx-glr-usage-data-formatting-examples-exec-format-structured-data-py

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+fov    yes
+round  yes
+ch     yes
+zplane yes
+xc_min yes      This should be the minimum value of the x coordinate of the tile.
+xc_max yes      This should be the maximum value of the x coordinate of the tile.
+yc_min yes      This should be the minimum value of the y coordinate of the tile.
+yc_max yes      This should be the maximum value of the y coordinate of the tile.
+zc_min no       This should be the minimum value of the z coordinate of the tile.
+zc_max no       This should be the maximum value of the z coordinate of the tile.
```
ambrosejcarr: if someone provides only the max_z, does this fail?

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
+"""
+Format Structured Data
+======================
+
+Format a dataset where the filenames of the tiles provide most of the metadata required to
+organize them into the 5D set (image_type, fov, round, ch, zplane) of 2D tiles  (y, x).  The
```
ambrosejcarr: I suggest providing some common options for `image_type` here.

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+# -------------------------------------------------------------
+# fov0 contains r0, ch0, zplane0-2.  All tiles share the same physical coordinates.
+# fov1 contains r0-1, ch0, zplane0-2.  All tiles share the same physical coordinates.
+import csv
+import os
+import numpy as np
+import skimage.io
+import tempfile
+
+fovs = [
```
ambrosejcarr:
```suggestion
# columns: r, ch, zplane
fovs = [
```

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+        'yc_min': 0.0,
+        'yc_max': 0.1,
+        'zc_min': 0.005,
+        'zc_max': 0.005,
+    },
+]
+
+data = np.zeros((10, 10), dtype=np.float32)
+
+inputdir = tempfile.TemporaryDirectory()
```
ambrosejcarr:
```suggestion
# create example image tiles that adhere to the file naming schema
inputdir = tempfile.TemporaryDirectory()
```

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+for fov_id, fov in enumerate(fovs):
+    for round_label, ch_label, zplane_label in fov:
+        path = os.path.join(
+            inputdir.name, f"primary-f{fov_id}-r{round_label}-c{ch_label}-z{zplane_label}.tiff")
+        skimage.io.imsave(path, data)
+
+coordinates_path = os.path.join(inputdir.name, "coordinates.csv")
```
ambrosejcarr:
```suggestion
# write coordinates file
coordinates_path = os.path.join(inputdir.name, "coordinates.csv")
```

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+
+###################################################################################################
+# Construct an experiment out of the raw files.
+# ---------------------------------------------
+
+outputdir = tempfile.TemporaryDirectory()
+from slicedimage import ImageFormat
+from starfish.core.experiment.builder.structured_formatter import format_structured_dataset
+
+format_structured_dataset(
```
ambrosejcarr: This is so beautifully simple. 👍

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
ttung: It does not fail.  If both `zc_min` and `zc_max` are set, we treat it as a range.  If `zc_min` xor `zc_max` are set, then we treat it as a scalar.

Do you think that is wrong?

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
ttung: Will do.

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
neuromusic: +1

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
ambrosejcarr: Nope, that's perfect. :+1:

docs/source/_static/data_formatting_examples/exec_format_structured_data.py
```diff
@@ -0,0 +1,151 @@
...
+of the supported tile formats (:py:class:`~slicedimage.ImageFormat`).  For example, the file
+`nuclei-f0-r2-c3-z33.tiff` would belong to the nuclei image, fov 0, round 2, channel 3, zplane 33.
+
+The CSV file must have a header, and must contain the following columns:
+
+====== ======== ====================================================================================
+Column Required Notes
+------ -------- ------------------------------------------------------------------------------------
+fov    yes
+round  yes
```
dganguli: I'm not sure if the code accounts for this, but Matlab is 1 based indexing, vs. Python 0 based indexing. If a Matlab user uses 1 to mean 0, will your image stack creation still work as expected? If not, we should be clear in documentation that we require 0 based indexing.

dganguli: sorry, i should add that i've noticed Matlab users write out file names where round = 1-4 as opposed to 0-3 and I've had to add -1s in my code to account for this when making the correctly shape numpy/xarrays

ambrosejcarr: This is a great observation. 💯

ttung: Well, what happens is that we write the literal r/ch/z, so it will work.  If you index into the xarray using r=0 when you write out r={1,2,3}, then you will not be a happy camper.  However, if you index into the array using the exact values you wrote out, things will work.

ttung:
I think this is ready for review (mostly by @shanaxel42).

ttung:
I'm going to land this, @shanaxel42 feel free to post-commit review.

PR #1425, opened 2019-06-26
[easy] point starfish.data.osmFISH() to new dataset

### LINKED ISSUE #1294: osmFISH complete data set does not load
It is made of float64 numpy tiles, and has data outside of (0,1).

@ambrosejcarr @dganguli who converted this data set?  can we figure out a good way to fix this?

ttung:
Test plan:
```
exp['fov_000'].get_image('primary')
exp['fov_001'].get_image('primary')
exp['fov_002'].get_image('primary')
```

Fixes #1294

### DIFF
--- a/starfish/data.py
+++ b/starfish/data.py
@@ -127,7 +127,7 @@ def osmFISH(use_test_data: bool = False) -> Experiment:
             "experiment.json"
         )
     return Experiment.from_json(
-        "https://d2nhj9g34unfro.cloudfront.net/20181031/osmFISH/experiment.json")
+        "https://d26bikfyahveg8.cloudfront.net/osmFISH/formatted/20190626/experiment.json")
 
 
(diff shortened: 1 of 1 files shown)


shanaxel42:
> ```python
> imgs = experiment["fov_000"].get_image(FieldOfView.PRIMARY_IMAGES)
> ```

It looks like each FOV has more then one alignedImageGroup. Using `imgs = exp["fov_000"].get_images('primary')` `get_images` instead of `get_image` will give you an iterator object over all of them

kevinyamauchi:
Ah got it! Thanks, @shanaxel42! In the future, how do I know to use `get_images()` instead of `get_image()`?

ttung:
You should always use get_images(..).  get_image(..) is only there for compatibility reasons.

PR #1430, opened 2019-06-28
update napari to 0.1.3

### LINKED ISSUE #998: Is there a function that returns the outline of a segmented cell?
For example, after using Segmentation.Watershed to segment a cell, is there some function that takes the resulting np.ndarray as an input and returns another np.ndarray with the outlines of all segmented cells. This doesn't have to be the exact method; anything similar will do. 

I have tried looking around already for something similar, but I haven't been able to find it. If that function doesn't exist yet, I would like to help to write one since I need that function anyway.

Thank you!

Edit: To give some background to this, the reason I want to do this is because I want to detect and count dots that are only contained within cells, so I would use the outlines of cells from Segmentation.Watershed to crop the dots image to only include those outlined areas, and then count dots in those parts of the image. It is entirely possible that there is a function that does that already that I have just not been able to find yet. If that exists yet, please let me know.

### LINKED ISSUE #1422: Visualize Segmentation mask with original Images with starfish.display
Addresses needs in #1414

kne42:
- update napari version to `0.1.3`
- add segmentation masks visualization support (resolves #1422)

closes #998

### DIFF
--- a/starfish/core/_display.py
+++ b/starfish/core/_display.py
@@ -7,15 +7,16 @@ import numpy as np
 
 from starfish.core.imagestack.imagestack import ImageStack
 from starfish.core.intensity_table.intensity_table import IntensityTable
+from starfish.core.segmentation_mask import SegmentationMaskCollection
 from starfish.core.types import Axes, Features
 
 try:
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", module="vispy.visuals.isocurve", lineno=22)
-        from napari import Window, Viewer
+    from napari import Viewer
 except ImportError:
-    Window, Viewer = None, None
+    Viewer = None
 
+
+NAPARI_VERSION = "0.1.3"  # when changing this, update docs in display
 INTERACTIVE = not hasattr(__main__, "__file__")
 
 
@@ -100,17 +101,17 @@ def _spots_to_markers(intensity_table: IntensityTable) -> Tuple[np.ndarray, np.n
     # the cartesian product of (r, c) are created, and tiled once per feature (n_features)
     # we ensure that (ch, round) cycle in c-order to match the order of the linearized
     # array, used below for masking.
-    coords[:, 0] = np.repeat(intensity_table[Features.AXIS][Axes.X.value].values, code_length)
-    coords[:, 1] = np.repeat(intensity_table[Features.AXIS][Axes.Y.value].values, code_length)
-    coords[:, 2] = np.tile(np.tile(range(n_rounds), n_channels), n_features)
-    coords[:, 3] = np.tile(np.repeat(range(n_channels), n_rounds), n_features)
-    coords[:, 4] = np.repeat(intensity_table[Features.AXIS][Axes.ZPLANE.value].values, code_length)
+    coords[:, 0] = np.tile(np.tile(range(n_rounds), n_channels), n_features)
+    coords[:, 1] = np.tile(np.repeat(range(n_channels), n_rounds), n_features)
+    coords[:, 2] = np.repeat(intensity_table[Features.AXIS][Axes.ZPLANE.value].values, code_length)
+    coords[:, 3] = np.repeat(intensity_table[Features.AXIS][Axes.Y.value].values, code_length)
+    coords[:, 4] = np.repeat(intensity_table[Features.AXIS][Axes.X.value].values, code_length)
 
     sizes = np.repeat(intensity_table.radius.values, code_length)
-    xy = np.tile(sizes[:, np.newaxis], (1, 2))
     rc = np.zeros((sizes.shape[0], 2), dtype=int)
     z = sizes[:, np.newaxis]
-    sizes = np.concatenate((xy, rc, z), axis=1)
+    yx = np.tile(sizes[:, np.newaxis], (1, 2))
+    sizes = np.concatenate((rc, yx, z), axis=1)
 
     return coords, sizes
 
@@ -118,6 +119,7 @@ def _spots_to_markers(intensity_table: IntensityTable) -> Tuple[np.ndarray, np.n
 def display(
         stack: Optional[ImageStack] = None,
         spots: Optional[IntensityTable] = None,
+        masks: Optional[SegmentationMaskCollection] = None,
         viewer: Optional[Viewer] = None,
         project_axes: Optional[Set[Axes]] = None,
         mask_intensities: float = 0.,
@@ -125,7 +127,7 @@ def display(
         z_multiplier: float = 1
 ):
     """
-    Displays an image stack and/or detected spots using Napari (https://github.com/napari/Napari).
+    Displays an image stack and/or detected spots using napari (https://github.com/napari/napari).
 
     Parameters
     ----------
@@ -133,6 +135,8 @@ def display(
         ImageStack to display
     spots : IntensityTable
         IntensityTable containing spot information that was generated from the submitted stack.
+    masks : SegmentationMaskCollection
+        Segmentation instance masks used to annotate the submitted stack.
     viewer : napari.Viewer
         Napari viewer to append the ImageStack and/or spots to. If None, creates a new viewer.
         Note: appending is only supported in interactive environments.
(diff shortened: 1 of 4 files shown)


kne42:
when reviewing, please check that the axis ordering is right. not sure if I did it correctly

(example with ISS pipeline)
Screen Shot 2019-06-28 at 3 02 11 PM

starfish/core/_display.py
```diff
@@ -190,49 +194,52 @@ def display(
...
-    if Window is None or Viewer is None:
-        warnings.warn("Requires napari 0.0.6. Run `pip install starfish[napari]` to install the "
-                      "necessary requirements.")
-        return
-
-    from PyQt5.QtWidgets import QApplication
+    try:
+        import napari
+        if napari.__version__ != NAPARI_VERSION:
+            raise Exception
```
ttung: I would prefer that this be a specific exception type, or an exception with a specific payload that can be inspected to insure that it's what we expect.

starfish/core/_display.py
```diff
@@ -190,49 +194,52 @@ def display(
...
+    except Exception:
```
ttung: This is overly broad.  If there are other issues (perhaps `napari.__version__` doesn't exist?), it is improperly caught.

starfish/core/segmentation_mask.py
```diff
@@ -154,6 +170,38 @@ def from_label_image(
 
         return cls(masks)
 
+    def to_label_image(
+            self,
+            shape: Optional[Tuple[int]] = None,
```
ttung:
```suggestion
            shape: Optional[Tuple[int, ...]] = None,
```

ttung: Would it be accurate to describe `shape` as kind of a crop operation?

starfish/core/segmentation_mask.py
```diff
@@ -154,6 +170,38 @@ def from_label_image(
...
+            *,
+            ordering: List[Axes] = [Axes.ZPLANE, Axes.Y, Axes.X]
```
ttung:
```suggestion
            ordering: Sequence[Axes] = (Axes.ZPLANE, Axes.Y, Axes.X)
```

rationale: default parameter values should be immutable, as they are only created once. if the method modifies `ordering`, then on the subsequent run, `ordering` will have the wrong default value.

```
In [1]: def method(default_param=[1, 2, 3]):
   ...:     default_param.append(4)
   ...:     print(default_param)
   ...:     

In [2]: method()
[1, 2, 3, 4]

In [3]: method()
[1, 2, 3, 4, 4]

In [4]: 
```

starfish/core/segmentation_mask.py
```diff
@@ -101,6 +111,12 @@ def append(self, mask: xr.DataArray):
         validate_segmentation_mask(mask)
         self._masks.append(mask)
 
+        for axis in Axes:
+            if axis.value in mask.coords:
+                max_val = mask.coords[axis.value].values[-1]
+                if self.max_shape[axis] is None or max_val > self.max_shape[axis]:
```
ttung:
```suggestion
                if self.max_shape[axis] is None or max_val >= self.max_shape[axis]:
```

if the max_val is the same as max_shape, i think you need to up max_shape.

starfish/core/_display.py
```diff
@@ -258,13 +265,21 @@ def display(
...
+            coords=coords,
+            face_color="red",
+            edge_color="red",
+            symbol="ring",
+            size=sizes * radius_multiplier,
+            n_dimensional=True,
+            name='spots'
+        )
+
+    if masks is not None:
```
shanaxel42: could be simplified to just `if masks`

starfish/core/_display.py
kne42: This is intentional since we pinned napari to a specific version. `napari.__version__` not existing or some other unforeseen error occurring would mean that it is not behaving as expected and must not be the pinned version. Do you still think this is bad?

starfish/core/segmentation_mask.py
kne42: Ah, that's not what I was thinking of it as, and it would probably error if so. I was thinking more of when someone wanted to superimpose it on the original image so the shapes would match. Is this useful functionality?

starfish/core/segmentation_mask.py
kne42: this method doesn't mutate the default argument but I suppose it's better practice so I'll do it

starfish/core/segmentation_mask.py
kne42: wait could you explain this more? I don't quite understand

kne42:
Thanks for the reviews! Will be addressing soon

starfish/core/segmentation_mask.py
```diff
@@ -101,6 +111,12 @@ def append(self, mask: xr.DataArray):
         validate_segmentation_mask(mask)
         self._masks.append(mask)
 
+        for axis in Axes:
+            if axis.value in mask.coords:
+                max_val = mask.coords[axis.value].values[-1]
```
ambrosejcarr: why not the last coords value?

starfish/core/_display.py
ttung: I always prefer to catch the most specific exception possible because I don't want to mask some other error that may warrant a different fix.  In this case, your point makes a lot of sense for end users.

Two places this breaks down:
1. A developer who is experimenting with the starfish and napari bridge may find this problematic however.  Imagine that I have napari installed from a working tree, and I'm mucking around with the code there.  I could introduce a python parse error that this completely masks.
2. Someone introduces more code into this try-block, and doesn't think through the ramifications of having a catch-all exception handler around it.

tl;dr, I'm ok with you leaving this as is, because it's not a terribly common scenario that someone would encounter.  Be aware of doing general catch-all exception handlers though.

starfish/core/segmentation_mask.py
ttung: So let's say max_vals are (15, 16, 13, 10).  The max shape, in this case, ought to be 17.

However, you would set max_shape to 16 when evaluating max_val=15, and then when you get to max_val=16, you won't bump it to 17.

starfish/core/segmentation_mask.py
ttung: ohhh, i get it.  this _is_ useful functionality.

starfish/core/_display.py
kne42: What do you think about a catch-all that prepends the raised error message prior to the spiel about an incorrect napari version?

ttung: If I were building this, I would do, (in pseudocode):

```
try:
    import napari
except ImportError:
    raise ImportError(f"Requires napari {NAPARI_VERSION...")

try:
    version = napari.__version__
except Exception as ex:
    raise Exception("Could not identify napari version") from ex

if version != NAPARI_VERSION:
        raise Exception(f"Incorrect version of napari installed ({version}).  Please run ...")
```

I think your proposed solution works too, or you could just chain the exception:

```
except Exception as ex:
    raise ImportError(f"...") from ex
```

starfish/core/_display.py
```diff
@@ -258,13 +265,21 @@ def display(
...
 
-        viewer.add_markers(coords=coords, face_color="red", edge_color="red", symbol="ring",
-                           size=sizes * radius_multiplier, n_dimensional=True)
-
-    if new_viewer:
-        window.show()
-
-        if not INTERACTIVE:
-            app.exec()  # create blocking process to persist windows
+        viewer.add_markers(
```
sofroniewn: now `add_points`

starfish/core/_display.py
```diff
@@ -190,49 +194,52 @@ def display(
...
+            Axes.ZPLANE.value,
+            Axes.Y.value,
+            Axes.X.value
         ).values
 
-        layer = viewer.add_image(np.squeeze(reordered_array), multichannel=False)
+        layer = viewer.add_image(np.squeeze(reordered_array),
+                                 multichannel=False,
+                                 name='stack')
         layer.colormap = "gray"
```
sofroniewn: now colormap default is gray, but can also pass `colormap` keyword arg to the above add_image function

starfish/core/_display.py
sofroniewn: also there is no longer a keyword arg `coords`, instead the first arg is the just data array, (can still be a variable named coords)

starfish/core/segmentation_mask.py
kne42: you mean `mask.coords[axis.value][-1]`? it's because xarray is weird so a singular value returned is still an xarray, so we access it via the numpy array instead

kne42:
while we wait on `0.1.2` to release I've updated to `0.1.1`. make sure that your vispy version is `0.5.3` before testing

the intensity table does not appear to be rendering correctly in the ISS notebook; am I doing something wrong? (cc @ambrosejcarr @sofroniewn)

kne42:
@ambrosejcarr @ttung this is now ready for re-review and merge if everything looks okay to you :)

ttung:
travis is still red fyi.

dganguli:
It looks like Napari is handling overlapping cells in a funky way:

image

@sofroniewn seems to know what's up

sofroniewn:
@kne42 so the issue in `to_label_image` is that with each new mask you are overwriting with the entire  area covered by the bounding box, whereas you want to just overwrite the area where the mask is not zero. Shouldn't be too hard to fix

starfish/core/segmentation_mask.py
```diff
@@ -154,6 +170,43 @@ def from_label_image(
...
+            raise ValueError("shape less than the maximum of the data provided."
+                             "cropping is not supported at this time")
+
+        label_image = np.zeros(shape)
+
+        for i, mask in enumerate(self):
+            s = tuple(slice(mask.coords[o].values[0], mask.coords[o].values[-1] + 1)
+                      for o in ordering if o in mask.coords)
+            # align axes and insert into image
+            label_image[s] = (mask
```
sofroniewn: here need to just overwrite area where mask is True, not whole bounding box as defined in `s`

sofroniewn: Would this work

```python
        label_image = np.zeros(shape)

         for i, mask in enumerate(self):
            s = tuple(slice(mask.coords[o].values[0], mask.coords[o].values[-1] + 1)
                      for o in ordering if o in mask.coords)
            # align axes and insert into image
             temp_label_image = np.zeros(shape)

            temp_label_image[s] = (mask.transpose(*[o.value for o in ordering if o in mask.coords]).where(mask == False, i + 1))  # noqa: E712

label_image[temp_label_image > 0] = i + 1
```

kne42: that would work but it's memory inefficient

sofroniewn: sure - you can also convert everything to linear indices and then add the appropriate offsets based on the `mask.coords[o].values[0]` values

kne42: yeah, that's true

kne42: I've just updated, tell me if that fixes it

dganguli: it's fixed! thanks!

Issue #1435, opened 2019-07-03
multiprocessing pool processes spawned within combine_adjacent_features persist till parent process terminates


richardque:
When running my DARTFISH pipeline for multiple FOVs within an experiment, I noticed that multiple processes would continue to spawn with each additional FOV that was processed. I believe I tracked down the culprit to _create_spot_attributes within combine_adjacent_features.py, lines 325-335, a multiprocessing pool is called but doesn't seem to have a termination. Moving it within a with statement appeared to solve the issue for me. 
Was the pool left running by design and necessary for other functions?

ttung:
This is a bug.  Thanks for the report!

PR #1436, opened 2019-07-04
Instantiate the multiprocessing pool using `with`

ttung:
This ensures we shut down the pool when complete.

Test plan: `pytest -v -n4  starfish/core/spots/`
Fixes: #1435


berl:
+many
I need this to run a loop over multiple FOVs.  It fills up my system volume and errors out with `[Errno 28] No space left on device`

PR #1437, opened 2019-07-05
Starfish pre-alpha -> beta

ttung:
Traditionally alpha = in house use, beta = external testers.

Test plan: lol

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,7 +29,7 @@ For detailed information on installation and usage, see the documentation_.
 Quickstart
 ------------
 
-We are currently in **pre-alpha**, finishing proof of concept pipelines for each of the spaceTx
+We are currently in **beta**, finishing proof of concept pipelines for each of the spaceTx
 contributors that leverage starfish's shared object model. Follow the links in the table below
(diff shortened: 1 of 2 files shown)


neuromusic:
Are we still in beta? If so, when are we out of beta?

ambrosejcarr:
Depending on what our target feature set is, there could be arguments for all kinds of phasing for starfish.

I'd vaguely argue that we have key features that we still want to add, and some questions around acquisition and architecture that block us from being "stable".

We feel like we're in "limited beta" to me, simply because each user tends to need some attention to be able to adapt starfish to their use cases.

I'd say we're out of beta when we have a product we're confident our target users can use to accomplish the product's supported use cases with limited developer support.

ttung:
I prefer just plain old beta.

1. I don't want to discourage someone from picking it up because they think limited = google wave, where you have to be invited and all that.
2. It's not entirely clear what limited is.  It can mean different things to different people.

We are definitely not out of beta.

ambrosejcarr:
> I prefer just plain old beta.

Yes I think this is the right way to communicate our state to new users. 👍

PR #1456, opened 2019-07-18
on-demand loading of data.

ttung:
The imagestack's xarray is created with coordinates and NaNs for the values.  When someone accesses the `xarray` property, `get_slice()`, `set_slice()` (except when called by the code initializing the data array), `transform()`, `sel()`, or `max_proj()`, the array is initialized with the correct data.

Theoretically, this allows someone to initialize the ImageStack and examine the coordinates without having the tile data read.

Test plan: pass existing tests.

### DIFF
--- a/starfish/core/image/Filter/reduce.py
+++ b/starfish/core/image/Filter/reduce.py
@@ -180,7 +180,7 @@ class Reduce(FilterAlgorithmBase):
         """
 
         # Apply the reducing function
-        reduced = stack._data.reduce(
+        reduced = stack.xarray.reduce(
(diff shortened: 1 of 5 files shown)


ambrosejcarr:
> Theoretically, this allows someone to initialize the ImageStack and examine the coordinates without having the tile data read.

But there are no memory advantages to this approach, right?

ttung:
> But there are no memory advantages to this approach, right? This is purely a speed optimization.

That's correct.  To get memory advantages, we would need to do more work or the underlying libraries will need more work to better support sparse arrays.

PR #1457, opened 2019-07-18
Fixing travis build 

shanaxel42:
removing starfish build command from `make docker` step

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ slow: fast run-notebooks test-examples docker
 
 docker:
 	docker build -f docker/Dockerfile -t $(DOCKER_IMAGE) .
-	starfish --help
+	docker run -ti --rm $(DOCKER_IMAGE) --help
 
 
 help-integration:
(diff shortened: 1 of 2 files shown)


shanaxel42:
> can we add something to verify that the docker image is sane? even `starfish --help` might be ok?

added!

Makefile
```diff
@@ -115,7 +115,8 @@ slow: fast run-notebooks test-examples docker
 
 docker:
 	docker build -f docker/Dockerfile -t $(DOCKER_IMAGE) .
-	docker run -ti --rm $(DOCKER_IMAGE) build --fov-count 1 --primary-image-dimensions '{"z": 1}' /tmp/
+	starfish --help
```
ttung: I think you would want to run this in the docker container.

```suggestion
docker run -ti --rm spacetx/starfish -- --help
```

shanaxel42: yuuuuuup my b

PR #1465, opened 2019-07-23
simplify docs

neuromusic:
This PR simplifies the docs by moving contact info to the front page and the SpaceTx consortium to Level 1


neuromusic:
also added a badge for slack & revised contact section on README

ambrosejcarr:
I'm wondering what this does to `contributing.rst` since we're deleting the index, but otherwise this looks fine to me.

neuromusic:
@ambrosejcarr ah! I forgot to note and justify that. 

this leaves `CONTRIBUTING.rst`, which is what github surfaces when submitting issues/PRs. I think it makes sense to leave this out of the docs, since we can expect devs to look in the repo for it.

ambrosejcarr:
I think I'd prefer to keep a reference to the contributing guidelines in the docs somewhere. I only checked a few, but at least pandas, numpy, scipy, and sklearn all have references to contribution somewhere in their docs. 

@ttung @shanaxel42 what do you two think?

neuromusic:
@ambrosejcarr that's a good point. since github surfaces CONTRIBUTING.rst for issues, as well, it might make more sense to do something there more like sklearn: a high level summary with links to the specific "developer guide"-style contributing.

ok, I'll add back in the contributing and we can address de-duping CONTRIBUTING.rst in a later PR

PR #1489, opened 2019-08-13
Adding DecodedIntensityTable class

shanaxel42:
Introducing DecodedIntensityTable which is just a subclass of IntensityTable. I moved anything related to decoded spots into the DecodedIntensityTable class. This is mostly to reduce confusion in the codebase between the outputs of various spot finding and decoding methods. In most cases the very end of a pipeline will now be a DecodedIntensityTable.

PR #1515, opened 2019-08-29
Adding new modules and data SpotFindingResults class. 

shanaxel42:
This PR lays some of the ground work for the spot finding refactor. 

- Adds two new modules FindSpots and DecodeSpots (which will be the only ones left by the end of this).
- Adds their base class api contracts. 
- Adds new class SpotFindingResults which is a wrapper class around a dictionary of round/ch index pairs and the spots found there (SpotAttributes).

Depends on #1489

### DIFF
--- a/docs/source/api/spots/index.rst
+++ b/docs/source/api/spots/index.rst
@@ -38,6 +38,18 @@ Spot Detectors can be imported using ``starfish.spots.DetectSpots``, which regis
 .. autoclass:: starfish.spots.DetectSpots
     :members:
 
+NOTE: Starfish is embarking on a SpotFinding data structures refactor see `Spot Finding Refactor Plan`_
+DetectSpots will be replaced by the FindSpots module documented below.
+
+.. _`Spot Finding Refactor Plan`: https://github.com/spacetx/starfish/issues/1514
+
+.. code-block:: python
+
+    from starfish.spots import FindSpots
+
+.. automodule:: starfish.spots.FindSpots
+    :members:
+
 
 .. _decoding:
 
@@ -51,6 +63,15 @@ Decoders can be imported using ``starfish.spots.Decode``, which registers all cl
     from starfish.spots import Decode
 
 .. autoclass:: starfish.spots.Decode
+   :members:
+
+NOTE: Starfish is embarking on a SpotFinding data structures refactor see `Spot Finding Refactor Plan`_
+
+.. code-block:: python
+
+    from starfish.spots import DecodeSpots
+
+.. automodule:: starfish.spots.DecodeSpots
    :members:
 
 
--- a/starfish/core/spots/FindSpots/_base.py
+++ b/starfish/core/spots/FindSpots/_base.py
@@ -1,0 +1,27 @@
+from abc import abstractmethod
+from typing import Callable, Optional, Sequence
+
+import numpy as np
+
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.pipeline.algorithmbase import AlgorithmBase
+from starfish.core.types import Number, SpotFindingResults
+
+
+class FindSpotsAlgorithm(metaclass=AlgorithmBase):
+
+    @abstractmethod
+    def run(self, image_stack: ImageStack,
+            reference_image: Optional[ImageStack] = None, *args) -> SpotFindingResults:
+        """Find and measure spots across rounds and channels in the provided ImageStack."""
+        raise NotImplementedError()
+
+    @staticmethod
+    def _get_measurement_function(measurement_type: str) -> Callable[[Sequence], Number]:
+        try:
+            measurement_function = getattr(np, measurement_type)
+        except AttributeError:
+            raise ValueError(
+                f'measurement_type must be a numpy reduce function such as "max" or "mean". '
+                f'{measurement_type} not found.')
+        return measurement_function
--- a/starfish/core/types/__init__.py
+++ b/starfish/core/types/__init__.py
@@ -15,6 +15,7 @@ from ._constants import (
 )
 from ._decoded_spots import DecodedSpots
 from ._spot_attributes import SpotAttributes
+from ._spot_finding_results import SpotFindingResults
 
 Number = Union[int, float]
 CoordinateValue = Union[Number, Tuple[Number, Number]]
--- a/starfish/core/types/_spot_finding_results.py
+++ b/starfish/core/types/_spot_finding_results.py
@@ -1,0 +1,86 @@
+from typing import List, Mapping, MutableMapping, Optional, Tuple
+
+from starfish.core.types import Axes, SpotAttributes
+
+AXES_ORDER = (Axes.ROUND, Axes.CH)
+
+
+class SpotFindingResults:
+    """
+    Wrapper class that describes the results from a spot finding method. The results
+    mapping is a collection of (round, ch) indices and their corresponding measured
+    SpotAttributes.
+    """
+
+    def __init__(self, spot_attributes_list: Optional[List[Tuple]] = None):
+        """
+        Construct a SpotFindingResults instance
+
+        Parameters
+        -----------
+        spot_attributes_list : Optional[List[Tuple]]
+            If spots were found using ImageStack.transform() the result is a list of
+            tuples ((r, channel), SpotAttributes). Instantiating SpotFindingResults with
+            this list will convert the information to a dictionary.
+        """
+        spot_attributes_list = spot_attributes_list or []
+        self._results: MutableMapping[Tuple, SpotAttributes] = {
+            indices: spots
+            for indices, spots in spot_attributes_list
+        }
+
+    def __setitem__(self, indices: Mapping[Axes, int], spots: SpotAttributes):
+        """
+        Add the round, ch indices and corresponding SpotAttributes to the results dict.
+
+        Parameters
+        ----------
+        indices: Mapping[Axes, int]
+            Mapping of Axes to int values
+        spots: SpotAttributes
+            Describes spots found on this tile.
+        """
+        round_ch_index = tuple(indices[i] for i in AXES_ORDER)
+        self._results[round_ch_index] = spots
+
+    def __getitem__(self, indices: Mapping[Axes, int]) -> SpotAttributes:
+        """
+        Returns the spots found in a given round and ch.
+
+        Parameters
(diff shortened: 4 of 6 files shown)


starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,86 @@
...
+            If spots were fond using Imagestack.transform() the result is a list of
+            tuples (indices, SpotAttributes). Instantiating SpotFindingResults with
+            this list will convert the information to a dictionary.
+        """
+        self._results: Mapping[Tuple, SpotAttributes] = dict()
+        if spot_attributes_list:
+            for indices, spots in spot_attributes_list:
+                self._results[indices] = spots
+
+    def set_spots_for_round_ch(self, indices: Mapping[Axes, int], spots: SpotAttributes
```
ttung: nit: this function declaration style leads to unnecessary LOC changes.  would prefer:

```
    def set_spots_for_round_ch(self, indices: Mapping[Axes, int], spots: SpotAttributes) -> None:
```

or

```
    def set_spots_for_round_ch(
        self, indices: Mapping[Axes, int], spots: SpotAttributes) -> None:
```

or 

```
    def set_spots_for_round_ch(
        self, indices: Mapping[Axes, int], spots: SpotAttributes,
    ) -> None:
```

All of these approaches do not produce unnecessary changes if the method name changes.

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Parameters
+        ----------
+        indices: Mapping[Axes, int]
+            Mapping of Axes to int values
+        spots: SpotAttributes
+            Describes spots found on this tile.
+        """
+        round_ch_index = tuple(indices[i] for i in AXES_ORDER)
+        self._results[round_ch_index] = spots  # type: ignore
```
ttung: You can get rid of the `# type: ignore` if you make self._results a `MutableMapping`.

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        self._results: Mapping[Tuple, SpotAttributes] = dict()
+        if spot_attributes_list:
+            for indices, spots in spot_attributes_list:
+                self._results[indices] = spots
+
+    def set_spots_for_round_ch(self, indices: Mapping[Axes, int], spots: SpotAttributes
+                               ) -> None:
+        """
+        Add the r,ch indices and corresponding SpotAttributes to the results dict.
```
ttung:
```suggestion
        Add the round, ch indices and corresponding SpotAttributes to the results dict.
```

to match what you are doing elsewhere.

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,86 @@
...
+        Return all SpotAttributes across rounds and chs.
+        """
+        return self._results.values()
+
+    @property
+    def round_labels(self):
+        """
+        Return the set of Round labels in the SpotFindingResults
+        """
+        return list(set(sorted(r for (r, ch) in self.round_ch_indices())))
```
ttung: if you take a set of sorted, then you lose the sorting.  did you want sorted(set(...))?

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @property
+    def ch_labels(self):
+        """
+        Return the set of ch labels in the SpotFindingResults
+        """
+        return list(set(sorted(ch for (c, ch) in self.round_ch_indices())))
```
ttung: ditto.

starfish/core/types/__init__.py
```diff
@@ -15,6 +15,7 @@
 )
 from ._decoded_spots import DecodedSpots
 from ._spot_attributes import SpotAttributes
+from ._spot_finding_resutls import SpotFindingResults
```
ttung:
```suggestion
from ._spot_finding_results import SpotFindingResults
```

(and as such, the file probably needs to be renamed)

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,88 @@
+from typing import List, Mapping, MutableMapping, Optional, Tuple
+
+from starfish.core.types import Axes, SpotAttributes
+
+AXES_ORDER = (Axes.ROUND, Axes.CH)
+
+
+class SpotFindingResults:
```
ambrosejcarr: Given this wraps a dict, I would prefer this be implemented to mimic the dictionary container, and more clearly document what it contains. This should simplify a lot of the accessors you've written. For example, 

```diff
+     def values(self):
-     def all_spots(self):
-         """
-         Return all SpotAttributes across rounds and chs.
-         """
          return self._results.values()
```

See: https://docs.python.org/2.5/ref/sequence-types.html

ttung: Couple design questions:

1. Is it ever the case we would want to do spot finding on a zplane basis?
2. Why not just pass in round, ch as arguments to the methods in this class?

starfish/core/types/_spot_finding_resutls.py
```diff
@@ -0,0 +1,88 @@
...
+        Return the set of Round labels in the SpotFindingResults
+        """
+        return list(sorted(set(r for (r, ch) in self.round_ch_indices())))
+
+    @property
+    def ch_labels(self):
+        """
+        Return the set of ch labels in the SpotFindingResults
+        """
+        return list(sorted(set(ch for (c, ch) in self.round_ch_indices())))
```
ttung:
```suggestion
        return list(sorted(set(ch for (r, ch) in self.round_ch_indices())))
```

starfish/core/types/_spot_finding_resutls.py
shanaxel42: Yea so I started down this route but found that things became too confusing. For example get_spots_for_round_ch would instead just be  __getitem__(key) and then when you used it it was totally clear what you were getting. I'm not super opinionated on this though so open to changing it back if that's what you'd prefer I just thought this was more descriptive

shanaxel42: I think tony and I resolved these in person, but tony yell at me if not

starfish/core/types/_spot_finding_results.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, spot_attributes_list: Optional[List[Tuple]] = None):
+        """
+        Construct a SpotFindingResults instance
+
+        Parameters
+        -----------
+        spot_attributes_list : Optional[List[Tuple]]
+            If spots were found using ImageStack.transform() the result is a list of
+            tuples (indices, SpotAttributes). Instantiating SpotFindingResults with
```
ttung: The format for the indices is not very clear.  It appears to be a Tuple of index values in the order of AXES_ORDER, but I'm just inferring that from the code.

starfish/core/types/_spot_finding_results.py
```diff
@@ -0,0 +1,86 @@
...
+        Return all SpotAttributes across rounds and chs.
+        """
+        return self._results.values()
+
+    @property
+    def round_labels(self):
+        """
+        Return the set of Round labels in the SpotFindingResults
+        """
+        return list(sorted(set(r for (r, ch) in self.keys())))
```
ttung: i still think sorted(set(..)) would do what you want.  the extra list() shouldn't be necessary.

```
>>> sorted(set([1, 2, 3, 4]))
[1, 2, 3, 4]
>>> ^D
```

starfish/core/types/_spot_finding_results.py
```diff
@@ -0,0 +1,86 @@
...
+    def values(self):
+        """
+        Return all SpotAttributes across rounds and chs.
+        """
+        return self._results.values()
+
+    @property
+    def round_labels(self):
+        """
+        Return the set of Round labels in the SpotFindingResults
```
ttung:
```suggestion
        Return the set of round labels in the SpotFindingResults
```

starfish/core/types/_spot_finding_results.py
shanaxel42: That's exactly what it is! I'll make it more clear

docs/source/api/spots/index.rst
```diff
@@ -38,6 +38,18 @@ Spot Detectors can be imported using ``starfish.spots.DetectSpots``, which regis
...
+NOTE: Starfish is embarking on a SpotFinding data structures refactor see `Spot Finding Refactor Plan`_
+DetectSpots will be replaced by the FindSpots module documented below.
+
+.. _`Spot Finding Refactor Plan`: https://github.com/spacetx/starfish/issues/1514
+
+.. code-block:: python
+
+    from starfish.spots import FindSpots
+
+.. autoclass:: starfish.spots.FindSpots
```
ttung: Isn't FindSpots a module?  I think you want `automodule` here to recursively pull in all the implementations in `FindSpots`, and the weird logic to import in all the submodules of `FindSpots`.

docs/source/api/spots/index.rst
```diff
@@ -53,6 +65,15 @@ Decoders can be imported using ``starfish.spots.Decode``, which registers all cl
 .. autoclass:: starfish.spots.Decode
    :members:
 
+NOTE: Starfish is embarking on a SpotFinding data structures refactor see `Spot Finding Refactor Plan`_
+
+.. code-block:: python
+
+    from starfish.spots import DecodeSpots
```
ttung: Ditto.

starfish/core/spots/FindSpots/_base.py
```diff
@@ -0,0 +1,28 @@
...
+from starfish.core.pipeline.algorithmbase import AlgorithmBase
+from starfish.core.types import Number, SpotFindingResults
+
+
+class FindSpotsAlgorithm(metaclass=AlgorithmBase):
+
+    @abstractmethod
+    def run(self, image_stack: ImageStack,
+            reference_image: Optional[ImageStack] = None, *args) -> SpotFindingResults:
+        """Find and measure spots across rounds and channels in the provided
```
ttung: unwrap?  I'd check your editor to make sure you're set up to 100col.  This clocks in at 92col unwrapped for me.

PR #1516, opened 2019-08-29
Add physical Coordinate support to SpotFindingResults 

shanaxel42:
This PR adds physical coordinate information to the SpotResultsClass. This is needed for when the coordinates get transferred during decoding.

Depends on #1489 #1515 

See it in action in #1518

PR #1517, opened 2019-08-29
1.) Provenance logging refactor and support for SpotFindingResults

shanaxel42:
This PR does a bit of a provenance log refactor fixing some issues and also setting things up to work with the new FindSpots and DecodeSpots modules. 

Depends on #1515 #1489 #1516

### DIFF
--- a/starfish/core/pipeline/algorithmbase.py
+++ b/starfish/core/pipeline/algorithmbase.py
@@ -3,10 +3,7 @@ import inspect
 from abc import ABCMeta
 
 from starfish.core.imagestack.imagestack import ImageStack
-from starfish.core.intensity_table.intensity_table import IntensityTable
-from starfish.core.types import LOG
 from starfish.core.types._constants import STARFISH_EXTRAS_KEY
-from starfish.core.util.logging import LogEncoder
 
 
 class AlgorithmBase(ABCMeta):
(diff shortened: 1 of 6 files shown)


starfish/core/util/logging.py
```diff
@@ -1,12 +1,44 @@
...
+from typing import List, Mapping
 
 import pkg_resources
 
 import starfish.core
-from starfish.core.types import CORE_DEPENDENCIES
+from starfish.core.types import CORE_DEPENDENCIES, LOG
+
+
+class Log:
```
ambrosejcarr: Love the idea. Is this only used by spot finding? My intuition is that this should trigger a bigger refactor.

starfish/core/pipeline/algorithmbase.py
```diff
@@ -21,29 +18,33 @@ def run_with_logging(func):
...
+                    result.log.update_log(args[0])
+                if 'FindSpots' in method_class_str:
+                    # Update the log on the resulting SpotFindingResults
+                    result.log.update_log(args[0])
+                if 'DecodeSpots' in method_class_str:
+                    # update log then transfer to DecodedIntensityTable
+                    spot_results = kwargs['spots']
+                    spot_results.log.update_log(args[0])
+                    result.attrs[STARFISH_EXTRAS_KEY] = spot_results.log.encode()
+                # OLD CODE FOR DETECT WILL GET REMOVED
```
ambrosejcarr:
```suggestion
                # TODO shanaxel42 OLD CODE FOR DETECT WILL GET REMOVED
```

starfish/core/util/logging.py
shanaxel42: Nah this is used by everything. This PR was basically the "bigger" logging refactor needed to keep everything working throughout the refactor

starfish/core/pipeline/algorithmbase.py
```diff
@@ -21,29 +18,33 @@ def run_with_logging(func):
...
-                    # update log with spot detection instance args[0]
-                    stack.update_log(args[0])
-                    # get resulting intensity table and set log
-                    it = result
-                    if isinstance(result, tuple):
-                        it = result[0]
-                    it.attrs[STARFISH_EXTRAS_KEY] = LogEncoder().encode({LOG: stack.log})
+            if result is not None:
+                method_class_str = str(args[0].__class__)
+                if 'ApplyTransform' in method_class_str or 'Filter' in method_class_str:
```
ttung: It's plausible that we have mechanisms that take multiple loggable objects.

It might be a good idea to have a base `ProvenanceTracked` class that all objects (`ImageStack`, `IntensityTable`, etc.) inherit, and then to walk through all the arguments to find objects that extend `ProvenanceTracked`.  Then scrape the logs from those objects and build up a provenance graph.

starfish/core/util/logging.py
```diff
@@ -1,12 +1,45 @@
...
+        """
+        self._log: List[dict] = list()
+
+    def update_log(self, class_instance) -> None:
+        """
+        Adds a new entry to the log list.
+
+        Parameters
+        ----------
+        class_instance: The instance of a class being applied to the imagestack
```
ttung: Not just applicable to `ImageStack` now, right?

starfish/core/pipeline/algorithmbase.py
shanaxel42: this seems like a larger refactor, like maybe something post spot finding stuff

ttung: Sure, can you tack that onto the logging epic?

There are a lot of gaps in the logging -- for instance, ApplyTransform plausibly should log the origin of the warping data.  Also can be covered afterwards.

Issue #1446, opened 2019-07-11
Harmonize apply and reduce


ambrosejcarr:
`split/apply/reduce/combine` are common, related frameworks that users expect to work harmoniously with standard APIs 

Starfish currently has: 

1. `starfish.Filter.reduce` which takes a string reference to a function and module and reduces the dimensionality of an input `ImageStack`. 

2. `starfish.ImageStack.apply` which takes a callable and applies it to chunks of the `ImageStack`, without changing the dimensionality. 

@dganguli and I suspect it would be more intuitive either for `reduce to be a classmethod of the `ImageStack` or for `Apply` to become a filter. This need not dictate that this be the sole code path for apply, as we recognize that multiple other pipeline components leverage it, but it could be that it becomes a private method.

ttung:
`apply` should be a filter (in addition to being an instance method).  reduce as a classmethod of ImageStack makes no sense.

ambrosejcarr:
> apply should be a filter (in addition to being an instance method). reduce as a classmethod of ImageStack makes no sense.

Cool with me, my requirement here is just consistency between the two.

PR #1519, opened 2019-08-29
funcs passed to apply and transform can use positional arguments

ttung:
With the existing code, funcs passed to apply and transform can only use keyword arguments.  This is unnecessarily limiting, and if we want to support #1446, we need to be able to work with a variety of functions with their own argument calling mechanisms.

### DIFF
--- a/starfish/core/imagestack/imagestack.py
+++ b/starfish/core/imagestack/imagestack.py
@@ -705,6 +705,7 @@ class ImageStack:
     def apply(
             self,
             func: Callable,
+            *args,
             group_by: Set[Axes]=None,
             in_place=False,
             verbose: bool=False,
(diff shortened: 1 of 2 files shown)


starfish/core/imagestack/imagestack.py
```diff
@@ -772,20 +773,23 @@ def apply(
...
+                *args,
                 group_by=group_by, in_place=True, verbose=verbose, n_processes=n_processes,
                 clip_method=clip_method,
                 **kwargs
             )
             return image_stack
 
         # wrapper adds a target `data` parameter where the results from func will be stored
-        # data are clipped or scaled by chunk using preserve_float_range if clip_method != 2
+        # data are clipped or scaled by chunk using preserve_float_range if
```
shanaxel42: *data is*...jkjk ;)

ttung: Lol.

To be fair, this entire sentence doesn't make a ton of sense.

ttung: I put up https://github.com/spacetx/starfish/pull/1551 which I'm going to have @ambrosejcarr approve since he wrote these comments.

PR #1520, opened 2019-08-29
map filter

### LINKED ISSUE #1446: Harmonize apply and reduce

ttung:
This allows a variety of mapping functions to be added to a starfish pipeline without writing a new Filter.

Test plan: Add a simple test.
Fixes #1446
Depends on #1519

### DIFF
--- a/starfish/core/image/Filter/map.py
+++ b/starfish/core/image/Filter/map.py
@@ -1,0 +1,178 @@
+import importlib
+from enum import Enum
+from typing import (
+    Callable,
+    cast,
+    Mapping,
+    Optional,
+    Set,
+    Union
+)
+
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.types import Axes, Clip
+from ._base import FilterAlgorithmBase
+
+
+class Map(FilterAlgorithmBase):
+    """
+    Map from input to output by applying a specified function to the input.  The output must have
+    shape as the input.
+
+    Parameters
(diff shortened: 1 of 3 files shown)


starfish/core/image/Filter/map.py
```diff
@@ -0,0 +1,178 @@
...
+
+from starfish.core.imagestack.imagestack import ImageStack
+from starfish.core.types import Axes, Clip
+from ._base import FilterAlgorithmBase
+
+
+class Map(FilterAlgorithmBase):
+    """
+    Map from input to output by applying a specified function to the input.  The output must have
+    shape as the input.
```
shanaxel42: must have the same shape?

ttung: thx fixed.

Issue #1321, opened 2019-05-07
Improvements to image segmentation


shanaxel42:
# Objective
*As a computational biologist analyzing imaging data containing cells, I want starfish to implement the tools that I use in my own workflow to segment cells so that starfish can provide end-to-end processing of my data and I can make use of starfish's output formats and visualization capabilities. Starfish must implement these in a way that I can understand and use without assistance from developers*

## Acceptance Criteria
* Refactor watershed segmentation to be more intuitive, watershed supports 3d images.
* Starfish can build a representation of cellular regions from FIJI ROIsets (enables manual segmentation) (#1370)
* Starfish can apply a pre-trained Ilastik model to segment images in SpaceTx Format
* Starfish implements an object model whose contracts fit these features.
* SegmentationMaskCollection can be queried will a cell_id to return a specified mask (#1487)
* Starfish has adequate user-facing documentation that our users can leverage segmentation without developer assistance
* Segmentation methods respond in interpretable ways when no cells are identified (#468)

## Validation
* Test that users (including WG5) can successfully use starfish methods. 

## Out of scope
* Reconciling segmentation artifacts at image boundaries (future work)

## Additional Notes

*Note: you must have the [zenhub chrome extension](https://chrome.google.com/webstore/detail/zenhub-for-github/ogcgkffhplmphkaahpmffcafajaocjbd?hl=en-US) installed to see the issues linked to this Epic.*

ambrosejcarr:
Watershed work will enable 3d support per #53

pdichiaro:
it is challenging to segment different cell types in the same image. I am wondering if it is possible to create a segmentation mask using different segmentation methods. For example an user wants to use watershed segmentation with a membrane staining to create automatically a mask. Then he wants to refine it manually in those aereas where the cell segnentation was not so accurate.

ambrosejcarr:
Hi @pdichiaro, This is a *great* question. 

On the question of automated mask generation, we want to provide examples of how to improve the inputs to watershed that might help with different cell sizes. @dganguli was telling me about approaches where watershed basins were built based on the kernel density of the decoded points, which should be sensitive to cell size, and could reduce the need to refine the segmentation. 

If interactive refinement is still necessary, it might be smoother to read the masks into [napari](https://github.com/napari/napari), and extend `starfish.display` to enable mask editing. @sofroniewn , am I correct that napari already enables editing of the shape layer?

pdichiaro:
[...]

PR #1528, opened 2019-09-03
[WIP] Add Segmentation Vignettes

ambrosejcarr:
## Objectives: 
- Understand the current limitations of starfish's segmentation module
- Frame feature additions which would improve functionality and ease of use for segmentation

I've put this up for @dganguli 's visibility since this might be useful for the paper. 

## Current Status
Create `SegmentationMaskCollection` objects from the following examples:
- Example of training and then applying a pre-trained Ilastik module and reading the resulting data
- Example of creating and loading ROIsets from FIJI, based on starfish data
- Example of creating and loading Masks from FIJI, based on starfish data
- Example of loading geoJSON from @berl

## Next
- Update #1321 based on learnings here. 
- Seek feedback on scoped work in #1321


dganguli:
Can you put the files:
SongLinROIS_deduplicated.json
mFISH__001_registered-downsampled.png 

on S3? Worried that slack isn't forever.

dganguli:
For manual segmentation, using FIJI Segmentation manager, the segmentation results look "off". As though it's missing data at the boundaries -- can you verify that this is a segmentation error and not an error in SegmentationMaskCollection?

dganguli:
ROI manager may have same aforementioned boundary problem?

dganguli:
should verify starfish.display works on images an order of magnitude bigger than we're used to -- relevant to geo_json example in this notebook

dganguli:
@ambrosejcarr manual notebook demonstrates 3 different ways of turning manual segmentations into SegmentationMaskCollections. Ilastik notebook shows 1 way. All of these look good/useful -- can we write an API doc for exactly what we want based on these 4 methods and dump it here for review.

PR #1540, opened 2019-09-06
Unify FunctionSource in Map and Reduce

ttung:
This is needed to enable #1539, as that imports `FunctionSource` from `ImageStack` (and `Map`/`Reduce` import `ImageStack`).

Note that this is not entirely for free.  The function aliases (example: where we alias 'max' to 'amax' for `FunctionSource.np`) now come from one source, which may or may not be desirable.

Test plan: travis yo


ttung:
> general question, and sorry if this has already been discussed. But what's the advantage of having the function source logic over just have a function parameter passed into reduce? Like func=np.max

as discussed offline, allowing arbitrary callables to be passed in makes provenance tracking extremely challenging.  However, we could plausibly use code inspection to solve that problem.

ambrosejcarr:
> as discussed offline, allowing arbitrary callables to be passed in makes provenance tracking extremely challenging. However, we could plausibly use code inspection to solve that problem.

If you see a path where this is achievable, I believe usability benefits strongly from just passing in the function.

ttung:
> If you see a path where this is achievable, I believe usability benefits strongly from just passing in the function.

It's _somewhat_ achievable.  I'll look into it.

ambrosejcarr:
> It's somewhat achievable. I'll look into it.

After PR is merged is OK with me, if this blocks.

ttung:
> After PR is merged is OK with me, if this blocks.

#1568

PR #1541, opened 2019-09-09
Get rid of shared state for LocalMaxPeakFinder

ttung:
Shared state is fundamentally incompatible with multithreading.

1. Each tile computes its own threshold.
2. Don't share the spot coords.

Part of #834

### DIFF
--- a/starfish/core/spots/DetectSpots/local_max_peak_finder.py
+++ b/starfish/core/spots/DetectSpots/local_max_peak_finder.py
@@ -70,13 +70,6 @@ class LocalMaxPeakFinder(DetectSpotsAlgorithm):
         self.is_volume = is_volume
         self.verbose = verbose
 
-        # these parameters are useful for debugging spot-calls
-        self._thresholds: Optional[np.ndarray] = None
-        self._spot_counts: Optional[List[int]] = None
-        self._grad = None
-        self._spot_props = None
-        self._labels = None
-
     def _compute_num_spots_per_threshold(self, img: np.ndarray) -> Tuple[np.ndarray, List[int]]:
         """Computes the number of detected spots for each threshold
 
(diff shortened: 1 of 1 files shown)


starfish/core/spots/DetectSpots/local_max_peak_finder.py
```diff
@@ -205,8 +193,10 @@ def _compute_threshold(self, img: Union[np.ndarray, xr.DataArray]) -> float:
         """
         img = np.asarray(img)
         thresholds, spot_counts = self._compute_num_spots_per_threshold(img)
-        threshold = self._select_optimal_threshold(thresholds, spot_counts)
-        return threshold
+        if len(spot_counts) == 0:
```
dganguli: nice catch!

starfish/core/spots/DetectSpots/local_max_peak_finder.py
```diff
@@ -70,13 +70,6 @@ def __init__(
         self.is_volume = is_volume
         self.verbose = verbose
 
-        # these parameters are useful for debugging spot-calls
```
dganguli: so. these parameters are indeed useful for debugging spot-calls. is there any way to save them, like for example if you're not using multiprocessing? or is state just impossible to have?

dganguli: debugging is the wrong word. these parameters, per tile, are critical to doing a quality control analysis of the spot calls.

dganguli: but on the other hand, this piece of the code should probably be pulled out of this spot detector and made general for use in all of the other spot detectors. @ambrosejcarr @shanaxel42 and i have talked about this. as such, i suggest we punt this issue over to the spot-finding epic.

ttung:
> Does this cause a significant slowdown?

I haven't tested that.

PR #1545, opened 2019-09-11
Don't waste memory/compute in preserve_float_range

ttung:
There are two flaws with the existing `preserve_float_range` implementation:
1. It wastes memory by copying even when it does not have to.
2. It does extra loops through the array.

This attempts to optimize `preserve_float_range`, while retaining flexibility for the caller to configure the behavior we want.  There is a new parameter, `preserve_input`, which controls whether or not it is important to preserve the input.  When it is true, we will make deliberate effort to preserve the input data, even at the cost of performance.  When it is false, we will overwrite the input data unless we are forced to make a copy (the two reasons that compel us to do so are if the dtype is different and if the input array is immutable).

Test plan: Add new tests to cover all the scenarios, and `make -j fast`

### DIFF
--- a/starfish/core/util/dtype.py
+++ b/starfish/core/util/dtype.py
@@ -3,20 +3,32 @@ from typing import Union
 import numpy as np
 import xarray as xr
 
+
 def preserve_float_range(
         array: Union[xr.DataArray, np.ndarray],
-        rescale: bool=False) -> Union[xr.DataArray, np.ndarray]:
+        rescale: bool = False,
+        preserve_input: bool = False,
+) -> Union[xr.DataArray, np.ndarray]:
     """
     Clips values below zero to zero. If values above one are detected, clips them
     to 1 unless `rescale` is True, in which case the input is scaled by
     the max value and the dynamic range is preserved.
 
+    Input arrays may be modified by this operation if `preserve_input` is False.  There is no
+    guarantee that the input array is returned even if `preserve_input` is True, however.
+
     Parameters
     ----------
     array : Union[xr.DataArray, np.ndarray]
         Array whose values should be in the interval [0, 1] but may not be.
-    rescale: bool
+    rescale : bool
         If true, scale values by the max.
+    preserve_input : bool
+        If True, ensure that we do not modify the input data.  This may either be done by making a
+        copy of the input data or ensuring that the operation does not modify the input array.
+
+        Even if `preserve_input` is True, modifications to the resulting array may modify the input
+        array.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


starfish/core/util/dtype.py
```diff
@@ -3,40 +3,95 @@
...
-    if np.any(negative):
-        data[negative] = 0
-    if np.any(data > 1):
-        if rescale:
-            data /= data.max()
-        else:
-            data[data > 1] = 1
-    return array.astype(np.float32)
+    casted = data.astype(np.float32, copy=False)
+    if casted is not data:
```
ambrosejcarr: I see this logic, but I think it's inconsistent with the documentation -- if we're casting to float32, we never preserve input data, but users might think otherwise based on the above parameters.

starfish/core/util/dtype.py
```diff
@@ -3,40 +3,95 @@
...
+        preserve_input = False
+    data = casted
+
+    if preserve_input or not data.flags['WRITEABLE']:
```
ambrosejcarr: `if not data.flags['WRITEABLE']` I think this should emit a warning, as the function will do something different than the user expects.

starfish/core/util/dtype.py
```diff
@@ -3,40 +3,95 @@
...
+        # if we still want a copy, check to see if any modifications would be made.  if so, make a
+        # copy.
+        belowzero = np.any(data < 0)
+        aboveone = np.any(data > 1)
+        if belowzero or aboveone:
+            data = data.copy()
+            _preserve_float_range_in_place(data, rescale and aboveone)
+    else:
```
ambrosejcarr: why no check for belowzero here?

starfish/core/util/test/test_dtype.py
```diff
@@ -0,0 +1,115 @@
+import numpy as np
+import pytest
+import xarray as xr
+
+from ..dtype import preserve_float_range
+
+
+def combos(make_xarrays: bool):
```
ambrosejcarr: What if someone passed an integer dtype? or float64 that's also below zero? Might be worth a few more test combinations.

starfish/core/util/dtype.py
ttung: If we do a cast, `astype` will make a copy, in which case we do preserve the input data.

starfish/core/util/dtype.py
ttung: How is that?

starfish/core/util/dtype.py
ttung: `_preserve_float_range_in_place` always clips to below-zero values to zero.

starfish/core/util/test/test_dtype.py
ttung: sure, i can add some examples of that.

starfish/core/util/dtype.py
ambrosejcarr: What would happen if they said `preserve_input=False`, but `data.flags['WRITEABLE']` was also False?

ttung: We'd make a copy.  `preserve_input==False` does not mean we're not allowed to make a copy.

ambrosejcarr: Got it. I understand.

starfish/core/util/dtype.py
ambrosejcarr: `copy=False` tripped me up. Got it, though.

PR #1561, opened 2019-09-19
[easy] Remove mp_dataarray

ttung:
Some of this is causing errors with the newest xarray version.  Since we don't need it any more, let's just blow it away.


ttung:
I'm landing this.  Feel free to post-land-review.

PR #1562, opened 2019-09-19
Fix mypy + xarray errors

ttung:
xarray 0.13.0 now has type annotations, which breaks a bunch of things.  This attempts to fix the issues.

It appears that xr.DataArray no longer acts like a Sequence[Number], so a bunch of things need to be casted.  It's probable that we should create a new type annotation for array-like objects (i.e., Union[Sequence, xr.DataArray, np.ndarray]).

Depends on #1561 

Test plan: travis.  https://travis-ci.com/spacetx/starfish/builds/128278048 has the errors from upgrading xarray to 0.13.0.

### DIFF
--- a/REQUIREMENTS-STRICT.txt
+++ b/REQUIREMENTS-STRICT.txt
@@ -3,8 +3,8 @@ appnope==0.1.0
 attrs==19.1.0
 backcall==0.1.0
 bleach==3.1.0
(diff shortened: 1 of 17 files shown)


ttung:
> Could you either make an issue or change this here? After that looks fine to me.

#1564

PR #1565, opened 2019-09-20
[easy] Fix superclass type annotations

ttung:
AssignTargetsAlgorithm should be operating with DecodedIntensityTable.


ttung:
Landing this because it blocks release.

PR #1567, opened 2019-09-20
Relax the constraint that AssignTargets operates on DecodedIntensityTables

ttung:
Upon review, I was probably too hasty in landing #1565.  :(

It doesn't appear that assigning targets has any requirement that the spots are decoded, so I'm putting this up for review.


ttung:
@shanaxel42 says no.

PR #1591, opened 2019-09-30
[WIP] updates for napari 0.2.0 release

sofroniewn:
This PR will update `starfish.display` for the `0.2.0` release of napari which is planned imminently but has not happened yet. I still need to test it out more thoroughly, but wanted to make the changes that I know we'll need to make

### DIFF
--- a/starfish/core/_display.py
+++ b/starfish/core/_display.py
@@ -17,7 +17,7 @@ except ImportError:
     Viewer = None
 
 
-NAPARI_VERSION = "0.1.5"  # when changing this, update docs in display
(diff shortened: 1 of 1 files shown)


sofroniewn:
I'm confused why the tests are passing - after #1570 they should break as I've added code that should break the version of napari currently on pypi `0.1.5`

ttung:
Napari tests are only run on master, or branches with the suffix '-alltest'

sofroniewn:
Closing in favour of #1599 which has the correct branch name and the only changes that are actually needed

PR #1602, opened 2019-10-04
Fix bugs in per-round-max-decoder

### LINKED ISSUE #1485: decoding method in seqFISH
after generation of intensityTable by LocalSearchBlobDetecor I use codebook.decode_per_round_max(intensities) to decode each spot. It does not work(error: all-Nan slice encountered).
Screenshot 2019-08-06 at 19 43 20

ttung:
1. When the entire row is nan, the decoder chokes.  This is remedied by decoding on an array where the nan values are replaced with 0s.
2. When the entire row is of equal intensity, the `np.argmax` arbitrarily picks the first column as the winner.  That erroneously decodes as ch=0 having the max intensity.  This code detects that scenario, and rewrites the ch to an impossible value in that situation.

Test plan: Wrote tests that failed with the existing code, applied fixes and verified that they now work.
Depends on #1600
Fixes #1485


starfish/core/codebook/codebook.py
```diff
@@ -651,7 +651,15 @@ def _view_row_as_element(array: np.ndarray) -> np.ndarray:
                 distances=(Features.AXIS, np.empty(0, dtype=np.float64)),
                 passes_threshold=(Features.AXIS, np.empty(0, dtype=bool)))
 
-        max_channels = intensities.argmax(Axes.CH.value)
+        intensities_without_nans = intensities.fillna(0)
+        max_channels = intensities_without_nans.argmax(Axes.CH.value)
+        # this snippet of code finds all the (feature, round) spots that have uniform illumination,
+        # and assigns them to a ch number that's one larger than max possible.
```
ambrosejcarr:
```suggestion
        # and assigns them to a ch number that's one larger than max possible to ensure that.
        # such spots decode to `nan`.
```

PR #1608, opened 2019-10-09
Fix broken build

ttung:
In #1540, I missed a conversion from Reduce.FunctionSource to types.FunctionSource.


ttung:
I'm going to land this to unbreak master.  #shame

Issue #908, opened 2018-12-20
Formalize the ability for spot detection to return extra data


ttung:
This should be done in the form of a class for returning data from spot detection.  Normally, that encompasses the IntensityTable, but it can hold additional payload.

ambrosejcarr:
Will address #708

ambrosejcarr:
I think this may need to be expanded to:

 "formalize the ability for _Pipeline Components_ to return extra data". 

The pixel decoder will need to emit a second imagestack that, instead of labels, contains the confidence in those labels.

ttung:
> I think this may need to be expanded to:

> "formalize the ability for Pipeline Components to return extra data".

> The pixel decoder will need to emit, in addition to the probability maps for each gene, a label image that contains the ML estimate for the gene corresponding to each pixel.

If pixel decoding moves to image space (i.e., a filter), we should rethink what exactly the remaining spot detection algorithms truly need.

ambrosejcarr:
It could be good to determine if this is still needed while you're deep into spot finding @shanaxel42 

It might be that the elimination of the CLI mitigates this problem.

PR #1615, opened 2019-10-17
10.) Formalizing the ability to return extra info from spot finding 

shanaxel42:
Changes the return value for every image_to_spots method to PerImageSpotResults, a named tuple that contains the SpotAttributes found in the image as well as any extra information for QC and debugging. The only spot finder that returns extra data at this point is PeakLocalMax which now returns threshold information. 

fixes: #908, #708

### DIFF
--- a/starfish/core/spots/DecodeSpots/trace_builders.py
+++ b/starfish/core/spots/DecodeSpots/trace_builders.py
@@ -3,8 +3,13 @@ from typing import Callable, Mapping
 import pandas as pd
 
 from starfish.core.intensity_table.intensity_table import IntensityTable
-from starfish.core.types import Axes, Features, SpotAttributes, SpotFindingResults, \
+from starfish.core.types import (
+    Axes,
+    Features,
+    SpotAttributes,
+    SpotFindingResults,
     TraceBuildingStrategies
+)
 from .util import _build_intensity_table, _match_spots, _merge_spots_by_round
 
 
@@ -19,14 +24,14 @@ def build_spot_traces_exact_match(spot_results: SpotFindingResults, **kwargs) ->
         Spots found across rounds/channels of an ImageStack
     """
     # create IntensityTable with same x/y/z info accross all r/ch
-    spot_attributes = list(spot_results.values())[0]
+    spot_attributes = list(spot_results.values())[0].spot_attrs
     intensity_table = IntensityTable.zeros(
         spot_attributes=spot_attributes,
         round_labels=spot_results.round_labels,
         ch_labels=spot_results.ch_labels,
     )
     for r, c in spot_results.keys():
-        value = spot_results[{Axes.ROUND: r, Axes.CH: c}].data[Features.INTENSITY]
+        value = spot_results[{Axes.ROUND: r, Axes.CH: c}].spot_attrs.data[Features.INTENSITY]
         # if no exact match set value to 0
         value = 0 if value.empty else value
         intensity_table.loc[dict(c=c, r=r)] = value
@@ -50,7 +55,7 @@ def build_traces_sequential(spot_results: SpotFindingResults, **kwargs) -> Inten
 
     """
 
-    all_spots = pd.concat([sa.data for sa in spot_results.values()], sort=True)
+    all_spots = pd.concat([sa.spot_attrs.data for sa in spot_results.values()], sort=True)
 
     intensity_table = IntensityTable.zeros(
         spot_attributes=SpotAttributes(all_spots),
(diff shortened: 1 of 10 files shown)


starfish/core/spots/DecodeSpots/trace_builders.py
```diff
@@ -68,7 +73,7 @@ def build_traces_sequential(spot_results: SpotFindingResults, **kwargs) -> Inten
 
     i = 0
     for (r, c), attrs in spot_results.items():
```
ttung: would not call this attrs in this context because attrs and spot_attrs gets confusing.

starfish/core/spots/FindSpots/blob.py
```diff
@@ -84,7 +85,7 @@ def __init__(
         except ValueError:
             raise ValueError("Detector method must be one of {blob_log, blob_dog, blob_doh}")
 
-    def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> SpotAttributes:
+    def image_to_spots(self, data_image: Union[np.ndarray, xr.DataArray]) -> PerImageSpotResults:
```
ttung: PerTileSpotResults?

starfish/core/types/_constants.py
```diff
@@ -40,6 +41,13 @@ class Coordinates(AugmentedEnum):
 """
 
 
+PerImageSpotResults = namedtuple('PerImageSpotResults', 'spot_attrs, extras')
```
ttung: Would make this a dataclass.

starfish/core/types/_constants.py
```diff
@@ -40,6 +41,13 @@ class Coordinates(AugmentedEnum):
...
+"""
+Named tuple that gets returned in every spot finding"s image_to_spots method, spot_attrs are the
```
ttung:
```suggestion
Named tuple that gets returned in every spot finding's image_to_spots method, spot_attrs are the
```

starfish/core/types/_constants.py
ttung: wait is this right?  this is just two strings.

starfish/core/spots/FindSpots/h_max.py
```diff
@@ -11,7 +11,8 @@
 
 from starfish.core.imagestack.imagestack import ImageStack
 from starfish.core.spots.FindSpots import spot_finding_utils
-from starfish.core.types import Axes, Features, SpotAttributes, SpotFindingResults
+from starfish.core.types import Axes, Features, PerImageSpotResults, SpotAttributes, \
```
ttung: i would wrap this like how you did it for `starfish/core/spots/DecodeSpots/trace_builders.py`

starfish/core/spots/FindSpots/blob.py
shanaxel42: Would we call it tile though since we don't include a z value? I went back a forth about this but all your TileKey TileFetcher code includes a r/ch/z for each tile so I wasn't sure it was the same

starfish/core/types/_constants.py
shanaxel42: The strings are just the names of the values put into the named tuple. So you can access them like PerImageSpotResults.spot_attrs

starfish/core/spots/FindSpots/blob.py
ttung: PerImageSliceSpotResults?

fwiw, I thought we should call n-dimensional tensors images (plural) rather than singular, which I think would have avoided this wonkiness.  You should consult with the proponents of calling the entire n-dimensional tensor "image". :)

If they say nothing, then I would go with your judgement.

starfish/core/types/_constants.py
ttung: Shouldn't it be:

```suggestion
PerImageSpotResults = namedtuple('PerImageSpotResults', 'spot_attrs', 'extras')
```

starfish/core/spots/FindSpots/local_max_peak_finder.py
```diff
@@ -13,7 +13,8 @@
 from starfish.core.config import StarfishConfig
 from starfish.core.imagestack.imagestack import ImageStack
 from starfish.core.spots.FindSpots import spot_finding_utils
-from starfish.core.types import Axes, Features, Number, SpotAttributes, SpotFindingResults
+from starfish.core.types import Axes, Features, Number, PerImageSliceSpotResults, SpotAttributes, \
```
ttung: this can be wrapped without `\`

starfish/core/spots/FindSpots/local_max_peak_finder.py
```diff
@@ -173,7 +174,8 @@ def _select_optimal_threshold(self, thresholds: np.ndarray, spot_counts: List[in
 
         return selected_thr
 
-    def _compute_threshold(self, img: Union[np.ndarray, xr.DataArray]) -> float:
+    def _compute_threshold(self, img: Union[np.ndarray, xr.DataArray]
+                           ) -> Tuple[float, Optional[np.ndarray], Optional[List[int]]]:
```
ttung: can you update the docblock for this?

PR #1622, opened 2019-10-18
Clean up binary mask

ttung:
1. Use typed axes/coordinates when possible.
2. Have `_get_axes_names` return axes/coordinates in the order they are expected in the numpy arrays.
3. Improve how we render the mask names.

Test plan: `pytest starfish/core/binary_mask/test`

### DIFF
--- a/starfish/core/binary_mask/binary_mask.py
+++ b/starfish/core/binary_mask/binary_mask.py
@@ -12,6 +12,7 @@ from typing import (
     MutableSequence,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Union,
 )
@@ -21,9 +22,9 @@ import xarray as xr
 from skimage.measure import regionprops
 from skimage.measure._regionprops import _RegionProperties
 
-from starfish.core.types import Axes, Coordinates
+from starfish.core.types import Axes, Coordinates, Number
 from .expand import fill_from_mask
-from .util import _get_axes_names, AXES_ORDER
+from .util import _get_axes_names
 
 
 def _validate_binary_mask(arr: xr.DataArray):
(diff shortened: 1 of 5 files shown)


starfish/core/binary_mask/binary_mask.py
```diff
@@ -157,14 +159,15 @@ def from_label_image(
...
-            Physical coordinates for each axis.
+            Physical ticks for each axis.
 
         Returns
         -------
         masks : BinaryMaskCollection
             Masks generated from the label image.
         """
         props = regionprops(label_image)
+        len_max_label = len(str(len(props) - 1))
```
shanaxel42: why is there the extra str conversion here?

starfish/core/binary_mask/expand.py
```diff
@@ -4,14 +4,14 @@
...
-from .util import _get_axes_names, AXES_ORDER
+from .util import _get_axes_names
 
 
 def fill_from_mask(
         mask: xr.DataArray,
         fill_value: int,
         result_array: np.ndarray,
-        axes_order: Sequence[Union[str, Axes]] = AXES_ORDER,
+        axes_order: Sequence[Union[str, Axes]] = (Axes.ZPLANE, Axes.Y, Axes.X),
```
shanaxel42: why not use your get_axes_order helper method here?

starfish/core/binary_mask/binary_mask.py
ttung: I was tripped up by this too.  Let's say we have 100 regions.  len(props) - 1 gets you 99 as an int.  Then str(99) gives you "99".  Then len("99") gives you 2.

starfish/core/binary_mask/expand.py
ttung: That basically happens in #1628

PR #1626, opened 2019-10-24
Composite codebook support

shanaxel42:
- Added a slicing method in codebook that allows a user to easily get a portion of the data.
- Added a new SimpleLookupDecoder for non multiplexed data that just makes a table of r/ch keys to target value and decodes. 
- Added a few tests to demonstrate workflows with composite experiment data

### DIFF
--- a/starfish/core/codebook/codebook.py
+++ b/starfish/core/codebook/codebook.py
@@ -1,7 +1,7 @@
 import json
 import uuid
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, TypeVar, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Set, Tuple, TypeVar, Union
 
 import numpy as np
 import pandas as pd
@@ -17,11 +17,11 @@ from starfish.core.codebook._format import (
     MIN_SUPPORTED_VERSION,
 )
 from starfish.core.config import StarfishConfig
+from starfish.core.imagestack import indexing_utils
 from starfish.core.intensity_table.decoded_intensity_table import DecodedIntensityTable
 from starfish.core.intensity_table.intensity_table import IntensityTable
 from starfish.core.spacetx_format.util import CodebookValidator
 from starfish.core.types import Axes, Features, Number
-
 
 NormalizedFeaturesArgtype = TypeVar("NormalizedFeaturesArgtype", "Codebook", IntensityTable)
 
@@ -388,6 +388,20 @@ class Codebook(xr.DataArray):
         cls._verify_version(version_str)
 
         return cls.from_code_array(codebook_doc[DocumentKeys.MAPPINGS_KEY], n_round, n_channel)
+
+    def get_partial(self, indexers: Mapping[Axes, Union[int, slice, Sequence]]):
+        """
+        Slice the codebook data according to the provided indexing parameters. Used in a composite
+        codebook scenario.
+
+        Parameters
+        ----------
+        indexers : Mapping[Axes, Union[int, Sequence]]
+            A dictionary of dim:index where index is the value, values or range to index the
+            dimension
+        """
+        selector = indexing_utils.convert_to_selector(indexers)
+        return indexing_utils.index_keep_dimensions(self, indexers=selector)
 
     def to_json(self, filename: Union[str, Path]) -> None:
         """
--- a/starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
+++ b/starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
@@ -1,0 +1,53 @@
+from typing import Dict, Tuple
+
+from starfish.core.codebook.codebook import Codebook
+from starfish.core.intensity_table.decoded_intensity_table import DecodedIntensityTable
+from starfish.core.spots.DecodeSpots.trace_builders import build_traces_sequential
+from starfish.core.types import Axes, Features, SpotFindingResults
+from ._base import DecodeSpotsAlgorithm
+
+
+class SimpleLookupDecoder(DecodeSpotsAlgorithm):
+    """
+    Decode spots by assiging the target value of a spot to the corresponding target value of the
+    round/ch it was found in. This method only makes sense to use in non mulitplexed sequential
+    assays where each r/ch pair only has one target assigned to it.
+
+    Parameters
+    ----------
+    codebook : Codebook
+        Contains codes to decode IntensityTable
+
+    """
+
+    def __init__(self, codebook: Codebook):
+        self.codebook = codebook
+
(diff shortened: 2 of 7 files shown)


starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
```diff
@@ -0,0 +1,53 @@
...
+        DecodedIntensityTable :
+            IntensityTable decoded and appended with Features.TARGET and values.
+
+        """
+        lookup_table: Dict[Tuple, str] = {}
+        for target in self.codebook[Features.TARGET]:
+            for ch_label in self.codebook[Axes.CH.value]:
+                for round_label in self.codebook[Axes.ROUND.value]:
+                    if self.codebook.loc[target, round_label, ch_label]:
+                        lookup_table[(int(round_label), int(ch_label))] = str(target.values)
```
ttung: this seems slow, but the number of targets is probably not horrendous.

it wouldn't surprise me, however, if there were a faster way to do this.

Also, if you *are* going to do it the slow way, you might as well detect scenarios where multiple bit patterns decode to the same target.

starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
```diff
@@ -0,0 +1,53 @@
...
+
+        for r_ch_index, spot_attrs in spots.items():
+            target = lookup_table[r_ch_index] if r_ch_index in lookup_table else 'nan'
+            spot_attrs.data[Features.TARGET] = target
```
ttung: i suspect this will be unacceptable from the performance point of view.  iterating across all spots in python space is very expensive.

starfish/core/spots/DecodeSpots/test/test_composite_codebook.py
```diff
@@ -0,0 +1,238 @@
...
+        max_sigma=4,
+        num_sigma=5,
+        threshold=0,
+        measurement_type='max')
+
+
+def composite_codebook() -> Tuple[Codebook, ImageStack]:
+    """
+    Produce an Imagestack representing a composite experiment where the first 2 rounds are
+    multiplexed data and the last round is sequential data.
```
ttung: do you have a test where we exercise something like what @berl has (multiplexed data and sequential data mixed into one round)?

starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
shanaxel42: This doesn't actually iterate across all spots. It iterates across spot attribute tables in SpotFindingResults. So it only iterates across r/ch pairs

shanaxel42: it just assigns a target value to the entire spot attributes table for that r/ch pair

ttung: Ehh, isn't that across all spots?

starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
shanaxel42: This is the same code we use in Codebook.to_json for saving out the codebook. I'm not sure I understand this "you might as well detect scenarios where multiple bit patterns decode to the same target." how would that make sense for a non multiplexed assay?

ttung: If one passes in a multiplexed codebook into this decoder, it will happily decode it in probably strange and unexpected ways.

starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
ttung: @shanaxel42 explained this to me over slack.  

```
it's iterating across entries in SpotFindingResults which is the Mapping[(r/ch), SpotAttributes] data structure
so for every (r/ch) pair it looks up the associated target value then assigns that to the entire table
```

So she's right and I'm very wrong.

starfish/core/spots/DecodeSpots/simple_lookup_decoder.py
shanaxel42: oh yea but the description says it should only be used with non multiplexed data similarly to how the decode_per_round description says it should only be used for one hot encoded data

starfish/core/codebook/codebook.py
```diff
@@ -388,6 +388,19 @@ def open_json(
...
 
+    def get_partial(self, indexers: Mapping[Axes, Union[int, tuple]]):
+        """
+        Slice the codebook data according to the provided indexing parameters. Used in a composite
+        codebook scenario.
+
+        Parameters
+        ----------
+        indexers : Mapping[Axes, Union[int, tuple]]
+            A dictionary of dim:index where index is the value or range to index the dimension
```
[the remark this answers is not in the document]
shanaxel42: Changed to support lists of arbitrary values as well

starfish/core/spots/DecodeSpots/test/test_composite_codebook.py
shanaxel42: added!

starfish/core/spots/DecodeSpots/test/test_composite_codebook.py
```diff
@@ -0,0 +1,340 @@
...
+    per_round_decoder = PerRoundMaxChannel(codebook=codebook_multiplexed)
+    decoded_multiplexed = per_round_decoder.run(spots=multiplexed_spots)
+    assert np.array_equal(decoded_multiplexed[Features.TARGET].values, ['GENE_B', 'GENE_A'])
+
+    simple_lookup_decoder = SimpleLookupDecoder(codebook=codebook_sequential)
+    decoded_sequential = simple_lookup_decoder.run(spots=sequential_spots)
+    assert np.array_equal(decoded_sequential[Features.TARGET].values, ['GENE_C', 'GENE_D'])
+
+
+def test_composite_codebook_decoding_seperate_sized_stacks():
```
ttung: It would be good if this had a short description like `test_composite_codebook_decoding`.

starfish/core/spots/DecodeSpots/test/test_composite_codebook.py
```diff
@@ -0,0 +1,340 @@
...
+    assert np.array_equal(decoded_multiplexed[Features.TARGET].values, ['GENE_B', 'GENE_A'])
+
+    lookup_decoder = SimpleLookupDecoder(codebook=codebook)
+    decoded_sequential = lookup_decoder.run(spots=sequential_spots)
+
+    assert np.array_equal(sorted(decoded_sequential[Features.TARGET].values),
+                          ['GENE_C', 'GENE_D', 'GENE_E'])
+
+
+def test_composite_codebook_mixed_round():
```
ttung: ditto.

PR #1629, opened 2019-10-28
Switch to version exclusion for scikit-image workaround

ttung:
It's easier to exclude these versions forever and ever than it is to carry around this code workaround, I think.  Please reject if you don't agree. :) 

Also made the DataTypeWarning tests less brittle.  We only care that we raised our warning.  We don't care if others raised warnings.

Test plan: travis


ttung:
> Fine with me!

I'll take that as an approval. :)

Issue #858, opened 2018-12-07
Documentation Improvements 


ambrosejcarr:
Note on documentation types: https://www.divio.com/blog/documentation/

neuromusic:
Here's my proposal for the documentation refactor. Open to discussion.

### landing page

Refactor the landing page to facilitate users quickly getting to the places they need to be. I like the seaborn documentation structure:

- A short block of text describing starfish and linking people to specific places in the documentation
- A "contents" list which provides top level navigation to key pages of the documentation
- A "Features" list which provides links to the API and examples/vignettes for each major feature of the package

### Contents

The contents list would contain the following links

- Introduction (an introduction to the motivation for the package and a the key value proposition)
- Installation (how to install starfish, including how to use conda/virtualenv)
- Data Standard (the SpaceTx Image Format Specification)
- *Tutorial* (series of vignettes walking a user through the process of creating a pipeline)
- Examples (link to a gallery of vignettes/examples of using starfish, generated with Sphinx-gallery)
- Command Line (documentation on using the CLI)

### Features

This section would highlight key sections of the API documentation

- Filtering
- Spot-Finding
- Decoding
- Segmenting
- Pipelining

### Tutorial

One need for SpaceTx users is an end-to-end tutorial so that users can create a pipeline recipe. This will be a series of vignettes demonstrating the use of the API

0. Is this for me? (what assays does this support?)
1. Getting data into SpaceTx (needs to be aligned and corrected)
2. Filtering (how to select filter parameters. w/ napari?)
3. Spot-finding (how to select spot-finding parameters. w/ napari?)
4. Decoding (how to select decoding parameters)
5. Segmenting
6. Creating a pipeline recipe
7. Running a pipeline on an experiment
8. Collecting the results of an experiment

ambrosejcarr:
I will go over your proposal in detail. In the mean time, reviewing PRs reminds me that we will need some better documentation for our main objects, orthogonal to the usage workflow. 

`IntensityTable`, `Codebook`, and `ExpressionMatrix` are easier as we can point at external documentation, but we've heard from @berl and @kevinyamauchi that xarray is not very intuitive. It took our team a while to learn their indexing as well, so something that captures:
1.  Where to find [detailed information](http://xarray.pydata.org/en/stable/indexing.html) on how to index things
[...]

PR #1630, opened 2019-10-28
Update vocabulary for LabelImage I/O operations.

ttung:
@ambrosejcarr says we agreed to parrot xarray's convention.  I have no recollection of this, senator.

Test plan: travis


ambrosejcarr:
I guess we didn't agree, but Justin made a good pitch that *I* agreed to:

https://github.com/spacetx/starfish/issues/858#issuecomment-483846145

PR #1638, opened 2019-10-31
Adding AWS Scaling Vignette 

shanaxel42:
Looking for feedback on if this tutorial is the right level of steps needed to get someone setup to process their data, or if it should be more/less detailed. Also if anything is confusing. 

You probs want to look at the rendered tutorial since there's pics.


ttung:
I enabled rtd on your branch  https://spacetx-starfish.readthedocs.io/en/saxelrod-aws-vignette/

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
++++++++++++++++++++++++++
+
+Before we can register our jobs we need to set up an IAM role that has access to AWSBatchServices and
+our newly created s3 bucket. Navigate to the `IAM console <https://console.aws.amazon.com/iam/home>`_ an select *Roles* from the left panel.
+Click *Create Role* we've called ours `spacetx-batch-uploader`. From the list of available services to prevision your role with select *batch*. Then click through the rest of the wizard
+using the default settings and create the role.
+
+We also need to give this role read and write access to our newly created s3 bucket. To do this we make a new policy and attach it to the `spacetx-batch-uploader` role.
+
+Select *policies* form the left hand panel and click *create policy*. Click on the JSON editor and paste in the following code:
```
ttung:
```suggestion
Select *policies* from the left hand panel and click *create policy*. Click on the JSON editor and paste in the following code:
```

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
+
+    --job-queue: The name of your job queue to run your jobs. Ours is "first-run-job-queue"
+
+
+Now we run our script:
+
+::
+
+    $ python3 starfish-workflow.py \
+    >     --experiment-url "https://s3.amazonaws.com/aws-processing-example/iss-spacetx-formatted/experiment.json" \
```
ttung: This accesses s3 through http.  Does the bucket policy even apply in this case?

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Follow the `Getting Started Guide <http://docs.aws.amazon.com/batch/latest/userguide/Batch_GetStarted.html>`_ and ensure you have a valid job queue and compute environment. For this tutorial
+we used the default parameters (our job queue is still called first-run-job-queue).
+
+Here's what out Batch Dashboard looks like:
+
+.. figure:: images/aws-tutorial-figure5.png
+   :align: center
+
+Alright now it's time to register our batch jobs. From the template files open up the file named `register-process-fov-job.json`. This file describes a batch job that will create an ec2 instance using the docker container `spacetx/process-fov`
```
ttung: Is this a docker container we have to maintain?

docs/source/processing_at_scale/processing_with_aws.rst
shanaxel42: no but it does for copying the recipe file and then the results back to a bucket. Which is all in the same job

docs/source/processing_at_scale/processing_with_aws.rst
shanaxel42: maybe? It's pretty bare bones though.

docs/source/processing_at_scale/processing_with_aws.rst
ttung: Doesn't that imply the bucket needs to have public access for the HTTP endpoint to work?  I think it might be better to make this work over s3://.  I suspect it might just work.

docs/source/processing_at_scale/processing_with_aws.rst
ttung: will it automatically pick up the latest version of starfish?

shanaxel42: That's a good point.  To keep things up to date we'd need to rebuild them with each new release https://github.com/spacetx/starfish-workflows/blob/master/aws/merge-results-job-docker-files/Dockerfile. Or install starfish within the container

ttung: We could just add this to our release workflow.

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
+    $ python3 starfish-workflow.py \
+    >     --experiment-url "https://s3.amazonaws.com/aws-processing-example/iss-spacetx-formatted/experiment.json" \
+    >     --num-fovs 15 \
+    >     --recipe-location "s3://aws-processing-example/aws-processing-example/iss-spacetx-formatted/recipe.py" \
+    >     --results-bucket "s3://aws-processing-example/iss-spacetx-formatted-results/" \
+    >     --job-queue "first-run-job-queue"
+    Process fovs array job 39a13edd-8cca-4e7e-9379-aa3cf757c72e successfully submitted.
+    Merge results job ac5d49f5-a12e-4176-96e2-f697c6cf0a12 successfully submitted.
+
+To monitor the status of both jobs navigate to the `AWS Batch Dashboard <https://console.aws.amazon.com/batch/home?region=us-east-1#/dashboard>`_. You should see 2
```
ttung: The problem with this is that this encodes the region in the URL.  Is there a way to get a region-agnostic dashboard?

docs/source/processing_at_scale/processing_with_aws.rst
shanaxel42: Do you mean use Experiment.from_json() with an s3 path? How would that work without passing credentials?

docs/source/processing_at_scale/processing_with_aws.rst
shanaxel42: I think https://console.aws.amazon.com/batch/home will automatically navigate to the dashboard, so I'll jsut change it to that

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
+                "Resource": [
+                    "arn:aws:s3:::aws-processing-example/*"
+                ]
+            }
+        ]
+    }
+
+
+Name your policy and save it, we named our `spacetx-batch-uploader`. Now navigate back to your new role and attach your s3 uploader policy.
+Our `spactx-batch` role summery now looks like this:
```
ttung:
```suggestion
Our `spacetx-batch` role summery now looks like this:
```

docs/source/processing_at_scale/processing_with_aws.rst
ttung: If it's run from an EC2 instance with an IAM role, it should just work.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

`the AWS SDKs, AWS CLI, and Tools for Windows PowerShell automatically get the credentials from the EC2 instance metadata service and use them`

docs/source/processing_at_scale/processing_with_aws.rst
ttung: just noticed this:

```suggestion
Our `spacetx-batch` role summery now looks like this:
```

docs/source/processing_at_scale/processing_with_aws.rst
shanaxel42: You're right it works!

shanaxel42:
@ttung am I cool to merge this?

docs/source/processing_at_scale/processing_with_aws.rst
ttung: Can you validate that this works on a bucket without public read access? I believe it should but it would be good to.

docs/source/processing_at_scale/processing_with_aws.rst
```diff
@@ -0,0 +1,351 @@
...
+      }
+    }
+
+NOTE: if your starfish processing is memory expensive you can adjust the allocated memory for each created instance using the `memory` parameter.
+
+Then from the directory where this file lives run the following command:
+
+``aws batch submit-job --cli-input-json file://register-process-fov-job.json``
+
+You can check that your jobs had been successfully registered by navigating to the `Job Definitions page <https://console.aws.amazon.com/batch/home?region=us-east-1#/job-definitions>`_.
```
ttung: this was missed.

shanaxel42: : o man I swear I changed this, my bad

PR #1666, opened 2019-11-21
Linear image leveling

ttung:
Clip is nice and useful, but doesn't allow for the expansion of dynamic range when the peak intensity is less than fully saturated.

This PR adds linear image leveling.  Five modes are offered:

1. Clip supersaturated values.
2. Scale by the peak intensity of the entire image if any part of the image is supersaturated.
3. Scale by the peak intensity of each image chunk if any part of the image chunk is supersaturated.
4. Scale by the peak intensity of the entire image.
5. Scale by the peak intensity of each image chunk.

1-3 matches the existing modes of Clip, so we reimplement Clip using the new linear image leveler.

Test plan: Add test cases for the new image leveling approaches.  Existing Clip tests pass.

### DIFF
--- a/starfish/core/image/Filter/element_wise_mult.py
+++ b/starfish/core/image/Filter/element_wise_mult.py
@@ -6,7 +6,7 @@ import xarray as xr
 
 from starfish.core.imagestack.imagestack import ImageStack
 from starfish.core.types import Clip
-from starfish.core.util.dtype import preserve_float_range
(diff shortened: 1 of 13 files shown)


neuromusic:
if this reimplements Clip, should we plan to deprecate Clip?

ttung:
> if this reimplements Clip, should we plan to deprecate Clip?

Absolutely.  I'm marking it deprecated in my next PR.

PR #1682, opened 2019-12-02
add axis labels to display()

kevinyamauchi:
# Description
This PR adds axis labels to `starfish.display()` using the recently-added labeled axis support in `napari` (thanks, @HagaiHargil !).

# Approach
I added the `axis_labels` keyword argument to the `Viewer` constructor in `_display.py`. 

# Question
Currently, I just create a list of the `Axes` `Enum` values. Is there a better way to do this? Maybe get the axes from the stack?

# Testing:
I downloaded and viewed the test data from the ISS example notebook.

### DIFF
--- a/starfish/core/_display.py
+++ b/starfish/core/_display.py
@@ -17,7 +17,7 @@ except ImportError:
     Viewer = None
 
 
-NAPARI_VERSION = "0.2.0"  # when changing this, update docs in display
+NAPARI_VERSION = "0.2.6"  # when changing this, update docs in display
 INTERACTIVE = not hasattr(__main__, "__file__")
 
 
@@ -226,7 +226,15 @@ def display(
         from qtpy.QtWidgets import QApplication
         app = QApplication.instance() or QApplication([])
 
-        viewer = Viewer()
+        viewer = Viewer(
+            axis_labels=[
+                Axes.ROUND.value,
+                Axes.CH.value,
+                Axes.ZPLANE.value,
+                Axes.Y.value,
+                Axes.X.value
+            ]
+        )
         new_viewer = True
     elif isinstance(viewer, Viewer):
         new_viewer = False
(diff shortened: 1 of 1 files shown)


ttung:
Do we need to upgrade the napari version pin?

kevinyamauchi:
@ttung, yeah we would need to bump to napari 0.2.6. I changed the constant in `_display()`. Where is the version pinned for `pip install starfish[napari]`? I am in the process of updating the docstrings.

kevinyamauchi:
@ttung, if I read `setup.py` correctly, the `napari` version comes from the `_display.py`, so I think we are good there. I just updated the `display()` docstring.

starfish/core/_display.py
```diff
@@ -226,7 +226,15 @@ def display(
         from qtpy.QtWidgets import QApplication
         app = QApplication.instance() or QApplication([])
 
-        viewer = Viewer()
+        viewer = Viewer(
+            axis_labels=[
+                Axes.ROUND.value,
+                Axes.CH.value,
+                Axes.ZPLANE.value,
```
ttung: I would grab these three from starfish/core/imagestack/dataorder.py

kevinyamauchi: Will do!

kevinyamauchi: @ttung , it looks like `AXES_DATA` contains names and orders for round, channel, and z, but not x and y. Should I just assume (y, 3) and (x, 4)? Or is the axis ordering for y and x stored elsewhere?

kevinyamauchi: (you can rotate views in `napari`, so the y and x axes may get displayed)

ttung: Unfortunately, Y and X are not stored anywhere.  It has always been assumed Y then X.

kevinyamauchi: Sounds good!

kevinyamauchi: @ttung, I changed it such that the axis order comes from `starfish.core.imagestack.dataorder`.

PR #1684, opened 2019-12-02
Mechanism to reduce multiple masks into one

ttung:
This is the counterpart to the apply provided in #1655.

Depends on #1686 

Test plan: Added a test that reduced using xors. Because we have boring test data that don't overlap, this is one way of generating interesting output.

### DIFF
--- a/starfish/core/morphology/Filter/reduce.py
+++ b/starfish/core/morphology/Filter/reduce.py
@@ -1,0 +1,94 @@
+from typing import Callable, Optional, Tuple, Union
+
+import numpy as np
+
+from starfish.core.morphology.binary_mask import BinaryMaskCollection
+from starfish.core.types import FunctionSource, FunctionSourceBundle
+from ._base import FilterAlgorithm
+
+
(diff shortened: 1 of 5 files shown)


starfish/core/morphology/Filter/reduce.py
```diff
@@ -0,0 +1,95 @@
...
+import numpy as np
+
+from starfish.core.morphology.binary_mask import BinaryMaskCollection
+from starfish.core.types import FunctionSource, FunctionSourceBundle
+from ._base import FilterAlgorithm
+
+
+class Reduce(FilterAlgorithm):
+    """
+    Reduce the masks from one ``BinaryMaskCollection`` using by applying a specified function
```
shanaxel42:
```suggestion
    Reduce the masks to one ``BinaryMaskCollection`` by applying a specified function
```

ttung: I have attempted to clarify the language here.  It is indeed taking the masks from a single `BinaryMaskCollection` but I can see how the wording is confusing.

PR #1686, opened 2019-12-05
Make map/reduce APIs more intuitive

ttung:
Right now, specifying a FunctionSource along with function parameters makes for a confusing function call.

For instance, `Map('divide', 2, module=FunctionSource.np)` means the FunctionSource comes last, which is not intuitive.

This adds the ability for `FunctionSource`s to be called and return a Bundle that includes both the package name and the function name.  The call above would then become: `Map(FunctionSource.np('divide'), 2)`.

Backwards compatibility with the prior API is maintained, but a warning is generated.

If the top-level package is provided twice, it is treated as an error.

Test plan: added test cases to cover the new approach, the old approach that should generate the warning, and the ugly combination that should fail.


shanaxel42:
I actually didn't find the old method confusing. Was it breaking anything? I don't have a huge preference if you think this makes more sense but seems like a lot of changes

ttung:
It gets even more confusing when the function takes kwargs, because it becomes:

`Reduce('function_name', function_positonal_arg0, function_positonal_arg1, module=FunctionSource.blahblah, function_kwarg_0=val, function_kwarg_1=val)`

shanaxel42:
> It gets even more confusing when the function takes kwargs, because it becomes:
> 
> `Reduce('function_name', function_positonal_arg0, function_positonal_arg1, module=FunctionSource.blahblah, function_kwarg_0=val, function_kwarg_1=val)`

Ahhh ok then I support this.

Issue #1273, opened 2019-04-25
PixelSpotDecoder does not add `spot_id` to IntensityTable


ambrosejcarr:
... `spot_id` is needed for labeling an intensity table after segmentation. 

Reproduce: replace BlobDetection with PixelSpotDecoding in the ISS notebook. 

We should add tests for this.

PR #1687, opened 2019-12-05
[easy] adding spot ids to pixel results

shanaxel42:
Fixes: #1273 

Added small test


starfish/core/spots/DetectPixels/combine_adjacent_features.py
```diff
@@ -342,6 +342,7 @@ def _create_spot_attributes(
             passes_filter = np.array(passes_area_filter, dtype=np.bool)
 
             spot_attributes = SpotAttributes(pd.DataFrame.from_records(spot_attrs))
+            spot_attributes.data[Features.SPOT_ID] = list(range(len(spot_attributes.data)))
```
ttung:
```suggestion
            spot_attributes.data[Features.SPOT_ID] = np.arange(0, len(spot_attributes.data))
```

PR #1692, opened 2019-12-10
Pipeline component and implementation for merging BinaryMaskCollections

ttung:
This does not handle the case where the pixel/physical ticks do not line up.  This is for cases where we derive the data from the exact same set of images, and inherit their pixel/physical ticks from the same source.

Depends on #1690, #1691, #1696
Test plan: Added one positive test case, and tested the cases where the merge should fail.

### DIFF
--- a/starfish/core/morphology/Merge/__init__.py
+++ b/starfish/core/morphology/Merge/__init__.py
@@ -1,0 +1,10 @@
+"""Algorithms in this module merge multiple BinaryMaskCollections together."""
+from ._base import MergeAlgorithm
+from .simple import SimpleMerge
+
+# autodoc's automodule directive only captures the modules explicitly listed in __all__.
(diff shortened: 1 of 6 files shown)


starfish/core/morphology/Merge/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Algorithms in this module merge multiple BinaryMaskCollections together."""
+from ._base import MergeAlgorithm
+from .simple import SimpleMerge
+
+# autodoc's automodule directive only captures the modules explicitly listed in __all__.
+all_filters = {
```
shanaxel42:
```suggestion
all_methods = {
``` 
? are they all filters?

ttung: You are correct.  Will fix.

PR #1701, opened 2019-12-14
Use the new watershed component to finish componentizing the existing watershed

ttung:
Use the component from #1700 to replace the remaining procedural code in the existing watershed component.

Depends on #1700, #1699, #1694
Test plan: visually the ISS notebook looks the same.


ttung:
> Are you planning on keeping this api?

Apparently yes.

Issue #1726, opened 2020-01-09
xarray >= 0.14


berl:
#### Description
if xarray version is not  ->= 0.12-  >=0.14?  , `IntensityTable`'s `drop_sel` method doesn't exist.
edit: not sure when it was introduced but 0.12 doesn't have it  and 0.14 does...

PR #1683, opened 2019-12-02
[easy] update requirements

ttung:

PR #1727, opened 2020-01-10
Move xarray to 0.14.1

ttung:
`DataArray.drop_sel` was added in 0.14.1.  The old API (removed in #1683) is being deprecated so we are want the new API.  However, that is only added in 0.14.1.

Fixes: #1726
Test plan: alltest branch


ttung:
Replaced by #1728

PR #1730, opened 2020-01-10
Force version of dataclasses to .6

shanaxel42:


ttung:
Can you verify as part of your test plan that python 3.7 + dataclasses doesn't have any averse effects?

REQUIREMENTS.txt
```diff
@@ -4,7 +4,7 @@
 # installed, and bombs after the fact.
 python-dateutil < 2.8.1
 click
-dataclasses;python_version<"3.7"
+dataclasses;python_version==.6
```
ttung: shouldn't this just be `dataclasses == 0.6`?

shanaxel42: yuuuup

shanaxel42:
> Can you verify as part of your test plan that python 3.7 + dataclasses doesn't have any averse effects?

I'm currently using `3.7.3` on my computer and things are chill, do you want me to add a test?

REQUIREMENTS.txt
```diff
@@ -4,7 +4,7 @@
 # installed, and bombs after the fact.
 python-dateutil < 2.8.1
 click
-dataclasses;python_version<"3.7"
+dataclasses=0.6
```
ttung: i think this needs to be a double equal.

ttung:
> I'm currently using 3.7.3 on my computer and things are chill, do you want me to add a test?

nope, just wanted a quick sanity check.

PR #1731, opened 2020-01-13
Adding method to binary mask collection that imports external labeled images

shanaxel42:
Ex. if someone saves the output of ilastik object classification as a tiff file

### DIFF
--- a/starfish/core/morphology/binary_mask/binary_mask.py
+++ b/starfish/core/morphology/binary_mask/binary_mask.py
@@ -20,9 +20,11 @@ from typing import (
 
 import numpy as np
 import xarray as xr
+from skimage import io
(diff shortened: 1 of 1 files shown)


starfish/core/morphology/binary_mask/binary_mask.py
```diff
@@ -262,6 +264,30 @@ def from_label_image(cls, label_image: LabelImage) -> "BinaryMaskCollection":
             log,
         )
 
+    @classmethod
+    def from_external_labeled_image(cls, path_to_labeled_image: Union[str, Path],
```
ttung: Can you write a docs for this?

shanaxel42: yup!

PR #1735, opened 2020-01-14
Support old argument style for AreaFilter

### LINKED ISSUE #1734: area filter change not backwards compatible
#### Description
Manifested in https://github.com/spacetx/starfish/pull/1733

ttung:
This re-adds the positional argument style for specifying AreaFilter, while displaying a warning that it's deprecated.

Fixes #1734
Test plan: Add tests to cover the old style and the illegal path of specifying both positional and keyword arguments


ttung:
> When are we planning on deleting the positional ones again?

I guess after a release or two. :)

Issue #1725, opened 2020-01-09
add cultured mouse & human fibroblast dataset


neuromusic:
For demos and tutorials, we often use an ISS dataset with cultured cells, but this dataset is not available in `starfish.data`. We should provide API access to this dataset so we can use it for documentation and users can have a dataset with ground-truth indicators of cell types.

@dganguli 's description of these data...

> data correspond to sequening 4 bases of the beta-actin gene in co-cultured mouse and human fibroblasts. these 4 bases only differ in one position, a SNP between species. as such, these data offer ground truth information to validate a cell typing exercise

Path to spacetx experiment file: https://d2z4zivcmlmaj1.cloudfront.net/ISS/mouse_vs_human/experiment.json

Citation: https://www.ncbi.nlm.nih.gov/pubmed/23852452

neuromusic:
closed with #1741

PR #1741, opened 2020-01-21
adding mouse v human example to starfish.data

shanaxel42:
#1725

### DIFF
--- a/starfish/data.py
+++ b/starfish/data.py
@@ -53,6 +53,22 @@ def allen_smFISH(use_test_data: bool = False) -> Experiment:
         "https://d26bikfyahveg8.cloudfront.net/smFISH/mouse/formatted_with_DAPI/experiment.json")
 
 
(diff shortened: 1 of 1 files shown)


neuromusic:
Do we have any tests for the datasets module? Verifying the hash of a FOV or something?

shanaxel42:
> Do we have any tests for the datasets module? Verifying the hash of a FOV or something?

No - but those tests would take a long time.

ttung:
> Do we have any tests for the datasets module? Verifying the hash of a FOV or something?

spacetx file format is somewhat resilient against data corruption.  I think the goal of tests for `starfish.data` should be to verify that the data is *available*.  If we can verify that the top level `experiment.json` is accessible, that is good enough for me.  More than that requires more effort than I think is worth doing.

neuromusic:
okay, that all makes sense.

Let's get this in and add some data availability tests as part of the durability epic.

Issue #1392, opened 2019-06-04
ExpressionMatrix construction propagates "nan" spots


ambrosejcarr:
## Issue
Decoding uses "nan" as the placeholder for a gene id that does not decode. When Expression Matrices are constructed, we propagate "nan" as a gene value, which typically has a high count, as usually about half of the spots do not decode. 

## Proposal 
"nan" should not be propagated, but should be recorded as a metadata that reflects the number of spots in a cell that did not decode.

neuromusic:
Based on conversation w/ @mattcai & @shanaxel42 ...

we should include both counts of "total spots" and "unassigned/non-decoded spots" as metadata

PR #1791, opened 2020-02-06
removing nan propagation to expression matrix

shanaxel42:
Fixes: #1392

Removes 'nan' as a target value in ExpressionMatrix and reformats is as a metadata column called 'number_of_undecoded_spots`.  Also renames the 'nan' cell_id row to "unassigned".


starfish/core/intensity_table/decoded_intensity_table.py
```diff
@@ -153,6 +153,13 @@ def to_expression_matrix(self) -> ExpressionMatrix:
                 "can be called. See starfish.spots.AssignTargets.Label.")
         grouped = self.to_features_dataframe().groupby([Features.CELL_ID, Features.TARGET])
         counts = grouped.count().iloc[:, 0].unstack().fillna(0)
+        # rename unassigned spots
+        counts.rename(index={'nan': 'unassigned'}, inplace=True)
+        # remove and store 'nan' target counts
+        nan_target_counts = np.zeros(counts.shape[0])
+        if 'nan' in counts.index:
```
neuromusic: this should check the columns, not the index, right?

PR #1796, opened 2020-02-06
Add group_by for tilefetcher-based ImageStack construction

ttung:
When we construct an ImageStack from a tilefetcher, we now allow accept a `group_by` parameter which controls how we load the data.  If an axis is in the `group_by`, then we load all the data for any given value for that axis before moving on to another value for that axis.

For instance, if Axes.ROUND is in `group_by`, then we will load all the data for round=0 before we load the data for round=1.

Test plan: Added tests to verify that the loader works with a variety of `group_by`s.  Dan Goodwin's 3D data loads in 100seconds now instead of 10 minutes.


ttung:
@mattcai  fyi this just landed!

PR #1799, opened 2020-02-07
Clarify existing rules for requirements

ttung:
1. `pin-requirements` takes into consideration the existing pins, and repins only the minimal set of changes required to satisfy the top-level requirements.  Run this when you add a new requirement.
2. `pin-all-requirements` wipes away all the pins and pins based on the latest.  Do this periodically so we don't have stale requirements.

Test plan: run both rules manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,20 +78,24 @@ help-docs:
 
 ### REQUIREMENTS #############################################
 #
-check-requirements:
-	if [[ $$(git status --porcelain REQUIREMENTS* requirements/REQUIREMENTS* starfish/REQUIREMENTS*) ]]; then \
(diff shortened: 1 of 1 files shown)


shanaxel42:
`pin-requirements takes into consideration the existing pins, and repins only the minimal set of changes required to satisfy the top-level requirements` 

What does satisfy the top level requirements mean? It just updates the new ones?

ttung:
> > pin-requirements takes into consideration the existing pins, and repins only the minimal set of changes required to satisfy the top-level requirements

> What does satisfy the top level requirements mean? It just updates the new ones?

top-level requirements are the requirements in REQUIREMENTS.txt.

PR #1808, opened 2020-02-10
[easy] Remove workaround for non-3D images

### LINKED ISSUE #1778: [todo] improvements to TrackpyLocalMaxPeakFinder
https://github.com/spacetx/starfish/blob/47de54081c61e68932364a7fb53ed0f6cb5752b5/starfish/core/spots/FindSpots/trackpy_local_max_peak_finder.py#L133

ttung:
ImageStacks are always 5D, so spot detection should always expect a 3D images (no R/C).

Fixes #1778


shanaxel42:
ok wait I approved this but now I have a question. `spot detection should always expect a 3D images` this  isn't true if someone runs spot detection with 2d tiles?

PR #1823, opened 2020-02-28
[Easy] - Changing return value of PixelDecoding to DecodedIntensityTable

shanaxel42:


mattcai:
PixelDecoding should return a DecodedIntensityTable to match SpotDecoding and support methods like `to_expression_matrix()` but it was incorrectly returning an IntensityTable.

PR #1830, opened 2020-03-15
Add support for kwargs to peak_local_max

ttung:
Test plan: run spot detection on a spot that's uniformly illuminated.  Without the `num_peaks_per_label=1` kwarg, it finds a peak at each pixel.  With the kwarg, it finds only 1 peak.

### DIFF
--- a/starfish/core/spots/FindSpots/local_max_peak_finder.py
+++ b/starfish/core/spots/FindSpots/local_max_peak_finder.py
@@ -50,6 +50,8 @@ class LocalMaxPeakFinder(FindSpotsAlgorithm):
     verbose : bool
         If True, report the percentage completed during processing
         (default = False)
+    kwargs :
+        Additional keyword arguments to pass to skimage.feature.peak_local_max
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


starfish/core/spots/FindSpots/local_max_peak_finder.py
```diff
@@ -309,7 +317,7 @@ def run(
         n_processes : Optional[int] = None,
             Number of processes to devote to spot finding.
         """
-        spot_finding_method = partial(self.image_to_spots, *args, **kwargs)
+        spot_finding_method = partial(self.image_to_spots, **self.kwargs)
```
mattcai: If *args and **kwargs are no longer used here, should they be removed from `run()` parameters?

ttung: The issue is that the superclass's `run()` method has `*args, **kwargs`.

mattcai: I see, thanks for explaining.

Issue #1836, opened 2020-03-24
local_max_peak_finder throws error when `is_volume=False`


mattcai:
#### Description
When using `LocalMaxPeakFinder` to find spots on a 2D image or 2D slices of 3D volume, indices in `masked_image` doesn't match size. In addition, when finding spots on 3D image volume, the wrong label region of `masked_image` is set to `False`.

#### Code to Reproduce 2D image issue
[gist](https://gist.github.com/mattcai/9b75a635e764070e5f1c51f47db8d92e)

#### Expected Results
No error is thrown. `lmp_spots` contains hundreds of spots.

#### Actual Results
Error message when finding spots on 2D slices of 3D volume (`is_volume=False`)
```
~/starfish-CI/starfish/starfish/core/spots/FindSpots/local_max_peak_finder.py in image_to_spots(self, data_image, **kwargs)
    244         for spot_prop in spot_props:
    245             if spot_prop.area < self.min_obj_area or spot_prop.area > self.max_obj_area:
--> 246                 masked_image[0, spot_prop.coords[:, 0], spot_prop.coords[:, 1]] = 0
    247 
    248         # store re-calculated regionprops and labels based on the area-masked image

IndexError: too many indices for array
```

PR #1839, opened 2020-03-24
Fix localmaxpeakfinder spot_props filter

mattcai:
Fixed filtering of `spot_props` whose areas are too large or too small in `image_to_spots` of `local_max_peak_finder`. It can now handle 2D xarrays (i.e. `is_volume=False`) and can correctly set pixels to `False` (removing a `spot_prop`) for 3D xarrays.

In addition, added details for how to set `local_max_peak_finder` parameters in its tutorial.


neuromusic:
fixes #1836

ttung:
please remove the "expression" file before you commit, however.

PR #1895, opened 2020-06-30
user guide clean up

mattcai:
Removed low priority unfinished examples and sections in the User Guide. Cleaned up old content that is now redundant. Added white top-hat examples to removing autofluorescence and illumination correction.

see: https://spacetx-starfish.readthedocs.io/en/mcai-userguide-cleanup/


neuromusic:
LGTM!

PR #1928, opened 2020-12-10
Add necessary coords for IntensityTable when using Nearest Neighbors strategy

mattcai:
A starfish user wants to try the Nearest Neighbor Trace Builder but had issues because the `IntensityTable` was [missing the `spot_id` coord and `features` dim](https://forum.image.sc/t/unable-to-retrieve-values-from-decodedintensitytable/44913). This pull request adds those to the `IntensityTable` so a full starfish pipeline (as demonstrated in the [Quick Start tutorial](https://spacetx-starfish.readthedocs.io/en/latest/gallery/quick_start/plot_quick_start.html)) can be built and run. This does not necessarily mean the Nearest Neighbor Trace Builder will give satisfactory results when tested on real data.

### DIFF
--- a/starfish/core/spots/DecodeSpots/trace_builders.py
+++ b/starfish/core/spots/DecodeSpots/trace_builders.py
@@ -75,7 +75,7 @@ def build_traces_sequential(spot_results: SpotFindingResults, **kwargs) -> Inten
     return intensity_table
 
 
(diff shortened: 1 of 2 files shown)


neuromusic (approved):
LGTM

PR #1935, opened 2021-02-18
Upgrade to a consistent version of pip in all workflows

ttung:
Always upgrade to $PIP_VERSION.

A couple minor changes:
1. Add a pip freeze to the test workflows to ensure we know what's being installed.
2. Do not include pip in requirements.txt because windows freaks out if it's asked to install pip in a requirements.txt file.


neuromusic (approved):
LGTM

PR #1933, opened 2021-02-17
Import match_histograms from skimage.exposure

ttung:
It was [moved to skimage.exposure @ 0.16.0](https://scikit-image.org/docs/0.16.x/release_notes_and_installation.html#deprecations)

Fixes: #1932

Test plan: Travis

PR #1941, opened 2021-03-23
Support skimage version higher than 0.17.2

pjb7687:
@berl


ttung:
#1933 is already in the repo.

berl:
thanks @ttung 
so it works from here but not the current `pypi` release

berl:
@ttung is this reason enough for a release?

ttung:
Yeah, I think @neuromusic was going to cut a release after he made some doc changes

PR #1957, opened 2021-09-07
[WIP] Replace Travis CI with Github Actions

njmei:
Because it looks like the Travis CI for this repo is no longer running, this PR proposes to replace it with the free (for public repos) `github-actions`.

The current configuration adds python 3.7 and 3.8 tests because the current travis config only tests python 3.6 and there was no test visibility at all for python 3.7 or above, so there was no idea of what broke in 3.8+.

Going forward, this project may need to drop support for python 3.6/3.7 due to pretty drastic changes in `xarray` and `scikit-image`.


njmei:
Hmm... this needs some more work before it would be ready for primetime. Closing for now. Will resubmit the PR once everything 'works' and commits are cleaned up.

PR #1961, opened 2021-10-13
SpotFindingResults Serialization to Disk (and loading)

ctcisar:
I have added two simple methods to save `SpotFindingResults` to disk and reload them.  This is handled by calling the methods in `physical_coord_ranges`, `log`, and `_results` to save each object, and then writing a `json` file that identifies the indices and paths to each of these files.  The loader opens this `json` and uses the loading method from each of said objects to recreate the original `SpotFindingResults` object.

This can be useful for storing, comparing, and troubleshooting intermediate results.  Storing the contents of each `xarray`-like object as a `netcdf` allows for compatibility across most package version differences, unlike saving the entire `SpotFindingResults` to a Python pickle.

### DIFF
--- a/starfish/core/types/_spot_finding_results.py
+++ b/starfish/core/types/_spot_finding_results.py
@@ -121,6 +121,7 @@ class SpotFindingResults:
         """
         json_data: Dict[str, Any] = {}
 
+        pwd = os.getcwd()
         os.chdir(os.path.dirname(output_dir_name))
         base_name = os.path.basename(output_dir_name)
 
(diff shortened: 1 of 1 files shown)


njmei:
Would it be possible to not write the *.json file and instead write *all* the `SpotFindingResults` data (including log info) into 1 netcdf file?

berl:
@ctcisar  thanks for this!  Making it easier to view intermediate results is super useful. I think @njmei may have also recently mentioned something like this being missing.  
It would also be nice to have some test coverage here- something simple like https://github.com/spacetx/starfish/tree/master/starfish/core/types/test that generates some data, saves it and reloads it using your new methods would be great.

ctcisar:
@berl I added a testing file in the same format as the provided test, hopefully it's up to spec.

In the process of writing this, I realised that alternating `(obj type Log).encode()` and `Log.decode()` will result in each result getting nested in an additional dict, structured as `{"log": [data]}`.  I do not know if this is intended behavior, but have fixed the loader to accommodate for this.

PR #1965, opened 2021-10-19
Fix coords assignment in CombineAdjustFeatures.run

sgratiy:
This PR fixes a bug in with the `CombineAdjustFeatures.run` that was flipping round and channel coordinates in the resulting `DecodedIntensityTable` compare to the input intensities.

The fix:
Use explicit parameter calls when building table coordinates
Correct shape definition for the IntensityTable docstring

This [notebook](https://gist.github.com/sgratiy/2053e72cfc0ff4ca6460ef4f981dc45f) shows demonstrates the intensity table coordinates before and after the fix.

### DIFF
--- a/starfish/core/intensity_table/intensity_table.py
+++ b/starfish/core/intensity_table/intensity_table.py
@@ -33,7 +33,7 @@ class IntensityTable(xr.DataArray):
     An IntensityTable records the numeric intensity of a set of features in each
     :code:`(round, channel)` tile in which the feature is identified.
     The :py:class:`IntensityTable` has shape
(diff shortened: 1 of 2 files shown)


sgratiy:
@berl I have added the notebook demonstrating the problem in the main description.

berl:
thanks @sgratiy.
For the record, this was indeed a bug in how  `starfish/core/spots/DetectPixels/combine_adjacent_features.py` created a `DecodedIntensityTable`. This was previously missed because users (and the starfish codebase) never did anything with the `IntensityTable` part of a `DecodedIntensityTable` created in this way. I'll revert the revert and we can pretend like this never happened.

PR #1984, opened 2022-06-22
Clone of https://github.com/spacetx/starfish/pull/1978

ttung:
This is just to trigger the CI.


ttung:
@nickeener it looks like this is passing now.  Ideally, you should just copy my changes into your PR and land them in the context of your PR.  @berl any concerns about merging?

PR #1985, opened 2022-07-28
Various Small Fixes/Improvements

nickeener:
After working with starFISH extensively there have been a few bugs/lack of features that I've noticed and simply fixed for myself. This PR contains 3 such changes plus a small fix for the `CheckAll` decoder to return an empty `DecodedIntensity` when it finds no targets instead of causing an obscure error, and a replacement of the seqFISH notebook that demonstrates the new decoder.

The 3 fixes/improvement include:

SpotFindingResults saving/loading - This includes the changes suggested in #1961 (with the permission of the original author). I've simply bundled those into this PR. Files changed: `starfish/core/types/_spot_finding_results.py`

Loading an 3d external label image - Added functionality so that a 3D external label image could be used to create a `BinaryMaskCollection` object: Files changed: `starfish/core/morphology/binary_mask/binary_mask.py`

2D `BlobDetector` - Currently `BlobDetector` will give incorrect results when the `is_volume` parameter is set to False. This prevents users from finding spots in a 3D image by treating each z slice as a separate 2D image. Changes add this functionality. Files changed: `starfish/core/spots/FindSpots/blob.py`


nickeener:
I've also added a few more improvements to my CheckAll decoder that should reduce the run time significantly

ttung:
@berl Could you review this? I'd like your stamp before I merge this.

berl (approved):
looks good to me. Sorry for the lag on this

Issue #2006, opened 2023-11-15
MinDistanceLabel does not enforce minimum distance, leading to over-segmentation


iimog:
#### Description
The minimum distance passed to `starfish.morphology.Filter.MinDistanceLabel` is not enforced. In particular, if multiple local maxima exist with identical values, all of them are retained, independent of their distance. This leads to surprising cases of over-segmentation, where perfectly contiguous masks are split into multiple components (see example below).

#### Steps/Code to Reproduce
Using this exemplary mask (saved as `mask.png`):
![mask](https://github.com/spacetx/starfish/assets/7403236/49f361c1-bf45-40bf-bb45-6575d78ca88e)

```python
import numpy as np
import skimage as ski
import matplotlib.pyplot as plt
from starfish import ImageStack
from starfish.morphology import Binarize, Filter

# Load the image, transform it to a BinaryMaskCollection and apply the MinDistanceLabel
img = ImageStack.from_numpy(np.expand_dims(ski.util.img_as_float32(ski.io.imread("mask.png")), (0,1,2)))
binarized = Binarize.ThresholdBinarize(.5).run(img)
masks = Filter.MinDistanceLabel(120, 120).run(binarized)

print("Nuclei found:", len(list(masks.masks())))
plt.imshow(masks.to_label_image().xarray.squeeze(),interpolation="nearest")
```

#### Expected Results

Two clearly separated nuclei.
![properseg](https://github.com/spacetx/starfish/assets/7403236/ba7d4d8c-d99c-4df3-b7d8-ec7855cc188e)

#### Actual Results

Four detected nuclei, the top one separated into three distinct areas (the small line between the top and bottom has its own class).
![overseg](https://github.com/spacetx/starfish/assets/7403236/e12d9c1f-8c23-4900-9433-1b4a817afd15)

#### Cause and possible solution
The cause of this problem is, that the internally called `skimage.filter.peak_local_max` (https://github.com/spacetx/starfish/blob/853f56c7c02b15397adb921db5e3bde02fdadb63/starfish/core/morphology/Filter/min_distance_label.py#L61-L66) is not passed an appropriate value for `min_distance`, so the default value of 1 is used. Setting the `footprint` according to the minimum_distance ensures, that there can be only one maximum value in the respective area, but if that value occurs multiple times, all occurrences are returned as local maxima and thus used as seeds in a watershed. This is certainly undesirable as it leads to cases of over-segmentation, as exemplified. 
[...]

PR #2008, opened 2023-11-17
Add min_distance parameter to peak_local_max call

iimog:
Before this change, multiple local maxima were returned even within the min_distance if their values were identical. This lead to over-segmentation in some cases.

See #2006

### DIFF
--- a/REQUIREMENTS.txt
+++ b/REQUIREMENTS.txt
@@ -22,6 +22,7 @@ scipy
 showit >= 1.1.4
 slicedimage==4.1.1
 sympy ~= 1.5.0
+urllib3 <1.27, >=1.25.4
 tqdm
 trackpy
 validators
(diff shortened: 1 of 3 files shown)


berl:
@nickeener  @ctcisar  do you have any recommendations for the CI fails above?

nickeener:
@berl the failure appears to be in the ```starfish/test/full_pipelines/api/test_iss_api.py::test_iss_pipeline_cropped_data``` test which is not something that we ever messed with, sorry.

iimog:
I found the problem in the iss test. Indeed, the assumption on the number of detected cells had to be updated from 6 to 4. I think this makes sense. The previous pipeline output was this:
iss_old

The new pipeline output is this:
iss_new

So the number of detected cells did change from 6 to 4.

iimog:
The docker build could not install all dependencies via pip because of this error:
```
ERROR: botocore 1.32.6 has requirement urllib3<1.27,>=1.25.4; python_version < "3.10", but you'll have urllib3 2.0.7 which is incompatible.
```
Pinning `urllib3` to that range in REQUIREMENTS fixes the build.

berl:
this is awesome!  all that work put in by _many_ people  making end-to-end tests of these methods did exactly what it was supposed to. I'll review and merge

berl (approved):

PR #2018, opened 2024-09-06
Dropping support for Python 3.8 and setting Python 3.9 as minimum

shachafl:
Due to security issues of upstream dependencies that stopped supporting Python 3.8, such as scikit-learn, etc.
I have updated requirements files, texts and others to support this update.

This PR should also close PR #2015


README.rst
```diff
@@ -58,8 +58,8 @@ starfish supports python 3.8-11 and can easily be installed from PyPI:
 Python Version Notice
```
berl, line 58: I think we can leave out this  "Python Version Notice" unless we see something imminent with 3.9 support going away.

berl (approved):
looks good.  I ran pytest on this on an M1 mac and it also passed with
`====== 372 passed, 3 skipped, 8 deselected, 65 warnings in 1165.18s (0:19:25) ====`
I made one minor comment above but otherwise ready to merge.

README.rst
shachafl: I wanted to discourage new users from picking python 3.9 for starfish as the scientific python community already recommended dropping support for python 3.9 at the end of 2023: https://scientific-python.org/specs/spec-0000/#2023---quarter-4
It is probably just a matter of time, when a critical security issue with a dependency will force us to drop python 3.9.

But I don't feel strongly about it, if you think it is better to take the warning out.

Issue #2022, opened 2024-10-07
2d radius calculation in blob detection (blob.py) is incorrect


shachafl:
#### Description
The radius calculated for blobs in 2d and 3d based on pixels should be different, but instead, it  uses the same formula (probably a copy+paste error by the author).
This can be easily corrected by changing the 2d case from np.sqrt(3) to np.sqrt(2).
See Notes of https://scikit-image.org/docs/stable/api/skimage.feature.html#skimage.feature.blob_log

#### Relevant Code
starfish/starfish/core/spots/FindSpots/blob.py

```
def image_to_spots(
    ...
    # measure intensities
        data_image = np.asarray(data_image)
        if self.is_volume:
            z_inds = fitted_blobs_array[:, 0].astype(int)
            y_inds = fitted_blobs_array[:, 1].astype(int)
            x_inds = fitted_blobs_array[:, 2].astype(int)
            radius = np.round(fitted_blobs_array[:, 3] * np.sqrt(3))
            intensities = data_image[tuple([z_inds, y_inds, x_inds])]
        else:
            z_inds = np.asarray([0 for x in range(len(fitted_blobs_array))])
            y_inds = fitted_blobs_array[:, 0].astype(int)
            x_inds = fitted_blobs_array[:, 1].astype(int)
            radius = np.round(fitted_blobs_array[:, 2] * np.sqrt(3))
            intensities = data_image[tuple([y_inds, x_inds])]
    ...
```

PR #2023, opened 2024-10-08
fixing blob radius calculation in 2d

shachafl:
Replaced np.sqrt(3) with np.sqrt(2) when calculating 2d blob radius.
Close issue #2022


shachafl:
@berl , @njmei , can one of you please review this PR?

njmei (approved):

PR #2033, opened 2024-11-14
Create dependabot.yml to update Actions version

shachafl:
This new workflow will automatically open PR to update new Actions version:
actions/cache
actions/checkout
actions/setup-python

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


berl (approved):
good idea.

shachafl:
Didn't work.
Solution is either moving file to parent folder #2039 or repository admin needs to enable "dependabot version updates" under settings->code security:
image

PR #2039, opened 2024-11-30
Github Actions auto update

shachafl:
moving dependabot.yml to .github/ folder

PR #2043, opened 2024-12-04
Delete .github/workflows/dependabot.yml

shachafl:
Tried to delete and remove dependabot.yml file from the wrong  folder, but this didn't register in PR #2039


berl (approved):
let's try this again!

PR #2050, opened 2025-02-07
rolling back to Sphinx v5 as v6-8 introduce issues on ReadTheDocs

shachafl:
Sphinx v6-8 introduce issues on ReadTheDocs navbar Starfish logo (top left) and dropdown "More" sub-menu (see below).
Could also be with relation to other dependencies, like: docutils, sphinx-rtd-theme, etc...

In Sphinx v6 the left logo disappears but dropdown menu still works, and the dropdown stops working in v7-8
image

Rolling back to Sphinx v5 and some related dependencies solves the issue:
image


berl (approved):
looks good. And the other various version bumps are due to `pip-tools` keeping everything else up to date?

shachafl:
Yes, minor upgrades to other dependencies.
FYI, I ended up removing `pip-tools` to keep the same method used in the `Makefile` for dependency updates.

PR #2061, opened 2025-03-27
Drop Python 3.9 support due to critical security issue with Dask dependency

shachafl:
We have kept support for Python 3.9 beyond its recommended life span although it was dropped by the scientific community in April 2024 (https://numpy.org/neps/nep-0029-deprecation_policy.html).
A critical vulnerability in Dask <= 2024.8.2 forces us to drop Python 3.9 as Dask dropped support for Python 3.9 in version 2024.8.1.
After this change we will still have at least 3 working versions of Python (3.10-12).

Subtasks:

1. Add MacOS and remove python 3.9 in automatic GitHub Actions tests (.github/workflows/starfish-prod-ci.yml)
2. Update 4 requirements files
3. Bump python version to 3.10 in docker/Dockerfile
4. Update docs (README.rst & docs/source/installation/index.rst)
5. Update comments in examples/quick_start/plot_quick_start.py
6. Remove Python 3.9 in packaging files (pyproject.toml)
7. Update CHANGELOG.md


shachafl:
The security issue with Dask has been removed in the Github Security section, so need to drop Python 3.9 yet.

Regarding the test failure on MacOS with new Apple silicon (M1+), it is probably due to slight differences in floating-point arithmetic between different hardware architectures. This could be mitigated by adding a small tolerance when comparing image registration results: 
```
# Adding a tolerance
tolerance = 1e-4

assert np.allclose(
    expected_registered_values,
    registered_image.xarray[2, 2, 0, 40:50, 40:50],
    atol=tolerance
)
```

shachafl:
Superseded by #2174

Issue #2062, opened 2025-04-01
Registration fails with latest starfish (and scikit-image)


iimog:
#### Description
I ported an existing pipeline from `starfish` 0.2.2 to version 0.3.2 (also updating python and dependencies). The registration step used to work well. Without modifications to the code, the registration fails now. There is no error message, images are simply not registered.

#### Steps/Code to Reproduce
I boiled down the problem to the following reproducible example ([data uploaded for 30 days here](https://upload.disroot.org/r/LsMfrrzX#Xx8GT37w6ddzuQ2XgL1TibvWrnmoL2pEQhowOWqmPjk=)):
```python
from starfish import Experiment
from starfish.image import ApplyTransform, LearnTransform
from starfish.types import Axes
from starfish.util.plot import diagnose_registration
import matplotlib.pyplot as plt

exp = Experiment.from_json("experiment.json")
fov = exp.fovs()[0]
img = fov.get_image("primary")

learn_translation = LearnTransform.Translation(
    reference_stack=img.sel({Axes.ROUND: 0}), axes=Axes.ROUND
)
warp = ApplyTransform.Warp()

transforms_list = learn_translation.run(img)
print(transforms_list)

registered_img = warp.run(img, transforms_list=transforms_list)
fig, ax = plt.subplots(1, 2, figsize=(10, 5))
diagnose_registration(img, {Axes.ROUND: 0}, {Axes.ROUND: 1}, ax=ax[0], title="before")
diagnose_registration(
    registered_img, {Axes.ROUND: 0}, {Axes.ROUND: 1}, ax=ax[1], title="after"
)
plt.savefig("reprex.png")
```

This example is simplified and uses a different strategy for registration (using nuclei rather than intensity projected spots) but I wanted to keep it simple and both approaches suffer from the same problem.

#### Expected Results
```
tile indices: {<Axes.ROUND: 'r'>: 0}
translation: y=0.0, x=0.0, rotation: 0.0, scale: 1.0
tile indices: {<Axes.ROUND: 'r'>: 1}
translation: y=-6.0, x=92.0, rotation: 0.0, scale: 1.0
```

![Image](https://github.com/user-attachments/assets/a30d7c6b-86ad-4e85-a830-6d3bcb8a61ca)

#### Actual Results
```
tile indices: {<Axes.ROUND: 'r'>: 0}
translation: y=0.0, x=0.0, rotation: 0.0, scale: 1.0
[...]

Issue #2057, opened 2025-03-21
Add MacOS testing to Github Actions workflow


shachafl:
To ensure cross-platform compatibility we need to add MacOS to CI/CD workflow, as we currently only test on Ubuntu and Windows.
This is a quick addition to the OS matrix in `starfish-prod-ci.yml`.
This would increase Github Action usage, but still well below the 2000 monthly hours of a free tier.

shachafl:
Adding to `starfish-prod-ci.yml` was easy, resolving the failed tests, is less easy.
Got
```
FAILED starfish/test/full_pipelines/api/test_iss_api.py::test_iss_pipeline_cropped_data - assert False
```
This happens in the image registration part.
Quickest fix is just to add a tolerance when asserting, but it is worth doing more "digging" to check if some parameter exist that might resolve it.

shachafl:
@berl, we should make a new branch for this issue (instead of coupling to dropping Python 3.9).
I have more content in the link:
https://github.com/spacetx/starfish/pull/2061#issuecomment-2764746715

I would recommend to start by downgrading numpy to version 1.24.4 from the current 1.26.4

iimog:
I just checked on yet another machine Debian 12 (13th Gen Intel(R) Core(TM) i7-13700HX) with almost identical python environments to my other linux machine (Ubuntu 24.04.2 with AMD Ryzen 9 5900X 12-Core Processor) and my Mac (Sequoia 15.4.1, Mac Book Air 13'', M3, 2024). While all tests pass on the Mac and on the AMD machine, the same test (`starfish/test/full_pipelines/api/test_iss_api.py::test_iss_pipeline_cropped_data` and only this one) failed on the Intel machine because of discrepancies in the registered intensities. Adding the tolerance of `1e-4` to `allclose` makes the test pass.

PR #2009, opened 2024-03-05
Scripts and dependencies updates for Python 3.8 - 3.10

shachafl:
The motivation for this PR was significant algo improvements for blob detection in scikit-image 0.20 released on February 2023, together with some recommendations for the python scientific community about maintaining support for outdated packages and python interpreters (https://scientific-python.org/specs/spec-0000/).

Nevertheless, at this point I only recommend dropping support for python < 3.8, as it seems still very popular, but I don't feel strongly about it if the team prefer to drop support for python < 3.10.

To update dependencies and solve conflicts I used pip-tools (https://github.com/jazzband/pip-tools):

### Make new virtual environment and upgrade/install packages
$ python3 -m venv venv
$ source venv/bin/activate
$ python3 -m pip install --upgrade pip setuptools wheel pip-tools

### To update dependency list, I removed REQUIREMENTS.txt file, and run (this step is only needed when wishing to change dependency versions, see pip-tools docs):
$ rm REQUIREMENTS.txt
$ pip-compile requirements.in --output-file REQUIREMENTS.txt

### Install/update python packages, equivalent to "python3 -m pip install -r REQUIREMENTS.txt"
$ pip-sync REQUIREMENTS.txt

### Install starfish from local repository, in development mode:
$ pip3 install -e .

After further updating requirements/REQUIREMENTS-CI.txt.in and using pip-tools again, I was able to install starfish following the developer guide (https://spacetx-starfish.readthedocs.io/en/latest/developer_guide/) and run the tests listed in starfish-prod-ci.yml using the Makefile.

Starting to summarize the PR (by commits):
1) Changes due to scikit-image ver 0.19-0.21 (not backward compatible): 0f63cd0, 6c4c2df, e21a642, e4e6499 (https://github.com/scikit-image/scikit-image/pull/6968), b21f2af6.

2) Upgrading to pandas 2.0 with backward compatibility: 
3653ad7

3) Napari >= 0.4.18 (& Vispy >= 0.12): dropped support for anisotropic spot size (https://forum.image.sc/t/anisotropic-point-sizes/83388). Here I left most of the code as anisotropic spot size is suppose to return at a later Napari/Vispy version. I only used np.mean() to get a single spot size, but don't feel strongly about it if min(), max(), etc. is preferred. 19b4e00

4) Updating dependencies in REQUIREMENTS.txt: d5c1835, 68d338d
5) Deprecated numpy dtypes: 7b24d34

[...]

PR #2063, opened 2025-04-01
Add normalization argument to LearnTransform.Translation with default value None

iimog:
these are passed to the internal call of `skimage.registration.phase_cross_correlation`. This allows users to overwrite the `normalization` argument to restore the skimage<0.19 behavior of `normalization=None`, thereby fixing: #2062

I replaced the unused `*args` with `**kwargs`. If the `*args` are still needed, I can instead add `**kwargs.`

I decided to add this in `run` because that required the least amount of change. But it might be a better idea to add it to init instead (or in addition). Therefore it is specified once when the Translation object is initialized (and can potentially be overwritten in the calls to run).

We might consider adding additional documentation or selecting `normalization=True` as the default.


shachafl:
Hey @iimog,
Thanks for catching that and preparing the PR.
You can blame me for the scikit-image upgrade from 0.18.3.
I did notice the change in phase_cross_correlation() and had to update a couple of test files, but for my case and the tests in `starfish` the differences weren't significant (see #2009).
Seeing your test example and thinking that most users would probably benefit better noise handling than uniform illumination issues. I think we should explicitly add the key-argument with default `normalization=None` to `phase_cross_correlation()` with a reference to the relevant section in scikit-image docs.
Ideally, you should also add your example to a test file, maybe `starfish/core/image/_registration/LearnTransform/test/test_translation.py`?

@berl what do you think?

iimog:
Hi @shachafl,
My pleasure. Actually, I want to thank you (rather than blaming you) for upgrading `scikit-image`. :blush: :smile: This has been a lot of work, and I'm excited to see `starfish` development continue. Also, installation in a current Python environment is now much easier.
I'm happy to revise my pull request according to your suggestion. If I understand correctly, I'll add the `normalization` argument to `init` (potentially with default value `None`) and leave the `run` method unchanged. I'm also happy to add a test case and propose some documentation.

shachafl:
Sounds good to me.
I think a default value `None` makes sense, at least until another user complains :-)

For documentation, besides the docstring in the `Translation` class, as a minimum I would recommend:

1. starfish/examples/tutorials/plot_image_registration.py
2. starfish/examples/quick_start/plot_quick_start.py

A word of caution, Apple Silicon (M1+) might show some differences in floating-point arithmetic compared to Intel, so if you can run on both that's great. I need to resolve some failed tests before adding MacOS to GitHub Actions (#2057).

iimog:
Awesome, I'll start working on this next week.

iimog:
I changed the code, tests, and documentation as discussed (and force-pushed).

However, one test is still failing: `starfish/test/full_pipelines/api/test_iss_api.py`. This test depends on the old default and not only the `expected_registered_values` but also the `gene_counts` are changing when using `normalization=None` as the new default. I wonder whether it is better to run the `ISS` pipeline setting `normalization="phase"` explicitly, rather than changing all the expectations. In other cases, I tested for both settings for `normalization`. @shachafl what do you think?

I'll also run the tests on an Apple Silicon machine later.

shachafl:
@iimog thank you for the update.
When I upgraded scikit-image to ver 0.20 I updated the registration test part of test_iss_api.py as the normalization default changed: (https://github.com/spacetx/starfish/commit/02c654fe6e3dc18b577037c5774d18a9360891a6)
I recommend reverting back (updating the test result) to support your change and the new default.

iimog:
I updated the expectations accordingly. Restoring the old `expected_registered_values` did not work, apparently the expectation still changed from before the upgrade to scikit-image (but they are much closer with `normalization=None` and all the other values match the old expectation).

 I don't know why the lint check fails, but it was not during the linting, but during the installation. So I don't think this is my fault. :slightly_smiling_face:

shachafl:
It all runs fine on my machine.
There is an issue with the cache in Github actions. I have opened an issue https://github.com/actions/setup-python/issues/1101.

iimog:
I successfully ran the tests on an M3 Mac (running Sequoia 15.3.1). Only the napari tests are failing (`pytest -m 'napari'`), both on my Linux machine and on the Mac. 

This has nothing to do with this pr, but one required change is `edge_color` to `border_color` here: https://github.com/spacetx/starfish/blob/b8fdfa94e2ac655a46aa265bbd36be466842f412/starfish/core/_display.py#L282
`edge_color` was deprecated in napari 0.5 and is removed in napari 0.6 (current stable version since May 1). However four of the tests still fail

```
starfish/core/test/test_display.py FF..FF..                              [100%]

[...]

=========================== short test summary info ============================
FAILED starfish/core/test/test_display.py::test_display[stack-spots-masks] - ValueError: cannot convert float NaN to integer
FAILED starfish/core/test/test_display.py::test_display[stack-spots-     ] - ValueError: cannot convert float NaN to integer
FAILED starfish/core/test/test_display.py::test_display[     -spots-masks] - ValueError: cannot convert float NaN to integer
FAILED starfish/core/test/test_display.py::test_display[     -spots-     ] - ValueError: cannot convert float NaN to integer
```

shachafl (approved):
Looks good to me.
Thanks again for doing all the hard work!

Last thing, please open a new issue with the failed napari testing. I wasn't vigilant with the napari testing so far.

PR #2068, opened 2025-05-09
add MacOS to Github actions workflow

shachafl:
Thanks to @iimog work on fixing #2062 with #2063 (setting normalization=None instead of default "phase" in starfish wrapper for scikit-image.registration.phase_cross_correlation() and adjusting expected values), it seems that we can now add MacOS to `.github/workflows/starfish-prod-ci.yml` without failing tests. As there is no normalization in the image registration step, we don't get the slight differences in arithmetic calculations on Intel hardware vs. Apple silicon (probably FFT implementation but not important).


berl (approved):
this is nice! I look forward to less testing at home.

Issue #1870, opened 2020-04-13
blob throws error when run with no `reference_image` and `is_volume=False`


mattcai:
#### Description
When using `BlobDetector` to find spots in an `ImageStack` without using a `reference_image` and with `is_volume=False` error is thrown because `BlobDetector` is hard coded to parse the result of `skimage.feature.blob_log` run on 3D numpy array.

Using a `reference_image` doesn't throw this error because it ignores the `is_volume` argument and runs everything like `is_volume=True`.

#### Steps/Code to Reproduce
[code to reproduce error](https://gist.github.com/mattcai/42d6123d934f988be37c5cd2b08f5e80)

#### Expected Results
Expect to return an `IntensityTable`

#### Actual Results
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-57-d5060f4c30ec> in <module>
      7     measurement_type='mean',
      8 )
----> 9 bd_spots = bd.run(image_stack=wth_imgs)

~/starfish-CI/starfish/starfish/core/pipeline/algorithmbase.py in helper(*args, **kwargs)
     21         @functools.wraps(func)
     22         def helper(*args, **kwargs):
---> 23             result = func(*args, **kwargs)
     24             if result is not None:
     25                 method_class_str = str(args[0].__class__)

~/starfish-CI/starfish/starfish/core/spots/FindSpots/blob.py in run(self, image_stack, reference_image, n_processes, *args)
    193                 func=spot_finding_method,
    194                 group_by=determine_axes_to_group_by(self.is_volume),
--> 195                 n_processes=n_processes
    196             )
    197             results = SpotFindingResults(imagestack_coords=image_stack.xarray.coords,

~/starfish-CI/starfish/starfish/core/imagestack/imagestack.py in transform(self, func, group_by, verbose, n_processes, *args, **kwargs)
    947             # Note: results is [None, ...] if executing an in-place workflow
    948             # Note: this return must be inside the context manager or the Pool will deadlock
--> 949             return list(zip(results, selectors))
    950 
    951     @staticmethod

/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/concurrent/futures/_base.py in result_iterator()
    596                     # Careful not to keep a reference to the popped future
    597                     if timeout is None:
--> 598                         yield fs.pop().result()
    599                     else:
    600                         yield fs.pop().result(end_time - time.monotonic())

[...]

PR #2076, opened 2025-05-16
fix: index shift correction for blob detection in 2d tiles

### LINKED ISSUE #2075: Blobdetector yields incorrect results
#### Description
Issue discovered by [Leo Juhlin](https://forum.image.sc/t/tutorial-with-blobdetector-yields-incorrect-results/112432/4) when running the example code for [howto-spotfindingresults](
https://spacetx-starfish.readthedocs.io/en/latest/gallery/how_to/assess_spotfindingresults.html#howto-spotfindingresults)
"y-values of detected spots, shown in y-column all being 0, and the radius of the detected spots being extremely large."
This is an index shift error that was missed in #2064 and can be easily fixed.
Unfortunately, the fix brings us back to an old issue #1870 of running blobdetector without reference and `is_volume=False`, because the (X,Y) tile is now missing the z-axis which leads to `IndexError: index 3 is out of bounds for axis 1 with size 3`

This is also related to #1985

#### Steps/Code to Reproduce
Run the jupyter notebook and add the line:
`spots[{Axes.CH:1, Axes.ROUND:0}].spot_attrs.data`

#### Expected Results
Image

#### Actual Results
Image

#### Versions
Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python 3.9.22 (main, May  7 2025, 10:28:35)
[GCC 13.3.0]
NumPy 1.26.4
SciPy 1.13.1
scikit-image 0.24.0
pandas 2.2.3
sklearn 1.6.1
xarray 2023.8.0
sympy 1.14.0
starfish 0.3.3

shachafl:
This closes #2075 & #1870 

There might be another solution to this as I couldn't figure why when running blob detection with `is_volume=False` and with a reference image gives a 4d result (z, y, x, sigma) vs. without a reference image which gives a 3d output (y, x, sigma).

### DIFF
--- a/starfish/core/spots/FindSpots/blob.py
+++ b/starfish/core/spots/FindSpots/blob.py
@@ -137,18 +137,18 @@ class BlobDetector(FindSpotsAlgorithm):
 
         # measure intensities
         data_image = np.asarray(data_image)
-        if self.is_volume:
+        if fitted_blobs_array.shape[1] == 4:
             z_inds = fitted_blobs_array[:, 0].astype(int)
             y_inds = fitted_blobs_array[:, 1].astype(int)
             x_inds = fitted_blobs_array[:, 2].astype(int)
             radius = np.round(fitted_blobs_array[:, 3] * np.sqrt(3))
             intensities = data_image[tuple([z_inds, y_inds, x_inds])]
-        else:
+        elif fitted_blobs_array.shape[1] == 3:
             z_inds = np.asarray([0 for x in range(len(fitted_blobs_array))])
             y_inds = fitted_blobs_array[:, 0].astype(int)
             x_inds = fitted_blobs_array[:, 1].astype(int)
             radius = np.round(fitted_blobs_array[:, 2] * np.sqrt(2))
-            intensities = data_image[tuple([z_inds, y_inds, x_inds])]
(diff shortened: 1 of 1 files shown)


berl:
so the reference_image handles things a bit differently, first running`skimage` spotfinding function on the whole stack, or at least that's my read of [this](https://github.com/spacetx/starfish/blob/988a92fdb184b28da6327ebc71ea9c998907584b/starfish/core/spots/FindSpots/blob.py#L192) after a `squeeze` which may be leading to the weird behavior here.

somewhat orthogonally, the `is_volume` parameter is supposed to change the behavior of the spotfinding from per z plane (`is_volume=False`) to finding 3D blobs (`is_volume=True`). This was intended to be used for different kinds of data (sparse sampling in z vs dense sampling) but should be able to work both ways on a given input stack. the underlying `skimage` functions don't know anything about this (they just return 3d information on a 3d image stack or 2d on a 2d stack).
not sure if this helps clarify anything @shachafl, but getting rid of the `is_volume` condition should be done carefully to maintain this functionality

shachafl:
closing this here, as a more comprehensive fix is done in #2154.
The new fix maintains the ability to process a 3d image as a set of 2d tiles instead.

PR #2235, opened 2026-06-08
Multi barcode decoder

Sandmann476:
I added the changes discussed after the repl.


berl:
@Sandmann476 super interesting idea- do you have any demos with this on real (crowded) data?  Looks like the linting is failing, so that should be relatively easy to fix before we see if anything else was impacted.

